PR #803, opened 2014-06-08
More robust cloning function

MartinKolarik:
Correctly clone Dates and RegExps, better detection of primitives and more tests.


sukha1996:
#803

Issue #1357, opened 2014-10-14
Computed Values in Components


WaltzingPenguin:
In 0.4.x I have code that looks like this:
http://jsfiddle.net/L38cnjqv/1/

After upgrading to 0.6.0, this code no longer works and gives the error "Computed properties without setters are read-only. (This may change in a future version of Ractive!)":
http://jsfiddle.net/L38cnjqv/2/

I can get around this by creating a do-nothing setter for my computed property.  However, the value of "b" does not update correctly; once you start typing into the input the value of "b" is usually one update behind the value of "b" inside of the component.  This issue appears to have been introduced by 0.5.x:
http://jsfiddle.net/L38cnjqv/3/

PR #1358, opened 2014-10-14
Prevent setting computed props with no setter across component bindings (#1357)

martypdx:
Avoid setting a computed property that doesn't have a setter that's involved in a component binding.

This seems rather odd, but I think it's the right thing to do, as it will fail otherwise. See #1357

### DIFF
--- a/src/shared/createComponentBinding.js
+++ b/src/shared/createComponentBinding.js
@@ -30,6 +30,8 @@ Binding.prototype = {
 	},
 
 	propagateChange: function ( value, newIndices ) {
+		var other;
+
 		// Only *you* can prevent infinite loops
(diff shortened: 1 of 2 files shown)


brucou:
what's the update on this? I see that is closed, does that mean that the bug no longer occurs? I have the same problem occuring to me (I am using version 0.6.1). Should I install 0.7 or rather apply the recommendation to not use a computed property ion to not use a computed property that doesn't have a setter that's involved in a component binding?

martypdx:
@brucou Just based on the date, I'd say this was fixed awhile ago. But we have been actively discouraging computed properties with same name as component parameters (though there's still open debate about it).

brucou:
Thanks Marty. Then I guess my problem has another source, I will investigate more thoroughly and post a test case if I can't solve it.

PR #1424, opened 2014-10-28
Add direct support for changing partials to Ractive instances - RFC

evs-chris:
Even after getting support for expressions in partial sections, I still have times that I need to change the template for a partial and have it propagate to all of places that it is rendered. This replaces the hacky

``` html
{{#rendered}}{{>partial}}{{/}}
```

``` js
ractive.partials.partial = 'new partial';
ractive.set('rendered', false);
ractive.set('rendered', true);
```

with

``` html
{{>partial}}
```

``` js
ractive.setPartial('partial', 'new partial');
```

Beyond not littering the template with special conditionals, the latter also has the benefit (?) of updating the partial everywhere it is rendered, including down the component hierarchy where the component is using an inherited partial.

This is definitely not ready for merging, as I haven't added any tests and am not sure that this is even something that is desirable for general inclusion.

Here's another sample that I think is fun for demo purposes and might be useful in a docs refresh move to Ractive:

``` html
Config:<br/><textarea rows="5" cols="40" value="{{cfg}}">{{{ 'Name: <input value="{{name}}" />' }}}</textarea>
<button on-click="setPartial('config', cfg)">Update Config</button><br/>
Template:<br/><textarea rows="5" cols="40" value="{{tpl}}">{{{ 'Hello, {{name}}!' }}}</textarea>
<button on-click="setPartial('partial', tpl)">Update Template</button>
<div>{{>config}}</div>
Output:<div>{{>partial}}</div>
```

You can add `Name: <input value="{{name}}" />` to the config box and `Hello, {{name}}!` to the template box and get an interactive form after clicking the buttons.

### DIFF
--- a/src/Ractive/prototype/resetPartial.js
+++ b/src/Ractive/prototype/resetPartial.js
@@ -34,6 +34,17 @@ export default function( name, partial, callback ) {
 			else if ( item.type === types.COMPONENT && item.instance ) {
 				collect( item.instance.fragment.items, dest, item.instance );
 			}
+
+			// if the item is an element, process its attributes too
+			if ( item.type === types.ELEMENT ) {
+				if ( isArray( item.attributes ) ) {
+					collect( item.attributes, dest, ractive );
+				}
+
+				if ( isArray( item.conditionalAttributes ) ) {
+					collect( item.conditionalAttributes, dest, ractive );
+				}
+			}
 		}
 	}
 


tomekmarchi:
Bug with resetTemplate on dynamic components. Example 
http://jsfiddle.net/1vt64zd7/1/

Works only if the component is explicitly placed. If you .toHTML it you will see it updated internally it just doesn't re-render correctly back where it was.

This is a must for dynamic applications that require hot-reloading. I myself am building an app that uses this exact setup. I'm also working on a hot reload plugin for exotic use cases like this.

Issue #1429, opened 2014-10-30
Complex Component Configuration


michaelsogos:
Hi guys,

I'm working with ractive components and i met this issue.
If my component have few simple configuration, then i can use attributes without any poblem:

//THIS IS THE COMPONENT OBJECT
var Grid = Ractive.extend({
    template: '&lt;div class="{{class}}" id="{{id}}">&lt;/div>'
});
Ractive.components.grid = Grid;

//THIS IS THE DOM IMPLEMENTATION
&lt;grid class="col-md-8" id="gridTestID">&lt;/grid>

THIS WORK !!!

But if i need a more complex configuration like this:
//COMPONENT TEMPLATE
&lt;script id="GridTemplate" type="javascript/ractive">
&lt;div class="{{class}}" id="{{id}}">
&lt;!-- here there is static dom parts -->
&lt;div class="header"> HERE SHOULD INJECT HEADER PARTS &lt;/div>
&lt;!-- here there is static dom parts -->
&lt;div class="body"> HERE SHOULD INJECT BODYPARTS  &lt;/div>
&lt;!-- here there is static dom parts -->
&lt;div class="footer"> HERE SHOULD INJECT FOOTER PARTS  &lt;/div>
&lt;!-- here there is static dom parts -->
&lt;/div>
&lt;/script>

//COMPONENT OBJECT
var Grid = Ractive.extend({
    template: "#GridTemplate"
});
Ractive.components.grid = Grid;

//THIS SHOULD BE THE DOM IMPLEMENTATION
&lt;grid class="col-md-8" id="gridTestID">
&lt;header>
&lt;ul>...&lt;/ul>
&lt;/header>
&lt;bodypart>...&lt;/bodypart>
&lt;footer>...&lt;/footer>
&lt;/grid>

How can i reach this?
My own solution (but not like me) is to use attributes like this
&lt;grid 
class="col-md-8" 
id="gridTestID"
header="{{[ ... ]}}"
body="{{[...]}}"
footer="{{[...]}}"

Issue #1317, opened 2014-10-08
Named yields for positioning


ghost:
Have read about yields in #1141 and also had come across named yields in Iron Router and Polymer's `<content>` with the `select` attribute. How's about Ractive also have something similar?

**modal template**

``` html
<div class="modal">
  <div class="header">{{yield header}}</div>
  <div class="body">{{yield content}}</div>
  <div class="footer">{{yield footer}}</div>
</div>
```

**using modal**

``` html
<modal>
  {{#yield to="header"}}Header stuff here{{/yield}}
  {{#yield to="content"}}Content stuff here{{/yield}}
  {{#yield to="footer"}}Buttons here{{/yield}}
</modal>
```

I typically find myself writing modals (Bootstrap modals) whose title, content and button sets are a bit different from each other, but their overall form and theme is the same. I wouldn't want to rewrite entire modals though. The above code is just a sample but the idea is there. Syntax could be a bit better. Thoughts?

Issue #1173, opened 2014-09-01
Overriding component partials


Rich-Harris:
This is a continuation of #1141 - specifically, the part about defining partials for a component. Right now, you can define template-wide 'inline partials' like so:

``` html
<ul>
  {{#each items}}
    {{>item}}
  {{/each}}
</ul>

<!-- {{>item}} -->
  <li class='{{ active ? "highlight" : "" }}'>{{id}}: {{description}}</li>
<!-- {{/item}} -->
```

This means you don't have to have separate files for partials (extra HTTP requests), add complexity to your build process, or deal with multiline strings in your JavaScript (eww).

My proposal is that we use the same syntax inside a component's contents, to define the partials _for that component_. Suppose you had a list component (maybe it's a virtual infinite scroll list, or something fancy like that), and you want to be able to specify how each item is actually rendered:

``` html
<fancy-list items='items'>
  <!-- {{>item}} -->
    <li class='{{ active ? "highlight" : "" }}'>{{id}}: {{description}}</li>
  <!-- {{/item}} -->
</fancy-list>
```

If the `<fancy-list>` component already had a default `item` partial, this would override it. Anything _outside_ the inline partial(s) would be treated as the content for the `{{yield}}` block, if the component template included one.

@lburgess raised a good point on #1141 - it'd be nice to be able to reuse partials that are defined in the _template_ but not in the _component_. One suggested syntax:

``` html
<fancy-list items='items'>
  <!-- {{>item: fancyItem}} -->
</fancy-list>

<!-- {{>fancyItem}} -->
  <li class='{{ active ? "highlight" : "" }}'>{{id}}: {{description}}</li>
<!-- {{/fancyItem}} -->
```

(or just `{{>item: item}}` if they had the same name). Reusing partials like this _might_ be a bit YAGNI-ish - I'm not sure that the initial implementation (of inline partials inside component blocks) should include that, but I'm always open to persuasion.

Two fallouts from this:
- template-wide inline partials can only be declared at the root (this may already be the case - it probably should be at any rate)
- it's time to deprecate `{{>content}}` - warnings at first, removal later. (Obviously if someone actually declares an inline `content` partial, no warning should be shown.)

This is more of an 'intent to implement' than anything, but if anyone has feedback and ideas then please chime in.

Issue #583, opened 2014-04-14
Ability to override partial within an in-line component


lburgess:
It would be nice if it were possible to override one or more partials within a child component, without having to sub-class to override the partial.

eg.

``` javascript
var ListComponent = Ractive.extend({
template: '<ul>{{#items}}<li data-id="{{id}}">{{>item}}</li>{{/items}}</ul>'
,partials: {
    item: '{{text}}'
}
});
```

``` html
<div id="my-template">
    <list-component>
        <!-- {{>item}} -->
        <a href="{{url}}">{{title}}</a>
       <!-- {{/item}} -->
   </list-component>
</div>
```

The ability to share a partial which exists within the parent context with multiple child components would also be nice, something like this perhaps?:

``` html
<div id="my-template">
    <list-component data-items="{{firstListOfItems}}" partial-item="{{>listItem}}"/>
    <list-component data-items="{{secondListOfItems}}" partial-item="{{>listItem}}"/>
    <list-component data-items="{{thirdListOfItems}}"/><!-- uses default partial -->
    <list-component data-items="{{listOfPeople}}" partial-item="{{>listItem2}}"/>
</div>

<!-- {{>listItem}} -->
       <a href="{{url}}">{{title}}</a>
<!-- {{/listItem}} -->

<!-- {{>listItem2}} -->
       {{firstName}} {{lastName}}
<!-- {{/listItem2}} -->
```

PR #1456, opened 2014-11-08
Scoped inline partials and named yields - RFC

evs-chris:
This attempts to address a handful of issues ( #583, #1317, #1429, #1173 ). It basically moves inline partial parsing to the conversion process instead of having a shotgun regex extraction before beginning conversion. After the template is parsed, the partial nodes are extracted from the template and attached to the parent nearest instance (base or component). This allows:

``` html
{{>part1}}
<!-- {{>part1}} -->
  I am part1 in the base instance
<!-- {{/part1}} -->
<cmp>
  <!-- {{>part1}} -->
    I am part1 attached to this cmp
  <!-- {{/part1}} -->
</cmp>
```

`cmp` can have a template (or partials) include `{{>part1}}`, and the instance from the template will have its provided partial overridden with the inline partial. Partials can still be defined anywhere in the template, and they will still be stuck on their nearest Ractive parent.
### Named yields

Yielders get an extra conversion step, like partial context, that allows a name to be specified after `yield` e.g. `{{yield foo}}`. Instead of referencing a single `yieldTemplate` though, they reference either the root content of the component (internally represented as the name `''`) or an inline partial from the content of the component. Only inline partials from the component content can be referenced. The same rule of one yielder per component still applies, but it is expanded to one yielder per name per component.

``` html
<cmp>
  base content
  <!-- {{>yieldMe}} -->
    I can be yielded
  <!-- {{/yieldMe}} -->
</cmp
```

``` js
var cmp = Ractive.extend({
  template: '{{yield}} {{yield yieldMe}}'
});
// results in 'base content I can be yielded'
```
### Proposed extra section

The inline partial syntax is sometimes a little cumbersome. With the moving of the partial extraction process to be inline with the rest of parsing, an additional section could easily be introduced, since all it would need to do would be to drop the appropriate node in the template. Partial extraction and scoping would handle the rest. I like something like

``` html
{{#block foo}}
  inline partial content as usual
{{/block}}
<!-- or -->
{{#yield foo}}
  the same
{{/yield}}
<!-- UPDATE -->
{{#partial foo}}
  bar
{{/partial}}
```
### Issues
- ~~There are no tests yet.~~
- ~~Only allows `{{ }}` and `[[ ]]` for inline partials definitions. This can be addressed. It is left out for now because I didn't want to create a regex for every parse attempt for partials.~~
- If there isn't a base instance with all necessary components in place passed to the parser, the components won't get their inline partials. I don't think components work if they aren't already available at fragment-making time anyway.
- Yielders and inline partials can be mixed within a component, which may be confusing. I see that flexibility as beneficial, though.
- The differences between partials and yielders has blurred in my head, so I'm not sure I haven't broken something terribly. The existing tests pass though.

### DIFF
--- a/src/virtualdom/items/Component/initialise/createInstance.js
+++ b/src/virtualdom/items/Component/initialise/createInstance.js
@@ -2,6 +2,7 @@ import types from 'config/types';
 import log from 'utils/log';
 import create from 'utils/create';
 import circular from 'circular';
+import extend from 'utils/extend';
 
 var initialise;
 
@@ -10,15 +11,19 @@ circular.push( () => {
 });
 
 export default function ( component, Component, data, mappings, yieldTemplate, partials ) {
-	var instance, parentFragment, ractive, fragment, container;
+	var instance, parentFragment, ractive, fragment, container, inlinePartials = {};
 
 	parentFragment = component.parentFragment;
 	ractive = component.root;
 
 	partials = partials || {};
+	extend( inlinePartials, partials || {} );
+
 	// Make contents available as a {{>content}} partial
 	partials.content = yieldTemplate || [];
-	partials[''] = partials.content;
+
+	// set a default partial for yields with no name
+	inlinePartials[''] = partials.content;
 
 	if ( Component.defaults.el ) {
 		log.warn({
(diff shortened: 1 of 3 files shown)


arxpoetica:
Bump on documentation. This feature might prove really useful, but I'm not understanding it fully.

evs-chris:
The docs in general are due for an overhaul, but this is documented for both sides of it. Is there something in particular you don't understand?

inline partials and scope: http://docs.ractivejs.org/latest/partials#inline-partials
yielding in components: http://docs.ractivejs.org/latest/components#-a-name-yield-a-the-yield-directive

arxpoetica:
My apologies, thanks for pointing me in the right direction!

PR #1496, opened 2014-11-17
First attempt at supporting directives in conditional attributes - RFC

evs-chris:
** *__Not ready to merge__* **

This takes the approach of conditional attribute handling and extends it to check for directives before updating the attributes on the parent element. The code in edge right now will let you put any sort of template that you'd like into the attributes area of an element. Then when those parts of the template update, it stringifies it, creates a `div` with the string for attributes, and reads attributes off of the `div`. This lets you do neat things, like have a partial to hold attributes and have it switch based on current context.

```js
ractive.resetPartial( 'attrs', 'class="foo" on-click="click()"');
```
```html
<button {{>attrs}}>Click</button>
```

The downside to this approach for directives is that mustaches are already processed (transcluded?), so any directive that wants to bind refs can't because it gets its value re-parsed from a rendered template. Any directives that do use mustaches, like proxy events and decorators, will get updated appropriately when their underlying refs do, but they will be completely torn down and set back up. That's not the most efficient thing to do, certainly.

I haven't been able to come up with a way to handle this at the fragment level to avoid parsing at render time. For instance, I'm not sure how to handle things like this contrivance:
```html
<button {{>attrs}}>Click</button>
```
```js
ractive.alert = function(message) { console.log(message); };
ractive.set( 'js', true );
ractive.resetPartial( 'attrs', '{{#js}}onclick="alert(\'hello\')"{{else}}on-click="alert(\'hello\')"{{/}}' );
ractive.find( 'button' ).click();
ractive.set( 'js', false );
ractive.find( 'button' ).click();
```

### What's Supported

So far, this supports events, lazy, twoway, and decorator. The code is still a little grungy with clean-up and features pending architectural guidance.

### DIFF
--- a/src/virtualdom/items/Element/Binding/GenericBinding.js
+++ b/src/virtualdom/items/Element/Binding/GenericBinding.js
@@ -11,9 +11,14 @@ GenericBinding = Binding.extend({
 		return this.element.node.value;
 	},
 
-	render: function () {
+	updateLaziness: function() {
 		var node = this.element.node, lazy, timeout = false;
-		this.rendered = true;
+
+		// drop the lazy handlers
+		if ( this.handler ) {
+			node.removeEventListener( 'input', this.handler, false );
+			node.removeEventListener( 'keyup', this.handler, false );
+		}
 
 		// any lazy setting for this element overrides the root
 		// if the value is a number, it's a timeout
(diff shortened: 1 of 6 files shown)


evs-chris:
Superceded by #2369

PR #1546, opened 2014-12-05
Avoid referencing the wrong keypath index during eventhandler rebinds

### LINKED ISSUE #1545: BUG Event directives do not work correctly when used within an iteration for an array which is later manipulated
This should cover it, but let me know if you require more details:
http://jsfiddle.net/vn9tr1sn/1/

evs-chris:
The callback is created with a capture on `i`, which is then incremented at the end of the loop. So, when the callback is called, `i` is now `i++` and weird things happen in the `EventHandler.keypaths`.

My javascript-fu is a little weak here. Is this the best way to handle this sort of thing?

Fixes #1545


anton-ryzhov:
This merge causes some issues — #1645

PR #1563, opened 2014-12-09
basic performance tests

Rich-Harris:
This scratches an itch I've had for a while - it provides a (for now, rather basic) way to test performance across different versions of Ractive.

Alarmingly, it looks like we've got slower since 0.6.1, at least in the tests that are included so far! Without this, we might have shipped that performance regression - now we have a chance to fix it.

screen shot 2014-12-09 at 4 25 05 pm

The README in the `perf` folder explains how to set it up and add more tests. Contributions of all kinds welcome, as ever.


evs-chris:
I'm experiencing something very odd here. I was trying to use the perf tests to track down a slow spot in my code, and when I run the tests in Chrome, they run in 1/3 the time if the profiler is running. I tried profiling my code and the same thing happens. Is the profiler made with cheetah blood or something?

Rich-Harris:
That sounds nightmarish. Maybe the solution to web development performance woes is

```js
console.profile( 'my app' );
myApp();
console.profileEnd( 'my app' );
```

PR #1567, opened 2014-12-10
gh-1566

### LINKED ISSUE #1566: Broken mapping chains prevent reference resolutions
See http://jsfiddle.net/vfdyah1n/. Haven't quite got to the bottom of this yet. At one point [this method](https://github.com/ractivejs/ractive/blob/dev/src/shared/parameters/createParameters.js#L88-L96) is called with `undefined` as the second argument, which seems to be related to the problem, though this didn't work in 0.6.1 either by the looks of it.

Rich-Harris:
**note ready for merge**

This fixes #1566. The `addMapping()` method is sometimes called with `undefined` as the second argument - by itself that's fine (if you have `<foo bar='{{baz}}'/>` you want to create a `bar` mapping on `<foo>` even if you don't yet know what `baz` resolves to), but it causes problems when you try to create mappings that 'leapfrog' through multiple levels. One solution is to only map one level at a time. (Dependants are still registered with the viewmodel at the top of the chain.)

However it breaks a load of other tests related to reverse mappings - I'm going to raise a separate PR about that.

### DIFF
--- a/src/shared/parameters/createParameters.js
+++ b/src/shared/parameters/createParameters.js
@@ -86,11 +86,9 @@ ComponentParameters.prototype = {
 	},
 
 	addMapping: function ( key, keypath ) {
-		// map directly to the source if possible...
-		var mapping = this.parentViewmodel.mappings[ keypath ];
 		return this.mappings[ key ] = new Mapping( key, {
-			origin: mapping ? mapping.origin : this.parentViewmodel,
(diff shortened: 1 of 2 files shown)


evs-chris:
This code was stolen and integrated in #1625, so I'll close this.

For the special references resolution, I stepped through it and it looks like it happens immediately, so I don't think that should be a problem.

PR #1577, opened 2014-12-13
Refactor

Rich-Harris:
**not ready for merge**

This branch exists for general tidy-up, pursuant to #1568. It may turn out that there are some things we'd like to tidy up that will involve some breakage - am going to jot those things down as comments here as I find them, so they can be revisited later. For example, I'd like to remove [this](https://github.com/ractivejs/ractive/blob/51b88a9649be762e36ce3af70aee1b08f7c9215d/src/Ractive/initialise.js#L22-L26) - I think it makes more sense to throw an error if `onconstruct` is used with `new Ractive(...)`. Then we can do away with [this](https://github.com/ractivejs/ractive/blob/51b88a9649be762e36ce3af70aee1b08f7c9215d/src/config/config.js#L121-L132). Some of these gains may seem slight, but in aggregate I think we can ditch a lot of rarely-used code.


martypdx:
Finally had a chance to look through this, definitely code is looking much better :+1:

Agree we should axe `onconstruct` in initialization options. :)

PR #1585, opened 2014-12-17
remove support for long-since-deprecated callbacks from promise-returning methods

Rich-Harris:
This PR removes support for the callback in `ractive.set(keypath, value, callback)`, which was deprecated long ago - `ractive.set()` and [various other methods](http://docs.ractivejs.org/latest/promises) return a promise.

### DIFF
--- a/src/Ractive/prototype/merge.js
+++ b/src/Ractive/prototype/merge.js
@@ -1,10 +1,8 @@
 import { isArray } from 'utils/is';
-import log from 'utils/log/log';
 import { getKeypath, normalise } from 'shared/keypaths';
 import runloop from 'global/runloop';
 
 export default function Ractive$merge ( keypath, array, options ) {
-
 	var currentArray,
 		promise;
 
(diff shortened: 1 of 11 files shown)


anton-ryzhov:
After this fix there is "Something very odd happened. Please raise an issue at https://github.com/ractivejs/ractive/issues - thanks!" error happen.

Minimal environment for reproduce — context, conditional component showed after render and non-empty select with 2-way-binding.

http://jsfiddle.net/t6nLx4cu/ (fails with "Ractive Edge", works with "Latest")

```js
new Ractive({
    el: 'body',
    template: '#layout',

    data: {
        context: {},
    },

    components: {
        component: Ractive.extend({
            template: '#component',
        }),
    },

    onrender: function() {
        this.set('show', 1);
    },
});
```

```html
<script id="layout" type="text/ractive">
{{# context }}
    {{# show }}
        <component value="{{ value }}" />
    {{/ show }}
{{/ context }}
</script>

<script id="component" type="text/ractive">
    <select value="{{ value }}">
        <option></option>
    </select>
</script>
```

anton-ryzhov:
With &lt;input&gt; too http://jsfiddle.net/t6nLx4cu/1/

martypdx:
The problem (and hinted solution) is [here](https://github.com/ractivejs/ractive/blob/dev/src/shared/parameters/Mapping.js#L59-L62).

Mappings that have never been resolved (`value` in your example) need to be forced to resolve when set by a component. 

It may be as simple as removing the above referenced lines and allowing the owning viewmodel to resolve the set.

anton-ryzhov:
I didn't understand should I use library somehow else to workaround this issue?

Why removing deprecated things had been broken basic functionality? I haven't been using oldstyle API somewhere.

Maybe this patch should be reverted until it will work well? I have to use edge version in my development until next release because of #1610 and #1608

PR #1589, opened 2014-12-18
Fixed invalid template keypath

EduardJS:
Solving `Cannot read property 'v' of undefined` error while sending invalid keypath as template to Ractive instance.


Madgvox:
This was fixed in https://github.com/ractivejs/ractive/pull/1901.

Issue #1548, opened 2014-12-05
Regression bug: root context not available anymore in components


dagnelies:
Hi,

Here is the fiddle:
http://jsfiddle.net/qy6gcfg3/

Works in latest, broken in edge.

Btw, observe the output of `this.get()`. When inside a component, it returns an internal `_data, _mapping` properties instead of the resulting data.

dagnelies:
Seems to work now on edge... I'll close it if you don't see objections.

Rich-Harris:
@dagnelies I came here because I was looking for the origin of a particular test case... am currently working on #1594 which simplifies a lot of things but reintroduces this issue in a different form.

I'll try and explain as concisely as possible:

* I recently did some work (https://github.com/ractivejs/ractive/pull/1381#issuecomment-60039921) that affects how inter-component data propagation works. This is partly for performance reasons (it's much more efficient!) and partly because the new model is just more sensible - it eliminates a whole load of bugs and complexity around data flow.
* The key difference is that, internally, data has an 'owner'. If you have `<widget foo='{{bar}}'/>` in your template, the value of `foo` inside the widget doesn't get 'copied' from `bar` outside the widget - instead, the widget knows where it needs to look to find the value of `foo`. So there's a 'single source of truth' and a clean separation, but...
* ...there's no longer a `foo` key on the `widget.data` object. So `JSON.stringify(.)` or `widget.get()` doesn't return the same thing as it did before, although `widget.get('foo')` works as expected.
* To minimise breaking changes that result from this, we introduced `ComponentData`, which uses getters (in ES5 situations) or cloning and tracking (in IE8).
* This turns out to involve quite a lot of back-and-forth, and I came to the view that we'd be better off without `ComponentData`, and in fact that we'd be better off without bothering with `ractive.data` at all (which is what #1594 is about)
* As far as I can see, you can still do everything you'd have wanted to use `ractive.data` for. [This test](https://github.com/ractivejs/ractive/blob/4fe19537a27f903795d6a94763cfbeaaf28bd9ad/test/__tests/componentData.js#L1125-L1143) - the one that brought me to this issue - is the only one left that doesn't pass.

[...]

Issue #1413, opened 2014-10-25
Data functions should be able to access the data that was passed in


Rich-Harris:
Fall-out from #1381: http://jsfiddle.net/rich_harris/rh8gtbhv/. cc @skeptic35

If a component's default `data` property is a function, it receives a single argument, which is the `data` property passed in at initialisation. It can either augment this object, or return a new object, which is used instead to create the component's viewmodel:

``` js
MyWidget = Ractive.extend({
  data: function ( data ) {
    console.log( data.answer === 42 ); // true
    console.log( data.foo === 'bar' ); // true
  }
});

var ractive = new MyWidget({
  data: { answer: 42, foo: 'bar' }
});
```

In 0.6.1, that was true for dynamic properties passed to inline components:

``` html
<my-widget answer='{{answer}}' foo='bar'/>
```

Since #1381, there's a catch - the `data` argument passed to the component's `data` function doesn't include attribute values, if they refer to properties that 'belong' to a different instance. So in the example above, there would be a `data.foo` property (because it's static rather than dynamic - i.e. that data belongs to `<my-widget>`), but no `data.answer`.

Because the viewmodel isn't set up by the time the function is called, that means there's no way to find the value of `answer`.
## Options
- Do nothing. But that would cause breaking changes for some people
- Add initial values to the `data` object. Arguably a bit messy (they would be left there, hanging off `this.viewmodel.data` like growths, since the mappings would override the properties when we later do `this.get('answer')`), but easy to do and should fix most if not all cases
- Re-order things so you can do `this.get('answer')` inside the data function. Might not be a bad idea anyway, but there's an awkward chicken-and-egg problem: you'd need to initialise the viewmodel with the data passed in at initialisation, call the function, and then reset the viewmodel. Is that bad? Maybe it's not so bad

Not totally clear on the best way forward. Commented-out test case, which needs updating once we decide what to do, is [here](https://github.com/ractivejs/ractive/blob/9f625fd42a1ba60cc2c9fed7fbddb5e51470e161/test/modules/components.js#L937-L956)

PR #1594, opened 2014-12-18
Experiment: removing ractive.data

Rich-Harris:
**not ready for merge**

This is an experiment that follows #1590 - [no-data](https://github.com/ractivejs/ractive/tree/no-data) is a branch off of [viewmodel](https://github.com/ractivejs/ractive/tree/viewmodel). It follows the thought therein to its logical conclusion - what happens if we get rid of `ractive.data` altogether?

Hear me out:

* There's an API for getting/setting data - doing `foo = ractive.data.foo` or `ractive.data.foo = 'bar'` is a hack, and always has been
* `ractive.data` doesn't know anything about computed properties and mappings, whereas `set()` and `get()` do
* If we remove `ractive.data` (while continuing to offer a means of doing the various things that `ractive.data` used to do), we can remove a decent chunk of code, ward off tricky edge cases (e.g. #1548), and make things simpler for people learning Ractive
* Therefore we should do it

Obviously that's contingent on not losing essential functionality, but I don't (yet) see any reason why we should.

## Magic mode

Magic mode should continue to work:

```js
var model = { name: 'Alice' };

var ractive = new Ractive({
  el: 'body',
  template: 'hello {{name}}',
  data: model
});

// this won't work...
ractive.data.name = 'Bob';

// ...but this would
model.name = 'Bob';
```

## Initialisation functions

This relates to #1413. Instead of doing this...

```js
Widget = Ractive.extend({
  template: '{{message}}',
  data: function ( d ) {
    d.message = 'the answer is ' + d.answer;
  }
});

var ractive = new Widget({
  el: 'body',
  data: { answer: 42 }
});
```

...you'd do this:

```js
Widget = Ractive.extend({
  template: '{{message}}',
  data: function () {
    return {
      message: 'the answer is ' + this.get( 'answer' );
    };
  }
});

var ractive = new Widget({
  el: 'body',
  data: { answer: 42 }
});
```

To my mind that's a better approach anyway - it's more consistent with how things work everywhere else, and it means you're not mutating someone else's object. Obviously there's a little bit of juggling involved to get things into a state where `this.get()` does something other than throw an error or cause an infinite loop, but it's totally possible.

---------

Will continue to work on this, and see if it is actually possible to replace all the existing functionality without having `ractive.data`. (In modern browsers, we could have a lazy accessor that returns the result of `ractive.get()` and prints a deprecation warning. Wouldn't work in IE8, but since no-one is using that as their dev browser, anyone affected would see the warning.) Anyone have any thoughts in the meantime - any reason why this would actually be a terrible idea?


martypdx:
@Rich-Harris I've been background noodling on this for awhile. I see where you're coming from. I wonder if for `0.7` we leave it in, but change (and rename) `parameters` options to be false so that we don't incur the work and performance penalty, unless you need it to not break things.

What I've really started to think is that viewmodels need to be separate from ractive instances. With keypaths becoming objects, it seems that if you had keypath trees, and keypath alias objects (aka mappings) that pointed from one keypath tree to another - you don't need to manage lists of dependencies, dependencies register directly with the keypath object of interest.

It would mean components start out as "viewmodel-less" and only incur the cost of adding new mappings (which would significantly simpler and cheaper) and adding new computations. This should make them more like partials in terms of performance.

Does this makes any sense? Maybe clear as mud, but I think I might be onto something here...

Rich-Harris:
@martypdx I'm definitely intrigued, though I don't think I'm fully inside your head yet!

It would certainly be a significant step away from the current approach, whereby a keypath, even if it has a special internal representation, is conceptually just a string (i.e. an address) rather than what sounds suspiciously like a `Ractive.Model`. What happens if `this.get('foo')` is an object `{bar:someObj}`, and you do `ractive.set('foo.bar', someOtherObj)`? Seems like there'd be some extra work involved in making sure that `ractive.get('foo.bar') === ractive.get('foo').bar` continued to be true, unless I'm missing something (which is highly likely).

All of which makes me feel that we should probably defer it until 0.7 is out, otherwise we'll never get there! It does open up some interesting possibilities though - in the past, you've mentioned checking for object reference equality as a way to 'link' keypaths to each other. If we did go down this road, perhaps it would allow us to do currently-impossible things like this:

```html
<ul>
{{#each records}}
  <li>{{name}}: {{value}} <button on-click='set("selected",this)'>edit this</button></li>
{{/each}}
</ul>

<label>
  Edit selected record ({{name}}):
  <input value='{{selected.value}}'>
</label>
```

Goes against the current (fluxy) grain of thinking, but I'm not totally on board with the new orthodoxy anyway. Still, we should probably table that until the time is right.

> I wonder if for `0.7` we leave it in, but change (and rename) `parameters` options to be false so that we don't incur the work and performance penalty

If we can do everything we need with `ractive.get()`, what's the value of doing so, other than making it slightly easier to migrate from 0.6? (Not a rhetorical question, and the first word of this paragraph deserves emphasis.) I found the code became quite a bit simpler without support for `parameters`.

I say 'found' because the `dev` branch diverged quite a long way from this one... I've merged it back in but now I've created a whole load more work for myself. Will try and keep my focus and see this through to either mergeability or abandonment.

bart-simpson-generator

Rich-Harris:
There's just one test that isn't passing - the one mentioned in #1548. Would welcome any thoughts on how to handle that, before we make a decision about whether to merge this or not.

martypdx:
@Rich-Harris what if, for `0.7`, we kept the data tracking option (the one that kicks in with `parameters: 'legacy'` or on IE8) - but as an opt-in option, and dropped the ComponentData version of things. I bet that would remove >90% of the code you want to axe, but still leave the `data` option available.

Rich-Harris:
@martypdx I'm still unclear as to the motivation - is it purely to make migration easier? I've removed every instance of `.data` from the test suite, replacing with `this.get()` where necessary, and I haven't found a single case where we actually need `.data`.

I understand not wanting to break things but I reckon users of `.data` are a small minority - we should be encouraging people to use the correct APIs. That wasn't possible in all cases before, because of the circularity involved in retrieving data during initialisation, but that's a solved issue now. I feel like continuing to support every possible situation will slow our progress towards 1.0.

martypdx:
> replacing with this.get() where necessary

Ah, gotcha. Just had a peek at the code and the viewmodel is being created before configuration.

Let's give it a shot :+1:

martypdx:
> @martypdx I'm definitely intrigued, though I don't think I'm fully inside your head yet!

I'm often not fully inside my head either :) I started an implementation of this, so sometime later this week I'll push up a branch to help show rather than explain.

> It would certainly be a significant step away from the current approach, whereby a keypath, even if it has a special internal representation, is conceptually just a string (i.e. an address) rather than what sounds suspiciously like a Ractive.Model. 

Yes, I suppose it is. Keypath objects would be specific to the data they point to, though they're not the data per se. This allows them to do most of the work of the viewmodel methods directly. Those methods (get, set, etc.), as they exist currently, receive a keypath object and use the string to look up things like value in cache, wrappers, dependencies, etc. So why not use the keypath object instead of a multitude of caches? Keypath object, because they are in trees, also have an intrinsic notion of up and downstream changes.

There are a number of other cases solved by introducing a KeypathAlias, a simple wrapper that points to another keypath. Component mappings become super simple. Unresolved references can be assigned and then updated when resolved (without having to update any holders of the keypath reference outside of normal dependency notification). 

This would also allow the keypaths object to mutate internally to avoid having to update dependents to new keypath objects (array index updates). 

Special refs like `@index` and `@keypath` could exist as special keypath objects that wouldn't require special handling outside their implementation. It opens the door to synthetic data for arbitrary `@state`  that can be assigned to any node (context) in the keypath tree.

Rich-Harris:
@martypdx I'm starting to see it a bit more clearly now... I like where you're heading

PR #1606, opened 2014-12-24
Make sure binding attributes work for templates that get rendered more than once

evs-chris:
When I originally added this, I sort of assumed that templates were not used more than once, though I'm not entirely sure why. Having the same template used more than once had the interesting effect of only making binding attributes (lazy,twoway) on elements only apply to the first render of the template, since processing directives removes them from the attributes hash. This moves the recording of lazy and twoway attributes to the template instead of the element. This doesn't affect pre-parsed templates unless you pull the template after it has actually been rendered, in which case, it will have `bindingAttributes` keys for elements.

Also, this makes sure lazy event handlers are removed correctly.

Would appreciate a quick review to make sure this is OK when someone gets a spare cycle.


evs-chris:
I have adjusted this to leave the template alone and just skip the binding attributes when creating regular attributes. That was really the only thing I didn't feel right about in this PR.

Any objections?

martypdx:
LGTM

Rich-Harris:
:+1:

PR #1624, opened 2015-01-10
Make in-browser tests a little easier to run

evs-chris:
This makes sure that the plain AMD modules get loaded for in-browser testing and that there are no sourcemaps included to confuse Chrome's (and maybe others) debugger. This may not be as big a deal on case-insensitive filesystems where I'm not sure which file gets picked up (ractive.js the bundle, or Ractive.js the entry-point), but it helps a ton on linux. This is hopefully just an intermediate step until the test suite gets re-arranged.

Also fixes a bug where `requestAnimationFrame` was not being called on `window`, which apparently is not good http://stackoverflow.com/questions/9677985/uncaught-typeerror-illegal-invocation-in-chrome. I'm not sure why that didn't pop up until I adjusted the test suite, but it caused 20 or so cascading test failure.

Any objections?


Rich-Harris:
LGTM. I would guess that the rAF thing is happening because it's being treated as the `default` method of the `requestAnimationFrame` module when it's being treated as a standalone module, rather than as the `requestAnimationFrame` function, or something like that - so we may be able to revert that part of it once the rest of the test suite overhaul is out of the way.

Issue #1629, opened 2015-01-12
Enhancement: {{elseif}}


Madgvox:
I'm in a situation where I basically require a `switch` statement, with access to a 'default' action. It's certainly possible without an elseif, but it's uglier for it.

https://github.com/ractivejs/ractive/pull/662 mentions `{{elseif}}` being put off (understandably) until pain is felt. I'm feeling a bit of the pain. Maybe not enough to justify an implementation, but there's certainly a use case for it.

evs-chris:
This has been on my list for a while, as I have several templates that would benefit from not having to nest `if`s or repeat conditions so much. I looked into it a while back to see what it would take to implement.

Currently, the `else` is picked up as a special reference during post-parse processing and the if block is split into the main conditional and a negated conditional. I think the simplest approach would be to keep the special reference handling, but instead of negating the conditional, link the two (or more for `else if`s) together with a siblings list. Then as a scheduled task after an update, iterate the siblings list and render the first truthy one and unrender all of the others. Any `else` block would always be truthy, and thus, the default case.

As far as template syntax, I don't think anything with `{{# else if }}`-type syntax will end up looking right and could lead to weird scenarios like
```html
{{# if foo }}...{{/if}}
You should not do this...
{{# elseif bar }},,,{{/elseif}}
```

Closee/open sections also don't feel right to me:
```html
{{# if foo }}
  foo is truthy
{{ /# elseif bar }}
  bar is truthy
{{ else }}
  what happens with me, or should I be handlebar-style too?
{{ /elseif }}?
```

I personally lean toward a special, blessed method-call:
```html
{{#if foo }}
  foo is truthy
{{ elseif (bar) }}
  foo is falsey, bar is truthy
  my single argument would be converted into my condition
{{ else }}
  my block is always true, but will only render if my siblings have a falsey value
{{/if}}
```

In that case, `elseif`s basically get the same treatment as `else`s.

martypdx:
> `{{ elseif (bar) }}`

There's nothing special about the parens, could just be: `{{ elseif bar }}` right?

evs-chris:
In this case the parens are required, since the parser would need to pick it up as a method call. `else if` could be added as a special case directly in the mustache parser to avoid having to use parens, but I kind of like them here. Invalid js expressions in non-section mustaches look weird to me.

Madgvox:
[...]

PR #1631, opened 2015-01-13
First attempt at implementing elseif - RFC

evs-chris:
Here's an attempt at implementing `elseif` support per #1629.

### Parsing
It adds another blessed reference, `elseif()`. If one is encountered while processing a section, the function is stripped off and the remaining expression is stashed along with the following nodes for processing in `_parse cleanup`. In `cleanup`, upon encountering a section with an `l` (now an array), is split into conditional mustaches using the stashed expression, or just a `true` expression for plain `else`s. The number of siblings is also stored in the primary mustache in a new `b` attribute.

### Fragment creation
If a node with a `b` is encountered when creating Fragments, it and its siblings are passed a flag to notify them that they have siblings. Once all of the fragments are created, the ones with siblings are gathered up and given references to their sibling collection. Sections with siblings don't evaluate initially and wait for a task to be executed after the current turn of the event loop. The same happens when one of the siblings gets a new value. There is no caching yet, so each time, all of the siblings are reevaluated.

### Syntax
Since, `elseif` is a function, it does require the parens, which I prefer, but that may not be what everyone likes. To me, it looks pretty good though:
```html
{{#if foo}}
  foo is true
{{ elseif (bar) }}
  foo is false, bar is true
{{ elseif (baz || bat || check('ok')) }}
  yeah
{{ else }}
  ok
{{/if}}
```
~~Also, it was waaay less fiddly than special-casing `elseif` or `else if` in the parser.~~ I also considered `elsif`, `elif`, and accepting all three. For stuck together `else if` clauses, I actually prefer the `elsif` from Ruby.

### Status
Currently, all of the tests pass except those that expect the old parse trees to come out of the parser. I only have two very simple tests for `elseif`, with more pending feedback.

So, is the extra complexity worth the functionality provided? Do you hate the syntax? Would you prefer special-cased syntax in the parser?


Madgvox:
[X-posting](https://github.com/ractivejs/ractive/issues/1629#issuecomment-69698290) my comment on the syntax.

===

Reading back over your comment @evs-chris, I realized I view 'if' blocks like this conceptually:

 ```html
{{#if}} // we're starting an if block, start checking for 'ifs'
{{ if cond1 }} // test for first condition
{{ elseif cond2 }} // else test for second condition
{{ elseif cond3 }} // else test for third condition
{{ else }} // else default condition
{{ endif }}
{{/if}} // end the if block, stop checking for 'ifs'
```

If you scrunch the starting and ending condition into the block delimiters, you end up with this:

```html
{{#if cond1 }} // we're starting an if block
{{ elseif cond2 }} // else test for second condition
{{ elseif cond3 }} // else test for third condition
{{ else }} // else default condition
{{/if}} // we're ending the if block
```

Which I think is a pretty pretty way to show it.

evs-chris:
Turns out special casing is easier than I thought it would be:
```diff
index b9303bd..29fc5c0 100644
--- a/src/parse/Parser/expressions/memberOrInvocation.js
+++ b/src/parse/Parser/expressions/memberOrInvocation.js
@@ -24,13 +24,13 @@ export default function ( parser ) {
                        };
                }

-               else if ( parser.matchString( '(' ) ) {
+               else if ( parser.matchString( '(' ) || expression.n === 'elseif' ) {
                        parser.allowWhitespace();
                        expressionList = getExpressionList( parser );

                        parser.allowWhitespace();

-                       if ( !parser.matchString( ')' ) ) {
+                       if ( !parser.matchString( ')' ) && expression.n !== 'elseif' ) {
                                parser.error( expectedParen );
                        }

diff --git a/src/parse/Parser/expressions/primary/reference.js b/src/parse/Parser/expressions/primary/reference.js
index d57a83f..800a03a 100644
--- a/src/parse/Parser/expressions/primary/reference.js
+++ b/src/parse/Parser/expressions/primary/reference.js
@@ -46,7 +46,11 @@ export default function ( parser ) {
        }

        pattern = parser.relaxedNames ? relaxedName : namePattern;
-       name = parser.matchPattern( /^@(?:keypath|index|key)/ ) || parser.matchPattern( pattern ) || '';
+       name = parser.matchString( 'else if' ) || parser.matchPattern( /^@(?:keypath|index|key)/ ) || parser.matchPattern( pattern ) || '';
+
+       if ( name === 'else if' ) {
+               name = 'elseif';
+       }

        // bug out if it's a keyword (exception for ancestor/restricted refs - see https://github.com/ractivejs/ractive/issues/1497)
        if ( !parser.relaxedNames && !dot && !ancestor && keywords.test( name ) ) {
```

evs-chris:
To be clear, that patch allows any of
```html
{{#if foo}}
{{ elseif (bar) }}
{{ elseif baz }}
{{ else if bat }}
{{/if}}
```

Also, I didn't mention it in the first post, but this also works with other sections in a way that should be expected if you've used `{{else}}` with `{{#with}}` or `{{#each}}`. It also works with implicit condition and iteration e.g.
```html
{{#foo}}
{{ else if bar }}
{{/}}
```

Madgvox:
:+1:

guilhermeaiolfi:
Nice. :+1:

martypdx:
:+1:

evs-chris:
I believe this would call for a template version bump, right?

martypdx:
> I believe this would call for a template version bump, right?

Yes, but it's backward compatible so it shouldn't throw an error.

evs-chris:
I have bumped the version and created a PR for the template-spec. I think this is good to go if @martypdx and @Rich-Harris approve. I'm in the midst of a big project with an impending deadline and wanted to shore this up before I disappear for a while longer.

evs-chris:
After a bit of pondering, would it make more sense to have another node type as a container for linked siblings?
```js
{ t: 19, f: [{ t: 4, r: 'foo', f: ['...'] }, { t: 4, x: { r: ['bar'], s: '_1!==10' } }] }
```
I don't think it would really change the complexity of the logic because those sections still have to have a slightly deferred init process. Looking at both, I still prefer the current sibling count version, but I thought I'd ask.

Rich-Harris:
This is bloody impressive stuff. I reckon it'd be better to choose between `elseif` and `else if` and stick to it (I'd vote for `elseif` personally, since every other 'special' reference is a single word), even if it initially seems harmless to be more relaxed - allowing both will probably mean documentation and examples get fragmented and add to a general sense of confusion.

The deferred init stuff makes me *slightly* nervous - is there any mileage in rewriting blocks during post-parse cleanup in the same way as standard `if...else`? I.e.

```html
{{#if foo}}
  <!-- foo -->
{{elseif bar}}
  <!-- bar -->
{{elseif baz}}
  <!-- baz -->
{{else}}
  <!-- qux -->
{{/if}}

<!-- becomes as though we'd done... -->

{{#if foo}}
  <!-- foo -->
{{/if}}
{{#if !foo && bar}}
  <!-- bar -->
{{/if}}
{{#if !foo && !bar && baz}}
  <!-- baz -->
{{/if}}
{{#if !foo && !bar && !baz}}
  <!-- qux -->
{{/if}}
```

It adds some upfront complexity but skirts around the requirement for sections to know about their siblings - less that can go wrong at render time. @evs-chris if that sounds sane, am happy to jump in and have a crack at it, or if you'd rather I keep my mitts out/already tried and rejected this approach then that's fine too

martypdx:
> I'd vote for elseif personally, since every other 'special' reference is a single word

:+1: 

> `<!-- becomes as though we'd done... -->`

Actually it's more complicated, turtles all the way down:
```
{{#if foo}}
  <!-- foo -->
{{else}}
    {{#if !foo && bar}}
      <!-- bar -->
    {{else}}
        {{#if !foo && !bar && baz}}
          <!-- baz -->
        {{else}}
            {{#if !foo && !bar && !baz}}
             <!-- qux -->
            {{/if}}
        {{/if}}
    {{/if}}
{{/if}}
```

However we implement, I think the key is to avoid unnecessary re-rendering of branches that don't logically change.

evs-chris:
> I'd vote for elseif personally, since every other 'special' reference is a single word

Sounds good to me. I was shooting for least surprise when I added the spacey version.

> The deferred init stuff makes me slightly nervous - is there any mileage in rewriting blocks during post-parse cleanup in the same way as standard if...else?

I looked at that initially, but it looked like complex conditionals would get really ugly really fast. (As @martypdx commented whilst I was typing this)

> if that sounds sane, am happy to jump in and have a crack at it

You are more than welcome to have a crack at it!

evs-chris:
I merged the latest conditional attribute fix and got those working with elseif as well. Was there a reason that each template chunk in a `template.m` array was handled separately? I merged them all together and no tests failed, but I don't know if that's copacetic. I suppose it will be moot if @Rich-Harris wrangles multiple branches into a something less complicated than this.

I also didn't realize that and empty object `{}` was considered false for conditionals, as it is true for plain js. So the code in this PR now behaves accordingly.

evs-chris:
Also, removed support for `else if`.

evs-chris:
Closing this in favor of the much nicer #1676.

PR #1638, opened 2015-01-14
experimental jsdom test runner

Rich-Harris:
**not ready for merge (and may never be)**

I thought I'd have a quick crack at running the test suite in JSDOM instead of PhantomJS. Migrating the test runner was fairly straightforward, but some of the async tests never complete, so the whole thing times out eventually. Haven't delved into it too deeply and don't want to get buried in it now, but may as well leave this here anyway.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,7 @@
 language: node_js
 node_js:
 - '0.10'
-before_script:
-- npm install -g grunt-cli; chmod u+x scripts/deploy-edge-to-cdn.sh; grunt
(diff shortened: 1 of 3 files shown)


ghost:
I think I can close this one. Tests still use Phantom, but there's an added Electron setup (only used as a wrapper so Node tests can run via Karma). There's also Sauce Labs tests in place (still needs tweaking, but it's there). Also planning on using headless Chrome/Firefox for better ES6+ support.

Issue #1639, opened 2015-01-14
When using MagicMode on allready created property the setter is called twice


petterek:
When using Object.defineProperty in viewmodel, setting the value seems to 
be triggerd twice. 
The value is bound to a dropdown

    <select value="{{selectedFiletype}}">
        {{#fileTypes}}
        <option value="{{.}}">{{Name}}</option>
        {{/f}}
    </select>

Code that is called

    iRepository.ViewDocumentTypesForUser({ UnitId: data.Id }).then(
        function (res) {
             self.fileTypes = res;
        });
    

    var _selectedFiletype;

    Object.defineProperty(this, "selectedFiletype", {
        configurable: true,
        get: function () {
            return _selectedFiletype;
        },
        set: function (value) {
            _selectedFiletype = value;
            debugger;
            iRepository.GetDocumentListForUnitQuery({ UnitId: self.User.Id, 
    DocumentTypeId: value.Id }).then(
                function (res) {
                    self.allItems = res;
            });

        }
    });

Then the setter for selectedFileType is called twice when I change the 
dropdown

Madgvox:
Would it be possible for you to use a computed value for `selectedFiletype`? It behaves just like `Object.defineProperty`, and it doesn't muck with Ractive's own listeners.

```js
computed: {
  selectedFiletype: {
    get: function () {
        return _selectedFiletype;
    },
    set: function (value) {
        _selectedFiletype = value;
        debugger;
        iRepository.GetDocumentListForUnitQuery({ UnitId: self.User.Id, DocumentTypeId: value.Id }).then(
            function (res) {
                self.allItems = res;
        });

    }
  }
}
```

petterek:
No I wil not be able to use the "computed" feature. 
But is this a bug? Or am I doing something wrong?

Madgvox:
I don't know. I'd need to see more context or an isolated test case since I'm not even sure what the 'this' in your snippet is pointing at.

petterek:
Ok, I will try to make a fiddel to reproduce it

petterek:
http://jsfiddle.net/yxyh3omn/

Seems same thing is happening here

petterek:
Except when changing to the 'last' value of the list

petterek:
Acually it seems to call the sette 2 times every second time I change the dropdown

Madgvox:
Ok, so I stepped through the code and this is essentially what's happening.

What happens when the setter is only called once:

[...]

PR #1640, opened 2015-01-15
Switched variable 'dontTeardownWrapper' to 'teardownWrapper' (c'mon, double negatives are not confusing never)

Madgvox:
This was bugging the crap out of me while looking for the issue in https://github.com/ractivejs/ractive/issues/1639.


Madgvox:
Aaaand I forgot to switch the false checks. Shit.

Issue #1251, opened 2014-09-24
suggestions to make ractive more contributor-friendly


rstacruz:
As it is, it's a little hard to dive into the source of ractive. how would one go about fixing bugs or implementing new features? it's easy to figure out that `/ractive.js` is a file build from other sources, but a few things aren't immediately obvious:

 * the entry point (`src/Ractive.js`)
 * where the actual test files are (not `test/tests`, but rather `test/modules/`)
 * what's the best way to quickly run tests as I edit code?
 * where to use es6-style imports or amdjs

I have a few suggestions that may make ractive.js a bit more friendly to contributors:

* document these conventions somewhere, likely in the wiki or README.md.

* add an `.editorconfig` file. (http://editorconfig.org/)

* add a `scripts: { prepublish: "grunt" }` hook. this will automatically invoke grunt before `npm publish`, and also right after doing `npm install` when checking out ractive from source.

* add `grunt-cli` as a devDependency to make the above possible.

fskreuz:
Will close this in favor of https://github.com/ractivejs/docs.ractivejs.org/issues/47 for the documentation bit. For the others, I'll just leave this for reference:

> where the actual test files are (not test/tests, but rather test/modules/)
Things have tidied up a bit since then. `test/node-tests` for Node, `test/browser-tests` for browsers.

> what's the best way to quickly run tests as I edit code?
Mentioned in README.md. Do a `npm start`.

> where to use es6-style imports or amdjs
> what version of the es6 modules spec is being used
The code uses ES6 and taken care of by Rollup and Babel.

> add an .editorconfig file. (http://editorconfig.org/)
There's already one.

> add a scripts: { prepublish: "grunt" } hook. this will automatically invoke grunt before npm publish, and also right after doing npm install when checking out ractive from source.
> add grunt-cli as a devDependency to make the above possible.
 We use [Gobble ](https://github.com/gobblejs/gobble) now.

PR #1641, opened 2015-01-15
Switched 'dontTeardownWrapper' to 'teardownWrapper'

Madgvox:
This was bugging the crap out of me while looking for the issue in https://github.com/ractivejs/ractive/issues/1639. Double negatives are confusing.


evs-chris:
I haven't had much occasion to mess with wrappers, so my brain usually just elides this code. This makes the code a little more approachable per #1251, so :+1:

PR #1646, opened 2015-01-16
Fixes computations referencing mappings that throw when the mapping resolves

### LINKED ISSUE #1645: "d.dep.setValue is not a function" on nested components
I'm trying to use edge version and found some bug. I've prepared testcase as usual)

Bug appears after merge #1546 and looks related to #1542.

My case: http://jsfiddle.net/d4sgqma2/

All items are required to reproduce - `component1` inherits values from root object, `component2` rewrites context and uses function call and value is defined after first initialization.

Function itself isn't important — it fails with any one. http://jsfiddle.net/d4sgqma2/1/

```js
new Ractive({
    el: 'body',
    template: '<component1 />',

    components: {
        component1: Ractive.extend({
            template: '<component2 foo="{{ bar }}" />',
        }),

        component2: Ractive.extend({
            template: '{{ console.log(foo) }}',
        }),
    },
    
    onrender: function() {
        this.set('bar', 1);
    },
});
```

evs-chris:
This fixes #1645. I'm not 100% positive that this is the correct resolution for this, but it does work. Is there a better place to handle this and/or is there another corner case to cover?

Hat tip to @anton-ryzhov again for an excellent test case!

### DIFF
--- a/src/shared/parameters/Mapping.js
+++ b/src/shared/parameters/Mapping.js
@@ -95,7 +95,14 @@ Mapping.prototype = {
 			this.deps.forEach( d => {
 				var keypath = this.map( d.keypath );
 				this.origin.register( keypath, d.dep, d.group );
-				d.dep.setValue( this.origin.get( keypath ) );
+
+				// if the dep has a setter, it's a reference, otherwise, a computation
+				if ( isFunction( d.dep.setValue ) ) {
+					d.dep.setValue( this.origin.get( keypath ) );
+				} else {
+					// computations have no setter, get it to recompute via viewmodel
+					this.local.mark( d.dep.key );
+				}
 			});
 
 			this.origin.mark( this.keypath );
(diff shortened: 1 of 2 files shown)


anton-ryzhov:
Great! So fast!
I'd been localizing bug for a longer time)

martypdx:
I remember this use case back when we started on Bindings. Would be nice to standardize viewmodel dependent API so logic didn't need to be special cased, but what you have LGTM.

evs-chris:
Something like `viewmodel.resetValue( dep, value )` that would basically call `setValue` for `'default'` deps and trigger recomputation for `'computed'` deps without the full `mark` running?

martypdx:
Yeah, though ideally it's something on the dependent, so maybe computations have an implementation of `setValue` (and maybe it gets renamed `resetValue`). Goal is that code that iterates deps shouldn't have to know what "kind" of dependent it is:

```js
this.deps.forEach( d => {
    d.dep.doSomething();
});
```

PR #1647, opened 2015-01-17
Revert 1641

Rich-Harris:
@Madgvox I'm afraid #1641 caused 41 tests to fail! Totally my fault that it wasn't caught - in the process of updating the build/test scripts to not use grunt, I forgot to put `set -e` at the top of the scripts, which meant that Travis didn't bug out when the tests failed.

So I've had to revert the change. But I agree that double negatives are confusing - I figured that an easier way to get the tests to pass again would be to keep everything as it was, but replace `dontTeardownWrapper` with `keepExistingWrapper`.

Going to merge this immediately as there's currently no http://cdn.ractivejs.org/edge/ractive.js

### DIFF
--- a/src/Ractive/static/adaptors/magic.js
+++ b/src/Ractive/static/adaptors/magic.js
@@ -84,7 +84,7 @@ try {
 			this.updating = true;
 			this.obj[ this.prop ] = value; // trigger set() accessor
 			runloop.addViewmodel( this.ractive.viewmodel );
-			this.ractive.viewmodel.mark( this.keypath, { teardownWrapper: false } );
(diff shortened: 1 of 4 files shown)


Madgvox:
Apologies for that! I figured since Travis passed it it was all good. Out of curiosity, do you know what about the change caused the tests to fail?

Rich-Harris:
No worries! Not sure what was causing it, to be honest I didn't look into it as this weekend is a mad rush. If I had to guess I'd say it was something like an `undefined` value - if `x` was previously `true` and now switched to `false`, then undefined `y` would be treated the same as `x` where previously it was treated as the opposite ¯\\_ (ツ )_ /¯

Issue #1654, opened 2015-01-23
Edge ractive-legacy - errors on Object.defineProperty 


simond-14:
I'm trying to pickup fixes which I think relate to  #1431 but when I load the edge version of ractive from the CDN the site starts erroring in IE8.

It appears related to the use of Object.defineProperty / line 6423 throws an 'Object doesn't support this function'.

This reproduces the problem for me:

```
<html>
<head>
    <meta http-equiv="X-UA-Compatible" content="IE=8" />
</head>
<body>

<div id='container'></div>

<script id='template' type='text/ractive'>
    <p>Hello, {{name}}!</p>
</script>

<script src='http://cdn.ractivejs.org/edge/ractive-legacy.js'></script>

<script>
    var ractive = new Ractive({
        el: 'container',
        template: '#template',
        data: { name: 'world' }
    });
</script>
</body>
</html>
```

Is this a known issue?   Any help appreciated.

Madgvox:
Looks like the `parseJSON` module in the [6to5](https://github.com/6to5/6to5) package, which gets included in the build for `ractive-legacy.js`, does not check to see if `Object.defineProperty` is supported before using it. IE8 only allows `Object.defineProperty` to be used on DOM elements, so an error is thrown. Ractive is actually really good about checking for that sort of thing... apparently 6to5 is not.

simond-14:
Thanks for the quick response.   Is there anything that you can think of that I might be able to do to workaround / fix the problem in the short term?  I'm keen to avoid figuring out a workaround for the issue I came across in 0.6.1 latest when it looks resolved in edge.

Thanks again for the help - much appreciated.

Madgvox:
Initial searches on emulating `defineProperty` in IE8 are negative. It seems like Microsoft decided to make it as good as impossible to create getters and setters for objects _in only IE8_. The only thing I could think of was wrapping it in a function with a try-catch statement:

```js
Object._defineProperty = Object.defineProperty;
Object.defineProperty = function( obj, key, descriptor ) {
    try {
        Object._defineProperty( obj, key, descriptor );
    } catch ( err ) {}
}
```

Which is obviously not ideal as you have to go through a try-catch every time and if it's IE8 nothing happens at all. But maybe it'll work for you? ¯\\\_(ツ)_/¯

simond-14:
crumbs..  thanks for that.   Would the workaround affect much of ractive's functionality?   It seems like it could be a fairly fundamental part of the code..

[...]

PR #1656, opened 2015-01-23
work around 6to5 transpilation issue

Rich-Harris:
fix for #1654?

### DIFF
--- a/src/utils/parseJSON.js
+++ b/src/utils/parseJSON.js
@@ -10,12 +10,12 @@ import getKey from 'parse/Parser/expressions/shared/key';
 
 var JsonParser, specials, specialsPattern, numberPattern, placeholderPattern, placeholderAtStartPattern, onlyWhitespace;
 
-specials = {};
-
-specials[ 'true' ] = true;
-specials[ 'false' ] = false;
-specials[ 'undefined' ] = undefined;
-specials[ 'null' ] = null;
+specials = {
+	'true': true,
+	'false': false,
+	'undefined': undefined,
+	'null': null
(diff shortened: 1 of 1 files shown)


evs-chris:
I _think_ I tried that and 6to5 still helpfully converted it to the "better syntax". You may have to do something sillier like
```js
var specials = [];
specials[ 'true' + '' ] = true;
specials[ 'false' + '' ] = false;
specials[ 'undefined' + '' ] = undefined;
specials[ 'null' + '' ] = null;
```

evs-chris:
Ah yes, the original PR, which _almost_ worked, used this method and still got mangled was #1595. To all of those out there supporting IE8 (myself included), I am deeply sorry, and not just because I broke the legacy build trying to fix the legacy build :)

Rich-Harris:
Building it locally just now gave me this result:

```js
specials = {};

specials["true"] = true;
specials["false"] = false;
specials.undefined = undefined;
specials["null"] = null;
```

Which is okay, because `undefined` isn't a keyword. Maybe 6to5's behaviour changed... @sebmck, @thejameskyle any chance you can shed any light on this? Can you recall if 6to5 started quoting keyword property names at some point? I did a quick issue search but couldn't find anything obvious.

sebmck:
@Rich-Harris 6to5 has been quoting member expression keywords since 1.12.14. I'm not sure I completely understand the issue.

evs-chris:
LGTM. It looks like I most likely had 1.14.4 when I attempted that fix. I probably just did something wonky.

PR #1668, opened 2015-01-28
refactor parser

Rich-Harris:
**not ready for merge**

This follows on from https://github.com/ractivejs/ractive/pull/1631 - I wanted to see if my hypothesis that it would be easier to collapse sequential `{{elseif ...}}` blocks into a series of `{{#if ...}}...{{/if}}` sections. So far I'm fairly sure it is, but in the meantime I got waylaid - I realised that there are some good refactoring opportunities in the parser code.

In particular, the `parser.readExpression()` and `parser.flattenExpression()` methods were both rather odd - the `readExpression` function is just a text-to-AST-node converter like any other (except that the others are functions that you pass `parser` *into*, rather than being *methods* of `parser`, and `flattenExpression` doesn't refer to `parser` in any way).

There was a perfectly good reason for this - treating `readExpression` like other converters introduces some incredibly finicky cyclical dependencies. Since switching to Esperanto, that's not a problem (because ES6 modules :heart: cycles), so those methods can become standalone functions. This is good, because it means that all that expression parsing logic no longer gets included in the runtime-only build, which is now almost 5% slimmer as a result.

To be continued...


Rich-Harris:
Still refactoring here - all the 'converters' (modules that ingest the beginning of a string and turn it into an AST node) have been renamed, e.g. `readText(parser)`, `readElement(parser)` and so on, making the code a bit more self-explanatory.

More significantly, delimiters are represented slightly differently - they live on the parser object itself, and are only mucked about with when we encounter a delimiter change tag (e.g. `{{=<% %>=}}`). Previously, we were doing something frankly rather odd - I won't go into the details, but the upshot is that by avoiding unnecessary memory allocations and closures, parsing is something like 30% faster now.

Rich-Harris:
Just finishing up here. This doesn't do what it originally set out to do (an alternative `elseif` implementation that flattened the result at parse time to avoid any additional work at render time), but it provides a solid foundation for that and other changes.

The `{{...}}` tag is used for more things than it used to be. When the parser was first written, we didn't have `{{else}}` or `{{yield}}` or `{{#partial}}` or `{{>partial context}}` or various other constructs that are now part of the templating language. As they got added, it was always tempting to hack on additional functionality in the form of `if...else` blocks, particularly in [these](https://github.com/ractivejs/ractive/blob/864c183658b5e183675d5cf6ae5c1d057dc36c53/src/parse/converters/mustache.js) [two](https://github.com/ractivejs/ractive/blob/864c183658b5e183675d5cf6ae5c1d057dc36c53/src/parse/converters/mustache/content.js) files.

This PR changes things around a bit, so that things like yields and partial blocks are generated separately. It makes it *much* easier to handle the special cases around e.g...

```html
{{>hyphenated-partial-name additional(context(expression))}}
```

...and will mean that the parser is readier to accept any future innovations (for example we might choose to steal Ember's `{{#each person in people}}` syntax).

The resulting code is hopefully a bit easier for humans to read and for computers to optimise.

PR #1676, opened 2015-02-01
alternative {{elseif ...}}

Rich-Harris:
**not ready for merge**

see #1631 and #1668

### DIFF
--- a/src/config/types.js
+++ b/src/config/types.js
@@ -45,3 +45,4 @@ export var SECTION_IF_WITH   = 54;
 export var SECTION_PARTIAL   = 55;
 
 export var ELSE              = 60;
+export var ELSEIF            = 61;
--- a/src/parse/_parse.js
+++ b/src/parse/_parse.js
@@ -1,5 +1,5 @@
 import { TEMPLATE_VERSION } from 'config/template';
-import { COMMENT, ELEMENT, SECTION_UNLESS } from 'config/types';
+import { COMMENT, ELEMENT } from 'config/types';
 import Parser from './Parser';
 import readMustache from './converters/readMustache';
 import readTriple from './converters/mustache/readTriple';
@@ -103,7 +103,7 @@ StandardParser = Parser.extend({
 			this.error( 'A section was left open' );
 		}
 
-		cleanup( items, options.stripComments !== false, options.preserveWhitespace, !options.preserveWhitespace, !options.preserveWhitespace, options.rewriteElse !== false );
+		cleanup( items, options.stripComments !== false, options.preserveWhitespace, !options.preserveWhitespace, !options.preserveWhitespace );
 
 		return items;
 	},
@@ -147,7 +147,7 @@ parse = function ( template, options = {} ) {
 
 export default parse;
 
-function cleanup ( items, stripComments, preserveWhitespace, removeLeadingWhitespace, removeTrailingWhitespace, rewriteElse ) {
+function cleanup ( items, stripComments, preserveWhitespace, removeLeadingWhitespace, removeTrailingWhitespace ) {
 	var i,
 		item,
 		previousItem,
@@ -155,7 +155,7 @@ function cleanup ( items, stripComments, preserveWhitespace, removeLeadingWhites
 		preserveWhitespaceInsideFragment,
 		removeLeadingWhitespaceInsideFragment,
 		removeTrailingWhitespaceInsideFragment,
-		unlessBlock,
+		spliceArgs,
 		key;
 
 	// First pass - remove standalones and comments etc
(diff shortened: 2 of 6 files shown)


Rich-Harris:
@evs-chris @Madgvox @martypdx it took a few days, but here's what I was [proposing the other day](https://github.com/ractivejs/ractive/pull/1631#issuecomment-71870452). Essentially it implements `elseif` without any changes to render time logic - the blocks are collapsed into sequential `{{#if}}...{{/if}}` sections.

Due to the way empty arrays/objects are treated as falsy when rendering sections, we have to use an `unless` section for *all* the `elseif`/`else` blocks, unless it's a `{{#with foo}}` section in which case we need to wrap the sequential blocks in an `{{#if !foo}}`. But that's as much nesting as needs to happen:

```html
{{#each foo}}
  <!-- foo -->
{{elseif bar}}
  <!-- bar -->
{{elseif baz}}
  <!-- baz -->
{{else}}
  <!-- qux -->
{{/if}}

<!-- becomes as though we'd done... -->

{{#each foo}}
  <!-- foo -->
{{/each}}
{{#unless foo}}
  {{#if bar}}
    <!-- bar -->
  {{/if}}
  {{#if !bar && baz}}
    <!-- baz -->
  {{/if}}
  {{#if !bar && !baz}}
    <!-- qux -->
  {{/if}}
{{/unless}}
```

Those latter expressions are created by combining and manipulating the expression mini-ASTs.

Does this seem like a sensible approach? Any places you can see where it might fail?

evs-chris:
Cleaning up the cruft in the parser also made this much cleaner. Nice!

My only major concern with this approach was complex conditionals. Handling them at parse instead of cleanup greatly simplifies the handling, but a quick contrived test doesn't pass. I don't think the test is flawed, but it may be:
```diff
diff --git a/test/__tests/render.js b/test/__tests/render.js
index f5779a1..dc0f12f 100644
--- a/test/__tests/render.js
+++ b/test/__tests/render.js
@@ -52,7 +52,7 @@ runTest = function ( i ) {

        test( theTest.name, function ( t ) {
                magicModes.forEach( function ( magic ) {
-                       var view, data;
+                       var view, data, j, step;

                        data = typeof theTest.data === 'function' ? theTest.data() : deepClone( theTest.data );

@@ -75,6 +75,15 @@ runTest = function ( i ) {

                                t.htmlEqual( fixture.innerHTML, theTest.new_result, 'innerHTML should match result' );
                                t.htmlEqual( view.toHTML(), theTest.new_result, 'toHTML() should match result' );
+                       } else if ( theTest.steps && theTest.steps.length ) {
+                               for ( j = 0; j < theTest.steps.length; j++ ) {
+                                       step = theTest.steps[j];
+                                       data = typeof step.data === 'function' ? step.data() : deepClone( step.data );
+                                       view.set( data );
+
+                                       t.htmlEqual( fixture.innerHTML, step.result, step.message || 'innerHTML should match result' );
+                                       t.htmlEqual( view.toHTML(), step.result, step.message || 'toHTML() should match result' );
+                               }
                        }

                        view.teardown();
diff --git a/test/testdeps/samples/render.js b/test/testdeps/samples/render.js
index 6683cbc..63183fd 100644
--- a/test/testdeps/samples/render.js
+++ b/test/testdeps/samples/render.js
@@ -1091,7 +1091,25 @@ var renderTests = [
                result: '<p data-foo="bar">result</p>',
                new_data: { bar: null },
                new_result: '<p data-foo="other">result</p>'
+       },
+       {
+               name: 'complex elseif',
+               template: '{{#foo < 21 && bar < 12}}first{{elseif max(4, 5, x, 7) < y * 2}}second{{else}}third{{/}}',
+               data: { max, foo: 22, bar: 13 },
+               result: 'third',
+               steps: [
+                       {
+                               data: { foo: 20, bar: 11 },
+                               result: 'first'
+                       },
+                       {
+                               data: { foo: 30, x: 10, y: 200 },
+                               result: 'second'
+                       }
+               ]
        }
 ];

-export default renderTests;
\ No newline at end of file
+function max() { return Math.max.apply(Math, Array.prototype.slice.call(arguments, 0)); }
+
+export default renderTests;
```

I'm not sure exactly why, but duplicating a test just to change the data and result was making me mad. So I added support for multiple steps on a single render test.

Rich-Harris:
@evs-chris nope, the flaw was definitely in my code! The expression for the final `{{else}}` block needed to be flattened to...

```js
!(_0(4,5,_1,7)<_2*2)
```

...but instead became...

```js
!_0(4,5,_1,7)<_2*2
```

...which isn't the same thing at all. To work around any potential operator precedence issues I've changed it so that it just wraps expressions in parens before negating them. Thanks as ever for your vigilance :)

Issue #1672, opened 2015-01-30
npm and browser version templates mismatched for ractive 0.6.1 (in edge)


kurdin:
``npm install ractive`` and ``wget http://cdn.ractivejs.org/edge/ractive.js`` has same version number ``0.6.1`` but code is different and templates version number mismatched. 

This lead to the issue with compiled template on server can not be used in browsers.

Hmm. I just found that edge version is changed, so maybe this is not valid issue. 

Anyway, it would be nice to do ```npm install ractive@edge``` as well (maybe).

codler:
You can use `http://cdn.ractivejs.org/latest/ractive.js` instead

Rich-Harris:
This came up the other day in #1632 - `npm install ractive@edge` won't work (because you can't overwrite versions on npm, so no moving targets like `edge`), but `npm install ractivejs/ractive#build` does (it just points at the [build branch](https://github.com/ractivejs/ractive/tree/build), which Travis updates on each successful push).

Related: I've opened #1677 to bump the edge version to 0.7.0-edge, as it should be.

The template mismatch thing is a colossal nuisance... best solution is for us to get 0.7.0 out the door!

Rich-Harris:
0.7.x is finally out, so I'll close this issue

krasimir:
@Rich-Harris thanks for releasing 0.7 version. However, we still got the error above. We are pulling the client-side version from here http://cdn.ractivejs.org/latest/ractive.js The version right now is 0.7.1 and that's the same version that we install via npm. So at the end we have same 0.7.1 in the browser and in our gulp setup. I still see 
![image](https://cloud.githubusercontent.com/assets/528677/6707096/5413df0c-cd6c-11e4-96cb-96f96c1999e7.png) in the console.

Madgvox:
What specifically are you using to parse your templates? I've run into this before when using another package ([ractive-loader](https://github.com/rstacruz/ractive-loader)) to parse my templates because it was using a very old version of ractive.

krasimir:
@Madgvox we are using this package https://www.npmjs.com/package/ractive and as far as I can see it is the official one.

krasimir:
As a temporary solution I'm using that build http://cdn.ractivejs.org/releases/0.6.1/ractive.min.js and 0.6.1 version of the npm package.

Madgvox:
Yes, that is the correct package.

[...]

PR #1677, opened 2015-02-02
bump edge version -> 0.7.0-edge

Rich-Harris:
See #1672. Probably makes sense for the edge version to reflect the future rather than the present. Anyone have any objections (i.e. could it break anything)?

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "ractive",
   "description": "Next-generation DOM manipulation",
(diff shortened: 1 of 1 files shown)


evs-chris:
:+1: Seems like having edge and latest have the same internal version has only been painful if it had any effects at all.

Rich-Harris:
Yeah, I reckon I'll go ahead and merge this - if it does cause problems I'm sure we'll find out soon enough

PR #1699, opened 2015-02-09
Add support for regexp literals in expressions

evs-chris:
Based on the docs, I'd say that this was one of those 'round to it things. It makes something I'm working on a fair amount easier, so I thought I'd see if it makes sense for anyone else.

After @Rich-Harris lovely parser refactor, everything is pretty clean on implementation. The only two weird spots are ~~repeated flags that cause an exception (`/foo/gg`) aren't caught here and~~ early tag close detection in the parser died if the first expression in a mustache was a regexp (`{{ /foo/.test(someRef) }}`). ~~The former I don't think is something that would pop up, and I couldn't conjure a regexp to prevent it. It should be fairly easy to fix after the match, but I don't know that it's worth it.~~ The latter is taken care of by checking to see if the close matches a regexp literal before bailing, and if it does, rewinding the parser to roll on.

UPDATE: @MartinKolarik solved #1.


src/parse/converters/expressions/primary/literal/readRegexpLiteral.js
```diff
@@ -0,0 +1,17 @@
+import { REGEXP_LITERAL } from 'config/types';
+
+// as far as I can tell, this only falls over if modifiers are repeated erroneously
+var regexpPattern = /^\/(?:(?=\\\/)..|[^\/])+\/[gimy]{0,4}/;
```
MartinKolarik: `/a\\/b/` and `/a[b/c]d/` will both fail. I don't think you can write a regexp for a regexp, but maybe I'm wrong.

evs-chris: Hmmm. I had no idea you didn't have to escape `/` (and apparently a whole host of other confusing things) in js regexp literals. That seems dangerous and unintuitive to me, but I'm not a native to languages with literal regexps. A quick glance at the ECMAScript (well, the first one I found, anyways: http://bclary.com/2004/11/07/#a-7.8.5) EBNF doesn't look like those are strictly legal, but they definitely work in V8. So is it better to leave regexps out entirely or introduce them with caveats?

MartinKolarik: AFAIK those are all perfectly legal. I've actually started with refactoring the parser (a while ago, but didn't have time for Ractive lately) and I have a working regExp parser  (at least I think so). I can look for the relevant code and submit a PR if you want.

MartinKolarik: OK, it seems I was wrong. The regexp literal syntax isn't so complicated after all. This should cut it:
`/^\/(?:[^\n\r\u2028\u2029/\\[]|\\.|\[(?:[^\n\r\u2028\u2029\]\\]|\\.)*])+\/(?:([gimuy])(?![a-z]*\1))*(?![a-zA-Z_$0-9])/`

evs-chris:
@MartinKolarik Nice! My eyes are still a little crossed, but with a minor capture tweak to make the whole regexp appear in the first capture to satisfy Ractive's parsing method, it looks like that's got it covered.

MartinKolarik:
@evs-chris why `(?![-a-zA-Z_$0-9]))` and not only `(?![a-zA-Z_$0-9]))` (no dash at the beginning)?

evs-chris:
That was a copy/paste fail on my part after having played with it in the console. Not having used js regexps as much as others, I'm still amazed at how forgiving the js parser is.

Rich-Harris:
Once again @evs-chris just goes ahead and implements something I'd assumed would be way too difficult to bother with! Is this ready to merge? If so then :+1: :+1: :+1:

evs-chris:
Hat tip to @MartinKolarik for doing the hard part.

PR #1705, opened 2015-02-10
fixes #1703 - encapsulated styles are forwarded to components until they can be applied

### LINKED ISSUE #1703: Encapsulated CSS fails with some component compositions
Two cases:

1. A component `<outer>` has another component `<inner>` at the top level of its template. The inner styles are applied, the outer styles are not. [Demo](http://jsfiddle.net/rich_harris/2baxp628/)
2. A component `<foo>` has a template `<wrapper><div class='content'></div></wrapper>`. The wrapper's styles are applied, the foo styles are not. [Demo]()

Both are manifestations of the same problem, I think - top-level components do not know that they need to forward the CSS encapsulation to their top-level elements/components.

Rich-Harris:
Aside from the bug (#1703) this fixes, I think I prefer this approach anyway - CSS IDs (i.e., the unique IDs of component constructors that have encapsulated CSS) are passed down from fragment to fragment until they get applied by an element. Seems like the correct way to do it, rather than the slightly hacky approach of each element testing whether its parent element is the component's root element.

### DIFF
--- a/src/Ractive/config/custom/css/transform.js
+++ b/src/Ractive/config/custom/css/transform.js
@@ -2,12 +2,12 @@ var selectorsPattern = /(?:^|\})?\s*([^\{\}]+)\s*\{/g,
 	commentsPattern = /\/\*.*?\*\//g,
 	selectorUnitPattern = /((?:(?:\[[^\]+]\])|(?:[^\s\+\>\~:]))+)((?::[^\s\+\>\~\(]+(?:\([^\)]+\))?)?\s*[\s\+\>\~]?)\s*/g,
 	mediaQueryPattern = /^@media/,
-	dataRvcGuidPattern = /\[data-ractive-css="[a-z0-9-]+"]/g;
+	dataRvcGuidPattern = /\[data-ractive-css~="\{[a-z0-9-]+\}"]/g;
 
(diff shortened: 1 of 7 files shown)


martypdx:
Two questions:
* Why do components inherit parent styles? `{{yield}}` makes sense, but doesn't it break encapsulation for the component?
* As all elements are marked, do we still need the any decent css rule? 

```css
/* Ractive.js component styles */
p[data-ractive-css~="{2}"],   /* need this */
[data-ractive-css~="{2}"] p    /* but do we need this? */
{ 
    font-family: "Comic Sans MS"; 
}
```

Rich-Harris:
Sorry @martypdx, missed those questions. It *was* the case, briefly, that all elements were marked, but that was in error (they weren't being nulled out, basically), so descendant selectors are still necessary. I'm not totally sure I understand the first question? I'm not sure there's any practical way to prevent styles being inherited by child components (or that you'd want to, ultimately) - this fix was about making sure that at the very least the inheritance behaviour is consistent (previously whether or not styles were applied was dependent on whether components lived at the top level of another component's template, which was definitely wrong)

martypdx:
> It was the case, briefly, that all elements were marked

ah, I see.

Issue #1720, opened 2015-02-16
Inconsistencies


dagnelies:
Hi,

http://jsfiddle.net/cnopeqw7/1/

The data gets updated for both `foo` and `bar`  but does not propagate into the view.

Before telling it's a problem of propagating the change in the view, there is a broader question:
*What is the expected behavior of*
`ractive.set('bar', ractive.get('foo'))`
*To make a "copy" of `foo` or a "reference" to `foo`?*

Currently, it considers it a reference and updating one's data updates the other's data as well. However, the view is not properly updated, probably because it has no notion of references. This can be circumvented by calling `update()` at the right places but it feels very hacky.

simongfxu:
i consider this as a bug.

evs-chris:
In this case, it's neither a copy nor a reference. Ractive is not tracking anything by object identity internally, so this is one of the places it can't keep track of changes for you. I've had a similar issue in the past, and I've been thinking about proposing an API for doing something like a symlink within the viewmodel. In my case, I have something like a master/detail list, where one object is focused and changes automatically with selection. I don't have any code yet, but this is roughly how it would work:

```html
{{#each items}}<li on-click="link(@keypath, 'current')">{{.name}}</li>{{/}}
{{#current}}
  <input value="{{.name}}" />
{{/}}
```

@dagnelies would that cover your use-case? I don't know if it would be reasonable to try to use object identity to try to manage this or some other object tracking automatically, especially with large data.

Under the covers, `link` would create a mapping that references the current viewmodel on both sides with only the keypath changing. The reciprocal `unlink` would remove the mapping, and `link`ing an existing mapping would overwrite it. Does that sound reasonable, or have I grossly underestimated what it would take to make this work?

dagnelies:
@evs-chris I think your `link` proposal totally makes sense. It's clear and straightforward. I'd "+1" it ;)

Moreover, I'd go one step further and provide a way to either link or copy, rather than the current way which does "something" resulting in inconsistent view/data.

evs-chris:
I looked into a `link`/`unlink` a little, and with the way mappings work right now, links would only be allowed as top-level keys e.g. `ractive.link('foo.bar', 'baz.bat')` would not be legal because the target has more than one level in the keypath. Working around that would get expensive, but just being able to link root level keys would be worth it for me. I'll put together an RFC PR and see what happens.

[...]

PR #1722, opened 2015-02-17
Implementation of link/unlink for mapping keypaths within a viewmodel - RFC

evs-chris:
This ~~takes advantage of the work done on mappings~~ to alias one keypath to another root (symlink) within a single viewmodel. This is particularly handy for master/detail-type scenarios where you want to do something like:
```html
<ul>{{#each items}}<li on-click="link(@keypath, 'selected')">{{ .name }}</li>{{/each}}</ul>
<label>Name: <input value="{{ selected.name }}" /></label><br/>
<label>Priority: <input type="number" value="{{ selected.priority }}" /></label>
```

It's even more convenient when the data is hierarchical. With this changes to either the source or destination are reflected in both places as long as the link is in place. It introduces two additional methods on both the Ractive ~~and ViewModel~~ prototypes.

```js
Ractive.link( source /*: string */, dest /*: string */ ) /*: Promise */
Ractive.unlink( dest /*: string */ ) /*: Promise */
ViewModel.link( source /*: Keypath */, dest /*: Keypath */ )
ViewModel.unlink( dest /*: Keypath */ )
```

If the `dest` is already mapped, it will be remapped. If there is already a "real" value for `dest` it will be shadowed until it is unlinked.


evs-chris:
I've been working around this for a while by manually `update`ing shared keypaths for a while, but someone other than me (#1720) ran into it and I figured I'd take a swipe at a cleaner solution.

martypdx:
Freakishly clever @evs-chris :)

martypdx:
Can you hose up component mappings if you're not careful?

evs-chris:
I suppose you could... I didn't really consider that. But not anymore!

evs-chris:
Also forgot to mention that, due to the way mappings work, link `dest`s can only be at the root. Once @martypdx gets the keypath deps and keypath aliases implemented :wink:, deeper keypaths could be allowed.

martypdx:
Hey @evs-chris, I thought about this some more and I wonder if it would work better to register a dependency. Something like:

```js
function link( there, here ) {
    this.register( here, function( value ) {
        this.set( there, value );
    }.bind(this) );

   this.register( there, function( value ) {
        this.set( here, value );
    }.bind(this) );
}
```

But with some tracking for teardown, unlinking, etc.

That way it works with any keypaths, including across component boundaries.

martypdx:
btw - one way I've found that gives you the two way binding is to use an index with a keypath expression (http://jsfiddle.net/ptjadb5d/):

```handlebars
    {{#items:i}}
      <li on-click='set("selected", i)'>{{name}}, {{job}}
    {{/}}
    {{#with items[selected] }}
      <input value="{{ name }}"/>
      <input value="{{ job }}"/>
    {{/with}}
```

dagnelies:
This sounds like a useful feature. However, I wanted to dig a bit more about how it's meant to work.

Let's assume `link('foo','bar')`

Whenever `foo` or `bar` is updated, in an input field for example, both are updated, right? Or is it just "one-way" which would be kind of catastrophic.

Does it also work within/accross components?

> If the dest is already mapped, it will be remapped. If there is already a "real" value for dest it will be shadowed until it is unlinked.

IMHO, I think this would be an unexpected behavior. More expected would be to either null it or place a copy of the data.

Rich-Harris:
This looks *awesome*. I'm under the cosh so can't look at the implementation right now, but in the meantime

evs-chris:
I just added an implementation based on @martypdx's suggestion to just register dependencies. I modified `ViewModel.applyChanges` a little ( supply a keypath to `setValue` and allowing refining bindings to match based on a regex ) to make it easier. This version allows non-root links and does not shadow data, since it does a shallow copy when the link is set up. It also sets the link keypath to undefined on `unlink`. The mapped implementation is still in place for comparison.

@martypdx That's a handy method I've actually used before, but it falls over with hierarchical data. I had to fall back to manual updates and/or observers on the project I'm working on right now (trees).

@dagnelies It should work as expected when used with twoway bindings (see the tests). It should also work as expected across/within components since it basically just triggers `update`s when either side changes. The mapping version should work as well, as long as the link and component mappings don't overlap.

evs-chris:
Superceded by #2172

Issue #1721, opened 2015-02-16
tricky bug with ractive.set()


dagnelies:
Hi,

This happens using a line like `ractive.set('foo', null)`

The bug is involved in rather complicated stuff, dunno if I can transform it in a simple example. In the meanwhile, here is the stack trace:
```
Uncaught TypeError: Cannot read property 'firstKey' of undefined
unregister.js:10 Viewmodel$unregister
Mapping.js:140 unregister
unregister.js:11 Viewmodel$unregister
unbind.js:4 unbind__unbind
unbind.js:9 Section$unbind
unbind.js:12 unbindItem
unbind.js:6 Fragment$unbind
unbind.js:6 Element$unbind
unbind.js:12 unbindItem
unbind.js:6 Fragment$unbind
unbind.js:15 Component$unbind
unbind.js:12 unbindItem
unbind.js:6 Fragment$unbind
Yielder.js:82 Yielder__unbind
unbind.js:12 unbindItemunbind.js:6 Fragment$unbind
unbind.js:6 Element$unbind
unbind.js:12 unbindItem
unbind.js:6 Fragment$unbind
unbind.js:15 Component$unbind
unbind.js:12 unbindItem
unbind.js:6 Fragment$unbind
unbind.js:15 Component$unbind
unbind.js:12 unbindItem
unbind.js:6 Fragment$unbind
unbind.js:6 Element$unbind
unbind.js:12 unbindItem
unbind.js:6 Fragment$unbind
unbind.js:15 Component$unbind
unbind.js:12 unbindItem
unbind.js:6 Fragment$unbind
methodCallers.js:2 methodCallers__unbind
setValue.js:311 removeSectionFragments
setValue.js:304 reevaluateConditionalSection
setValue.js:142 reevaluateSection
setValue.js:48 Section$setValue
applyChanges.js:179 (anonymous function)
applyChanges.js:179 dispatch
applyChanges.js:151 notifyAllDependants
applyChanges.js:83 Viewmodel$applyChanges
runloop.js:107 flushChanges
runloop.js:34 runloop.end
set.js:41 Ractive$set
```

...I have no idea what's causing it, so I don't really know in what direction I should begin investigate. I just know several things have to happen before it occurs. More precisely, render the component, interact with the data within and remove it, then the bug occurs. Just by rendering it and removing it, nothing happens. The component is rather complex, and I couldn't boil it down / isolate it so far.

PS: using edge version
[...]

PR #1724, opened 2015-02-17
Guard for undefined keypath

dagnelies:
Fix for #1721
The `keypath` was undifined at this point, causing an uncatched exception/error.
...please note that I don't really know what I'm doing here. I just notice it works that way.


martypdx:
It's failing in the Mapping class, so let's try to put the guard clause there, not in the unregister method:

```js
map ( keypath ) {
    if( typeof this.keypath === undefined ) { return this.localKey; }
    return keypath.replace( this.localKey, this.keypath );
}
```

PR #1731, opened 2015-02-18
Keep expression mangling for caching from messing with non-reference _things

### LINKED ISSUE #1728: Complex referencing in mustache binds to a wrong variable if paths are the same but different digits present
I'm not sure if this have been reported yet:

In these two mustache expressions
`{{ local.metrics["metric_td_lm2_3_result"].passed }}`
and
`{{ local.metrics["metric_td_lm2_10_result"].passed }}`

latter would show exactly the same value as first, and I'm pretty sure it's because of internal cache handlers that most likely drop the digits of the keypaths... 

Workaround for now is to change signature slightly by adding a char or two:

http://jsfiddle.net/zLmp7wka/ Demonstrated here

evs-chris:
Fixes #1728

This keeps the unique string generator from replacing '_[0-9]+' with the undefined if the number is greater than number of references for the expression. Is that the most appropriate fix?

### DIFF
--- a/src/virtualdom/items/shared/Resolvers/ExpressionResolver.js
+++ b/src/virtualdom/items/shared/Resolvers/ExpressionResolver.js
@@ -119,6 +119,11 @@ function getUniqueString ( str, keypaths ) {
 	return str.replace( /_([0-9]+)/g, function ( match, $1 ) {
 		var keypath, value;
 
+		// make sure we're not replacing a non-keypath _[0-9]
+		if ( +$1 >= keypaths.length ) {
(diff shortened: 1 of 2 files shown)


martypdx:
> if the number is greater than number of references for the expression

Had a question, is it still borked if the number is _less than_ the number of references?

evs-chris:
Yes, but only if there are references that are the same. If the refs are different, then the `uniqueString` will still be mangled, but it should at least still be unique. Subbing in `undefined` everywhere killed the uniqueness and caused the staleness issue.

PR #1734, opened 2015-02-19
Guard for undefined keypath (bis)

dagnelies:
Hi,
The keypath was undifined at this point, causing an uncatched exception/error.
This is the alternative fix for #1721, more general than the previous fix  #1724.
Cheers,
Arnaud


evs-chris:
:+1:

Issue #1438, opened 2014-11-04
Unbinding event method calls throws an error when removing a component from the DOM (and possibly other scenarios)


Madgvox:
jsFiddle repro: http://jsfiddle.net/xxg93vw8/6/

If you toggle it to show and then try to toggle it to hide, an error is thrown.

I haven't tested it as thoroughly as I usually like to, but I know that the error is being caused by the `toggle` call in the component.

Issue #1390, opened 2014-10-21
Setting value of the Checkbox Component


krasimir:
Hello,

I have dynamically build checkboxes. Here is the code:

``` js
(new Ractive({
    el: 'body',
    oninit: function() {
      this.set('options', [
        {value: 1}, 
        {value: 2},
      ]);
      this.set('ch', [ {value: 1} ]);
    },
    template: '{{#each options}}\
      <input type="checkbox" name="{{ch}}" value="{{value}}">\
      {{/each}}\
      {{ch}}'
})).observe('ch', function (n) {
    console.log(JSON.stringify(n));
});
```

So, what is the syntax to make some of the boxes checked. Here is a JSFiddle:http://jsfiddle.net/6n5tg/

Issue #783, opened 2014-06-04
Possibly a bug with `checked` attribute in radio button


kenvunz:
Hi,

I've found this odd behaviour, not sure if it's a bug or I'm missing a step

http://jsfiddle.net/kenvunz/32xkw/

Step to produce:
- Click to select `b` radio button
- Then hit Update

Actual behaviour:
- None of the new rendered radio buttons are 'checked'

Expected behaviour:
- Radio button `d` should be 'checked'

Issue #460, opened 2014-03-01
0.4.0 — {{ items.filter() }} issue when check items's checkbox in Ractive TodoMVC


julien1441:
In the Ractive todoMVC example (https://github.com/Rich-Harris/Ractive-TodoMVC), you have this:

**Template**

``` html
<span id="todo-count">
   <strong>{{ items.filter( active ).length }}</strong> {{ items.filter( active ).length === 1 ? 'item' : 'items' }} left
</span>

{{# items.filter( completed ).length }}
   <button id="clear-completed" on-tap="clear_completed">
      Clear completed ({{ items.filter( completedFilter ).length }})
   </button>
{{/ end of filter }}
```

**todoList.js**

``` javascript
data: {
   // define our filters
   completed: function ( item ) { return  item.completed; },
   active:    function ( item ) { return !item.completed; },
}
```

The problem is when I check a task: the "clear completed" button doesn't appear and the _"x item left"_ is not updated (unlike here: http://todomvc.com/architecture-examples/backbone/).
The item's checkbox (item.completed): `<input class="toggle" type="checkbox" checked="{{completed}}">`

Whereas when I check the global checkbox (check all task), all is well updated instantly.
**todoList.js**

```
// global checkbox
toggle_all: function ( event ) {
   var i = this.get( 'items' ).length, completed = event.node.checked, changeHash = {};

   while ( i-- ) {
      changeHash[ 'items[' + i + '].completed' ] = completed;
   }

   this.set( changeHash );
}
```

PR #1740, opened 2015-02-19
WIP: Keypaths FTW

martypdx:
This is the start of moving viewmodel functionality into Keypath objects. At this point, viewmodel doesn't have caches anymore for values, computations, and wrappers - and the get and clearCache functionality have also been moved into the Keypath objects.

This means not having to do multiple hash lookups on each get. And it also begins to consolidate the logic around managing a particular keypath value.

Just the start of the process, remaining would be
* move other viewmodel methods
* optimize mappings by reusing Keypath object when keypath is the same or aliasing
* move dependency tracking
* I think reference resolving can be improved as well


evs-chris:
I like where this is headed! Will this let us drop the uniqueString-related stuff with caching swapping to keypaths? (Looks like it will, but I'm a little fuzzy right now) If so, +1000!

martypdx:
> Will this let us drop the uniqueString-related stuff with caching swapping to keypaths

It should because each Keypath object points to a discrete graph location that may use different strategies to get value like refer to something in `data`, or an adapter wrapper, or a virtual value like an array index accessible at a keypath like `items.*.@index`. 

This later category of `st@te` is a cool possibility to arbitrarily define data values, at specific contexts, that don't need to go in the model:

```html
{{#items}}
<li class='todo'>{{.}}</li>
{{/}}
<input value='items.@new'>
<button on-click='push("items", @new)'>Add</button>
```

evs-chris:
That is pretty slick!

Rich-Harris:
This is exciting stuff (not that I fully understand it yet... working on that). We might have to stop calling them keypaths though!

martypdx:
> We might have to stop calling them keypaths though!

Yeah I've wonder that. These new things are like GraphNodes, ModelNodes, something like that. But I think it best to see how things shake out and how much string manipulation will even be required for keypaths. The primary efficiency for strings will need to be between template interpretation and api calls that use strings into keypaths - maybe memoize or maybe just a map based on last key or something. We might need to still have keypaths purely as string encapsulation like we have currently.

martypdx:
More progress. Mappings are gone (the source keypath is directly used, or an alias is created if the keypath is different). Dependency tracking has moved into Keypaths (no more dependency maps). Indexes, reference expressions and arrays are hosed until some things get fixed.

Beginning to see some of the performance benefits, though that has been an explicit goal thus far. Explicit "mappings" are much quicker as is the "box" updating test.

martypdx:
UPDATE:

So I pushed some more stuff on this and probably time to tackle some explaining of where this is headed. The biggest change has been that the virutaldom items get a permanent handle to the part of the model it is interested in, and this is durable throughout its lifetime. The viewmodel then becomes a virtual model for the virtual dom. :)

That means no more rebinding in the virtual DOM, and special references and expression are fully handled by the viewmodel. Because the virtual viewmodel is stored in a tree that mimics the data, there's built in dependency mappings for up and down stream changes, and it makes notification simpler and less redundant.

Currently in the code, the Model class refers to one part of the viewmodel as a whole. There are some specialized models, like RootModel, ComputationModel and ReferenceExpressionModel (but because Sections can change type, there's no dedicated models based on types).

Mappings are simply a matter of referencing the Model from another viewmodel within the current viewmodel. Performance was one of the key concerns - I started this because of component mappings. You can see huge gains in update perf test (EDGE is keypath-ftw, CONTROL is current edge):

image

and init perf test:

image

I added a render iterate comparison - inline vs partial vs implicit component vs explicit component:

image

Generally 20-50% faster. Though I'm not seeing the component gains yet across the board, I haven't had a chance to profile and see what's up. You can see in this render test (based on update test):

image

That actually losing a small bit time for components, though gains in boxes.

I'm only up about 200-300 LOC at this point, but I haven't deleted out any of the rebinding and other code that is no longer needed.

The code still has a lot of messes lying around and biggest thing left to hook up is smartUpdate (and observers). And there's more I need to write up. But I wanted to put this out there, it looks pretty exciting.

martypdx:
Oh, on the iteration updates (boxes), I'm recreating all the fragments since no smart updates - so that should go down once those are in place.

Rich-Harris:
Holy moly! This is bloody great. Haven't had a chance to look at the code but those are some amazing numbers. I can't wait to repeat that dbmonster experiment!

Does this entail any changes to the public API, or is this totally internal?

> But I wanted to put this out there, it looks pretty exciting

That's something of an understatement

Rich-Harris:
Does it feel realistic to include this in 0.7?

martypdx:
> Does this entail any changes to the public API, or is this totally internal?

99% internal.

> Does it feel realistic to include this in 0.7?

No, I was thinking 0.8. Too big a change to not let people thrash on it in edge for awhile, and it's not even complete yet.

martypdx:
I've been continuing to chip away at this. I did quite a bit of work on the ReferenceExpressions (the idea of references is used to handle more such as Member references like `{{foo.0}}` that may have the actual array member change out from under them - will write up more on that when I get a chance).

And I got the array list sections up and running. From above, recall that arrays were in regression, now they're starting to shine and the splice array perf test now works:

image

image

(I'd ignore that partial result, I probably goofed something there - I can't see how it would be faster than inline).

Rich-Harris:
> Also, I tried db-monster and it runs about 40% faster.

SOLD.

This is clearly the most important piece of working happening on Ractive at the moment, and I'm conscious that changes of this magnitude are tricky to co-ordinate with other work. So if we need to down tools on other issues and concentrate on getting this in without causing merge conflict-o-rama, just say the word.

Likewise if there's anything any of us can do from a vantage point that's external to your brain, do shout...

martypdx:
> So if we need to down tools on other issues and concentrate on getting this in without causing merge conflict-o-rama, just say the word.

I did a merge at `0.7` release, so more or less up to date. Basically viewmodel would be the thing to avoid

> Likewise if there's anything any of us can do from a vantage point that's external to your brain, do shout...

Two things around the new "classes" in the viewmodel:

First is naming. I moved from `Keypath` to `Model`, but now I think they're really `Context`. Essentially the viewmodel constructs a tree of objects based on requests from the template and api that provide an interface into the data. A virtualmodel to go with the virtualdom :). One goal has been to eliminate rebinding from the virtualdom by providing a single, durable entry point and handling data changes from within the viewmodel. Naming question is about what to call this entry point.

The term `Binding` is already used in the virtualdom to refer to form inputs that set data. The term `Context` is used to refer to the "in-scope" object, though it's not a big leap to think of every mustache having a context - the piece of the viewmodel it's tied to. So, I'm leaning towards calling the base class `Context`. I thought of maybe `BindingContext` but that doesn't seem quite right.

Let me know what you think. Here are the list of current types, assuming that they would be contexts (these are all derivates of `Context`):
* `Context` - base class that provides bulk of functionality. Handles both objects and arrays, though support for the later is more explicit than in current viewmodel.
* `RootContext` - the root of the viewmodel, mostly some slightly different handling of being parentless
* `ComputationContext` - a context that is a computation (expression or computed)
* `ReferenceContext` - a context that tracks it's own dependants and children, but delegates to another context for handling get, set, shuffle, etc.
* `DynamicReferenceContext` - a reference context that's tied to another context to provide a value for what property of the it's parent it is (aka ReferenceExpression)
* `MemberContext`-  a reference context specialized for explicit references to array members like `{{foo.0}}`. It allows the actual array member to be changed while maintaining a durable end point for binding
* `UnresolvedContext` - a proxy to hold registrations and child properties until resolved to a "real" context
* `Index` - a specialled context for handling `@index`. 
* `Key` - same as above for `@key`

Second issue is related to not being able to have bidirectional module dependencies.  The context tree is recursive and there are derivate classes spawned by parent classes, so I'm having to put all those classes in the same module. (Concrete example: a normal Context in the join method will create a MemberContext if the key is numeric. MemberContext depends on Context). 

We could have all constructors plus methods that construct new objects in one file, and put the prototype methods elsewhere. So it's not a big deal, but wanted to ask if there's a way to get it to work. (Side note, I've been using the es6 classes which have been a joy to write, but plan on reverting to es5 classes before all is said and done).

Rich-Harris:
`Context` seems about right. I was starting to think of it as a `DataGraph` full of `DataNodes` which came in various flavours (`ComputationNode` etc), but 'node' is a loaded word in a DOM library whereas 'context' is already used to mean something related.

Having said that what do you call the leaf nodes, i.e. that represent primitive values? 

Either way, the less bikeshedding the better :)

> Second issue is related to not being able to have bidirectional module dependencies.

I'm surprised that isn't working - with ES6 modules it should be fine. It should only break when you do something contrived and nonsensical like `class Foo extends Bar {...}` in one module and `class Bar extends Foo {...}` in another. I couldn't quite figure out what was going on from a quick look at the code just now, but one possibility would be to use functions instead of classes.

> I've been using the es6 classes which have been a joy to write, but plan on reverting to es5 classes before all is said and done

Notwithstanding what I just said, are ES6 classes problematic? With babel's [loose mode](https://babeljs.io/docs/usage/loose/) they transpile sanely, as long as you're not doing anything wild. I'd been wondering about whether to use them more widely.

evs-chris:
Very exciting! I feel less bad that I'm swamped right now, since it also means I'm not making further merge conflicts.

FWIW, I'm cool with `Context`. I thought I'd throw out some other color options for the shed too with `Handle` and `Pointer`, as they seem to fit the viewmodel <-> virtualdom intermediary vibe.

I've also had pretty good luck with ES6 classes and babel. babel has me spoiled to the point that my fingers don't want to type non-ES6 anymore. I can't wait for the mallet operator and some sort of null-coalescing operator to come along.

martypdx:
> Having said that what do you call the leaf nodes, i.e. that represent primitive values?

Same thing `Context`. From a binding perspective there's no difference.

> I'm surprised that isn't working - with ES6 modules it should be fine. It should only break when you do something contrived and nonsensical like class Foo extends Bar {...} in one module and class Bar extends Foo {...} in another. I couldn't quite figure out what was going on from a quick look at the code just now, but one possibility would be to use functions instead of classes.

Here's the setup to try:

```js
// file 1:
import MemberContext from './MemberContext';
class Context () {
    method () {
        new MemberContext();
    }
}

// file 2
import Context from './Context';
class MemberContext extends Context {

}

exports default MemberContext;
```

> Notwithstanding what I just said, are ES6 classes problematic?

One concern would be use of `Object.getPrototypeOf` and `Object.getOwnPropertyDescriptor`. I assume there's a polyfill for those, but also concerned about overhead of those calls (easy enough to test I suppose). 

I pulled out the babel code so we could see it more clearily and put it [here](https://github.com/ractivejs/ractive/pull/1740/files#diff-78992a831a325a00b08cd8c36c0e2c5a)

Rich-Harris:
Ah, yes. I see the problem. I think it's fixable though: https://github.com/esperantojs/esperanto/issues/152

Babel has a loose mode, which transpiles classes slightly differently - it's not 100% spec-compliant but the resulting code executes faster. We just add `loose: [ 'es6.classes' ]` to the babel options in `gobblefile.js`.

martypdx:
Latest commit moved object hash iteration logic into viewmodel (as opposed to section). 

It's better on update of whole object hash and adding a key. But it's not performing as well on render, so need to look at what's going on there. I moved the list logic into the viewmodel, and wrote new version of shuffle but it's not hooked fully hooked up yet to Section.

The other thing that needs worked out is switching aliases when moving between a list each block and hash each block. Originally I had aliases living in the viewmodel context stuff, but there's some edge cases where it doesn't make sense for those to be applied to the data globally. I moved them into section (there's an EachBlock class now - will probably add others) which works great to scope the specials aliasing to the block - but need to figure out how to swap them. In an architecture where binding context is permanent, it's a more challenging aspect!

Rich-Harris:
> The other thing that needs worked out is switching aliases when moving between a list each block and hash each block

Maybe we can save ourselves the effort - I'm fairly sure it's a mythical use case. If someone is actually doing that it's almost certainly unintentional! I think the only time we need to change section type dynamically is probably going from falsy (unresolved, `null`, whatever) to list or hash.

martypdx:
I came to the same conclusion last night. If that happens we can just rerender fragments.

martypdx:
All render tests are now passing. :)  Though I need to switch to the new shuffle code, which works in tests in isolation - just need to wire into viewmodel shuffle and Section

martypdx:
More good news. I wired up the new "shuffle" array methods, check out the splice perf test:

image

Insane. Had to check in sandbox to make sure it was actually working (it is). 

I went back to [this SO question](http://stackoverflow.com/questions/28207737/frequent-data-update-for-ractive-js#comment45976508_28207737) which is a rolling log of 100 items. Here's update times for `0.7.2` (which already is faster than when I ran this initially) with the two ops `pop` and `unshift` (merge isn't working in new branch yet):

image

So hovering just under `10ms`. 

Now same thing in `keypaths-ftw`:

image

`1ms` for a `pop` and an `unshift`! 10x faster.

I had all the render tests passing, but broke some when I got the new array methods working, so I'll fix those before checking in...

Rich-Harris:
excited-geese

guilhermeaiolfi:
giphy

How many tests don't pass? Just so I can have an idea of how much work it needs before we can test it in our apps.

martypdx:
@guilhermeaiolfi:

These are test modules with fails. Total is ~321 of 2633

module | fail | total | notes
---- | ----- | --- | -----
adaptors | 13 | 18 | expected adaptors not enabled
animate | 3 | 3 | all with same error
arrayMethod | 5 | 44 | pattern observer test, 1 failed test
arrays | 15 | 38 | adaptors not enabled
componentData | 21 | 97 | haven't circled back to components
components | 13 | 63 | haven't circled back to components
computations | 4 | 33 | around set operations
decorators | 1 | 20 |
events |  23 | 183 |
find | 10 | 35 |
get | 2 | 2 |
magic | 15 | 20 | adaptors not enabled
merge | 18 | 32 | expected, merge not reimplemented
misc | 35 | 136 | 
observers | 49 | 76 | all pattern observers tests
partials | 5 | 76 |
rebind | 14 | 32 |
render | 24 | 432 | just rebroke these today
reset | 1 | 33 |
select |11 | 55 | expect two-way needs some updating
transitions | ? |  | dies after third test
twoway | 18 | 96 | expect two-way needs some updating
yield | 1 | 14 |

Rich-Harris:
@martypdx There are a couple of commits on the kftw-staging branch that ought to fix animations and two-way binding - 71f57f6a37b49c2eebad3ced0669ece3326e8afc and 30c4fa34e2c1d18754c893e99af16fd0c61862f9. Should be able to pull those directly into this branch

JonDum:
I understand about 40% of what's going on in this thread, but those perf numbers are insane!! Great job, @martypdx!! :+1: :+1:

Rich-Harris:
@martypdx I've just released a new version of esperanto (0.6.30) which is slightly smarter about module ordering within a bundle - if `class A extends B` and `B` has a method that references `A`, `B` is guaranteed to come first. So it should be possible to move some of those classes out of `Context.js` now - I tried it just now and it appears to work (having failed with 0.6.29). Can do a commit but I suspect we've diverged, so would probably just cause conflicts.

martypdx:
> @martypdx I've just released a new version of esperanto (0.6.30) which is slightly smarter about module ordering within a bundle

Thx, I'll try it out!

martypdx:
> @martypdx I've just released a new version of esperanto (0.6.30) which is slightly smarter about module ordering within a bundle
> > Thx, I'll try it out!

`gobble-ractive` doesn't appear to be updated, so do I need to do a local build of it? Or something else?

martypdx:
@Rich-Harris i did that locally and got newer version - but doesn't seem to resolve correctly. I've pushed commit to this branch with the separated files. `Reference` is being called before `Context` is available as super.

Rich-Harris:
> gobble-ractive doesn't appear to be updated, so do I need to do a local build of it? Or something else?

Did you mean gobble-esperanto? If so, neither it nor gobble-esperanto-bundle actually install esperanto themselves, because esperanto is required as a devDependency in its own right (yeah, I know...), so you should only need to update that. I've updated `package.json`, so `rm -rf node_modules ; npm i` should bring things up to date - not that it fixes things (yet). I suspect the algorithm for refining the module order might still need to be a touch more sophisticated.

Rich-Harris:
> I suspect the algorithm for refining the module order might still need to be a touch more sophisticated.

FYI @martypdx I've opened a new esperanto issue to try and get to the bottom of this cyclical dependency nonsense https://github.com/esperantojs/esperanto/issues/159

martypdx:
Thanks @Rich-Harris. Been slammed last couple of weeks, finally a few hours this morning to get back to this. Fixed all render tests, just merged `kftw-staging`. Would be good to fix whatever is keeping the test suite as a whole from running.

As far as the class ordering issue, let me know if you need anything from me on that.

Rich-Harris:
> Would be good to fix whatever is keeping the test suite as a whole from running.

Do you mean the way it hangs early on with no error? In the past I've tracked similar bugs down to [here](https://github.com/ractivejs/ractive/blob/dev/src/Ractive/prototype/render.js#L70-L72) - the auto-teardown fails for some reason, but it's sometimes hard to figure out which test is responsible (because it's a later test that fails). I haven't managed to suss this one out though

martypdx:
Okay, with your changes from `kftw-staging` things are better, but still hangs. If I comment out the `transitions` tests - the whole suite runs. Yet if I run just transitions.html tests, they all pass. :(  Those are the worst...

> Do you mean the way it hangs early on with no error? In the past I've tracked similar bugs down to here - the auto-teardown fails for some reason, but it's sometimes hard to figure out which test is responsible (because it's a later test that fails). I haven't managed to suss this one out though

Exactly! Probably something async going on...

Rich-Harris:
I've implemented what seems like a better module sorting algorithm - https://github.com/esperantojs/esperanto/pull/160 - so it *should* be possible to split `Context.js` apart again (I tried building 95fb936ba660b4bb37f131854bf9c82caab08e74 and it all looks good). I haven't released it yet as I wanted contributors to have a chance to say 'stop! that looks idiotic', but I will soon

martypdx:
### Status

#### Most functionality in-place except:

* No Adaptors (incudes array and magic, so ractive API only)
* No Pattern Observers
* No binding to .length property
* Mostly no .merge()
* Component live queries do not maintain order
* A few edge cases around Computed Properties and Components

#### Breaking Changes

* Forced Resolution binds to nearest context, not root
* Possibly a few of the tests below might lead to changes instead of fixes

#### Failing Tests Not Covered From Above

* Issues: event references in method call handler should not create a null resolver (#1438)
* get(): Returns mappings on root .get()
* Miscellaneous: Subclasses of subclasses inherit data, partials and transitions
* Miscellaneous: Multiple identical evaluators merge
* Miscellaneous: .unshift() works with proxy event handlers, without index references
* Miscellaneous: Bindings without explicit keypaths can survive a splice operation
* Inverted sections aren't broken by unshift operations
* Splice operations that try to remove more items than there are from an array are handled
* Miscellaneous: Regression test for #271
* Miscellaneous: Regression test for #297
* Miscellaneous: Two-way binding can be set up against expressions that resolve to regular keypaths
* Miscellaneous: Regression test for #460
* partials: partial functions selects same partial until reset
* partials: Named partials should not get rebound if they happen to have the same name as a reference
* Radio name inputs respond to model changes (regression, see #783)
* Two-way bindings: Changes made in oninit are reflected on render (#1390)

### Performance

Performance gains are holding across the board, rendering being 5-30% faster except object hash iteration which is marginally slower. Updating is much faster across the board in `0.8`:

image

image

Rich-Harris:
Awesome, thanks for the update. Looks like there's light at the end of the tunnel. Hopefully I'll get a chance to chip in over the next few days. I released esperanto@0.6.34 by the way, if that's helpful

martypdx:
>  I released esperanto@0.6.34 by the way, if that's helpful

Does that handle the circular class and refs?

Rich-Harris:
@martypdx yep, exactly

NikhilVerma:
Sorry to nag but there haven't been any updates to this PR for a while. Just curious on it's progress. I would really like the performance improvements offered by this. 

I am in the process of migrating a large mobile webapp from conventional dust templates to ractive templates. I'm absolutely in love with everything that Ractive has to offer however the general app start speed has been slowed down by around 30% (I am using pre-parsed templates). The biggest problem has been the absolutely huge increase in call stack depth due to recursion during template rendering.

I was wondering if this ticket might fix that?

martypdx:
@NikhilVerma:

> Sorry to nag but there haven't been any updates to this PR for a while. Just curious on it's progress. I would really like the performance improvements offered by this.

Been making commits on the weekends. :) Just about done with merge issues, then just need to fix some issues with twoway binding on `update`, and adaptors need to be fixed.

> The biggest problem has been the absolutely huge increase in call stack depth due to recursion during template rendering

Are you saying "stack depth" itself is an issue on mobile browsers with ractive? Or just more broadly that first render is slow(er)?

> I was wondering if this ticket might fix that?

Hard to say with more detail.  But you're welcome, and encouraged, to try it out. Rendering is solid, so it shouldn't be too hard to get an idea if it's any faster. If you clone, do `npm init` and then `npm run build` it should populate the `build` folder.

NikhilVerma:
@martypdx I just did And WOW it made a huge difference :)

Now a Ractive build of the webapp loads up 10% faster than the DustJS version! (Compared to 30% slower). Might even do some thorough testing and see how it runs on a production environment.

I mentioned stack depth because that's the biggest difference I noticed from the profiler. But obviously using static templates vs reactive ones is like comparing apples to androids so it's tricky to compare these things.

Thanks for all the effort you've put in this pull request. I love ReactJS simply because it's so similar in approach to other templating libraries, so it's a lot more easier to migrate existing projects to it compared to ReactJS.

martypdx:
> Now a Ractive build of the webapp loads up 10% faster than the DustJS version! (Compared to 30% slower). 

You just made my day. :grin:

fskreuz:
@NikhilVerma Probably related to this one: https://github.com/ractivejs/ractive/issues/1905

NikhilVerma:
Hey @martypdx ! Just wanted to check once again. If you have made any progress on this being merged into the edge version sometime? :) Does #2022 complicate things for you?

Issue #1733, opened 2015-02-19
Recent bug: "erm wut"


dagnelies:
Hi,

Here is the fiddle:

http://jsfiddle.net/v16b6vcj/1/

Basically, empty component arguments are causing the error.
...this is a very recent bug, it appeared at most in the last 2/3 days.

It's thrown here:

https://github.com/ractivejs/ractive/blob/dev/src/virtualdom/items/Component/initialise/createInstance.js

This occurs during initialization. Here is the full stack trace:
```
 Uncaught Error: erm wut
 createInstance.js:98 (anonymous function)
 createInstance.js:44 createInstance
 init.js:25 Component$init
 _Component.js:16 Component
 init.js:55 createItem
 init.js:31 (anonymous function)
 init.js:28 Fragment$init
 Fragment.js:20 Fragment
 initialise.js:62 initialiseRactiveInstance
 Ractive.js:18 Ractive
```

dagnelies:
Ok, found it. Updated.

Issue #1727, opened 2015-02-17
Boolean attributes in components


Rich-Harris:
For regular elements, an attribute without a value is interpreted as 'true'. For components that's not the case - I'd argue that it should be http://jsfiddle.net/rich_harris/nk649buL/

martypdx:
:+1:

MartinKolarik:
Are we talking about boolean attributes (checked, readonly, ...), or all attributes without a value? A non-boolean attribute without a value is interpreted as an empty string for regular elements.

martypdx:
My understanding is that this is would be limited to:
* component tags only
* missing values `<widget foo/>`, not empty values `<widget foo=''/>`

Rich-Harris:
> component tags only

Yes. I suppose if you wanted to be precise about it, the component would have to declare which values were supposed to be boolean... though in practice I can't think of a situation where you'd want an empty component attribute to equal the empty string (@MartinKolarik unless I'm overlooking something?). At the moment I'm writing `<foo bar='true'/>` which I don't love.

MartinKolarik:
>@MartinKolarik unless I'm overlooking something?

You probably aren't. It's a little counterintuitive that `<widget foo />` and `<widget foo="" />` don't mean the same thing, but I can live with that.

PR #1741, opened 2015-02-20
Various attribute handling adjustments

evs-chris:
Fat fingered the first submission. Meh.

### DIFF
--- a/test/__tests/components.js
+++ b/test/__tests/components.js
@@ -1,5 +1,3 @@
-import Model from 'helpers/Model';
-
 module( 'Components' );
 
 test( 'Components are rendered in the correct place', t => {
@@ -21,7 +19,7 @@ test( 'Components are rendered in the correct place', t => {
 });
 
 test( 'Top-level sections in components are updated correctly', t => {
-	var ractive, Component, component;
+	var ractive, Component;
 
 	Component = Ractive.extend({
 		template: '{{#foo}}foo is truthy{{/foo}}{{^foo}}foo is falsy{{/foo}}'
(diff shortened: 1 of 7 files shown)


evs-chris:
This should address #1733 and #1727. Attribute handling is a little bit fiddly, but I think this balances everything fairly well.

* Empty string attributes now keep the empty string instead of becoming 0. That should cover #1733
* Valueless attributes still come out as 0, but components treat that as true now.
* contenteditable elements with no value binding don't get a bonus `undefined` in their `toHTML` output (stumbled across while assembling a test)

Does that sound reasonable?

Sorry about the noise in the test files. I delinted them.

Rich-Harris:
Yep, this looks great. Ready to merge?

test/__tests/components.js
```diff
@@ -830,3 +829,23 @@ test( 'Component mappings used in computations resolve correctly with the mappin
 
 	t.htmlEqual( fixture.innerHTML, ractive.toHTML() );
 });
+
+test( 'Component attributes with no = are boolean true', t => {
+	new Ractive({
+		el: fixture,
+		template: '<cmp foo />',
+		components: { cmp: Ractive.extend({ template: '{{#foo}}yep{{/}}' }) }
```
martypdx: How about `{{# foo === true }}` instead?

evs-chris:
Good to go.

test/__tests/components.js
Rich-Harris: yeah that makes sense - I updated it

Issue #1737, opened 2015-02-19
keypaths breaking on unshift in edge


aphitiel:
See http://jsfiddle.net/4ukh3aqu/

After clicking `add1` the results of `{{@keypath}}` break for items existing before the unshift. Clicking `log` shows that event.keypath is still ok, except for that "special" case.

Push works as expected, as does latest.

aphitiel:
Issue starts with commit 7bf0d176394bf8a1aa1aeab35177ada4f03bad39 on branch edge, which seems to be based on 5ca8675c.

Sadly, I can't get that to build: `gobble: 02-esperanto transformation failed`
`EEXIST, file already exists '/home/aphitiel/git/ractive/.gobble/02-esperanto/.cache/Ractive.js.map'`
Any quick ideas?

Madgvox:
This looks like it might be the same issue as https://github.com/ractivejs/ractive/issues/1729.

evs-chris:
It looks like this is partially caused by `@keypath` finding an expression context from the `#if` just above it. Should https://github.com/ractivejs/ractive/blob/dev/src/virtualdom/items/shared/Resolvers/SpecialResolver.js#L64 also check that the context doesn't start with `'${'`?

martypdx:
> finding an expression context from the #if just above it.

I didn't think that `#if` blocks created context. Do they?

> check that the context doesn't start with '${'

But wouldn't it be valid to do: `{{#each getArray(foo)}}{{@keypath}}{{/each}}`?

Also, for reference, we have similar logic in [`Binding`](https://github.com/ractivejs/ractive/blob/dev/src/virtualdom/items/Element/Binding/Binding.js#L17-L20) that checks the end of the keypath for `}` because it's perfectly valid to bind to a property of the result of an expression. So, you could have `{{#with condition ? obj1 : obj2 }}{{#each .property}}{{@keypath}}{{/each}}`

evs-chris:
I didn't think they did, but stepping through, it looks like that's the issue.

I don't think you can really do anything with an expression context, like `set` with it, so is it useful as a ref beyond just displaying it?

martypdx:
In my mind, it's not whether or not they're expressions, it's that conditional blocks don't change or create context they just control inclusion/exclusion of template content. Take it out and see what breaks :)

aphitiel:
Some observations:
* There might be 2 different problems in http://jsfiddle.net/4ukh3aqu/:
[...]

Issue #1106, opened 2014-08-12
Reference expression sections break if their keypath becomes invalid


Rich-Harris:
Found while looking into #1102 - http://jsfiddle.net/rich_harris/p4r1sgo0/

PR #1742, opened 2015-02-20
Mark the entire array on unshift

### LINKED ISSUE #1729: ractive.unshift does not propagate changes
I stumbled over this while looking at an alternative implementation for `link`/`unlink`. It looks like `smartUpdate` is skipping existing indices, which, I believe, is a no-go for `unshift`, as they should all change.

http://jsfiddle.net/8u6sgmL7/

evs-chris:
This fixes #1729, but I'm not sure about #1737. Does this look like a reasonable fix?


Rich-Harris:
I think we might have to go deeper... all array mutations except `sort` and `reverse` are treated as their `splice` equivalents internally, so having a special case for `unshift` doesn't quite solve the problem as this fork of your #1729 fiddle shows: http://jsfiddle.net/rich_harris/LjLrkej2/

evs-chris:
From a slightly different perspective, it looks like the whole array keypath should be marked if it has non-shuffle-able deps. I just pushed a commit that covers simple cases that don't have shuffle-able deps and non-shuffle-able child deps. I'll push a failing test for that, but the best way to handle it is eluding me at the moment. If anyone has a nice clean was to handle that, please feel free to point me in the right direction or push a fix.

martypdx:
That's an ugly bit of problem. You'd essentially need to take all the entries in `.depsMap['default'][ keypath.str ]` and then look through dependents for each one in deps['default'][ str ], __then__ for each dependent find the inner context and see if the `context.str === keypath.str`.

Is that even a savings anymore over `this.mark( keypath )`?

evs-chris:
This last change marks everything beyond the splice index if the size of the array changes after the shuffling takes place. I'm not 100% sure that there's no redundant `mark`ing going on, but there is at least enough.

evs-chris:
I think I found the issue with #1737. When the fragment gets rebound during shuffling, the nested conditional, which should have no context, gets the context of the shuffle added, even though it's probably not right. I added a check for sections that shouldn't contribute to the context chain in the fragment rebind in the last commit. I'll add a test in a bit.

evs-chris:
I probably should have split this up, but I didn't realize how unrelated the unshift issues were.

Anyways, should any fragments other than sections ever get a context? It looks like finding context always walks up the tree until if finds a node with a context and non-if/unless sections are the only ones that ever introduce context. That last commit keeps any fragments that aren't sections from growing a context.

src/virtualdom/Fragment/prototype/rebind.js
```diff
@@ -3,7 +3,9 @@ import { assignNewKeypath } from 'shared/keypaths';
 export default function Fragment$rebind ( oldKeypath, newKeypath ) {
 
 	// assign new context keypath if needed
-	assignNewKeypath( this, 'context', oldKeypath, newKeypath );
+	if ( !this.owner || this.owner.hasContext ) {
+		assignNewKeypath( this, 'context', oldKeypath, newKeypath );
+	}
```
Rich-Harris: Could we just do this?

```js
if ( this.context ) {
  assignNewKeypath(...);
}
```

Or would that cause other problems? That way we wouldn't need `hasContext`.

Or maybe [this line](https://github.com/ractivejs/ractive/blob/dev/src/shared/keypaths.js#L71) should actually be

```js
if ( !existingKeypath || ( existingKeypath.equalsOrStartsWith( newKeypath ) || !existingKeypath.equalsOrStartsWith( oldKeypath ) ) ) {
  return;
}
```

Rich-Harris:
> Anyways, should any fragments other than sections ever get a context?

You're right, they shouldn't

src/virtualdom/Fragment/prototype/rebind.js
evs-chris: I tried that originally, but it failed a test that expected the fragment to grow a context in particular circumstances.

martypdx:
:+1:

src/virtualdom/Fragment/prototype/rebind.js
evs-chris: Specifically, it fails the test for #1106 in the misc tests.

aphitiel:
\o/ This looks great! One small nit though, if I may: I get 'Attempted to unrender a fragment that was not rendered' when unshifting to an array of objects containing arrays of varying lengths. Here is the smallest sample I could come up with:

```
{{#each items}}
  {{.title}}
  {{#each .tags}}{{.}}{{/each}}
{{/each}}
```

```
var ractive = new Ractive({
    el: 'main',
    template: '#template',
    data: {
      items: [
        {title: 'oneTag',  tags: ['A']},
        {title: 'TwoTags', tags: ['B', 'C']} ]}
});
ractive.unshift('items', {title: 'three'});
```
This works if data.items[0].tags is initialized with length 2.

evs-chris:
The error unrendering a fragment that isn't rendered turns out to be a little nastier than I thought it would. It seems that even though fragments get shuffled in list sections, they also ended up getting torn down and re-rendered on rebind.

Since, on shuffle, the only thing that really changes is the keypath, it seems the best way to handle it is to let the rebind/resolve machinery know that we only care about the keypath and not the value. I suspect this is how it was meant to work to begin with (?), but it got eaten in a refactor somewhere (probably my fault, as I vaguely remember dropping values from rebind at some point when fixing an index propagation bug). The upshot is that this should make iterative section updates a fair amount faster, especially for larger iterators.

So before I go much further, am I on the right path here, or is there a better way to handle this?

martypdx:
@evs-chris I think you've hit the nail on the head. Shuffle should reorder the fragments, but avoid mucking about within the fragments.

evs-chris:
I think that last commit has everything covered, but I would greatly appreciate a review to guard against anything particularly boneheaded on my part. There's a lot of pushing the `newValue` bool around to various bits and bobs in the fragment tree. I think everything is notified correctly now, but I got a little lost in the bubbling forest of resolvers for a bit.

It looks like this PR is a little bit of a performance drag for array methods. It got a little speed boost by shuffling instead of tearing down and re-rendering stuff, but it also slows down because it notifies non-shufflable deps about smartUpdates as well. There may be a way to address the slowdown, but everything I came up with was just as bad, if not worse, than what's here.

evs-chris:
It looks like this takes care of #1751 as well.

src/virtualdom/items/Section/prototype/setValue.js
```diff
@@ -88,15 +88,19 @@ function reevaluateSection ( section, value ) {
...
 
+	section.hasContext = true;
+
 	// If we already know the section type, great
 	// TODO can this be optimised? i.e. pick an reevaluateSection function during init
 	// and avoid doing this each time?
 	if ( section.subtype ) {
 		switch ( section.subtype ) {
 			case SECTION_IF:
+			section.hasContext = false;
```
martypdx: Doesn't line 145 below also need `section.hasContext = false;`?

evs-chris: I don't think so, since `{{#foo}}` sections introduce `foo` as a context, don't they? I'm a little hazy on that, but it follows that `{{#foo}}{{bar}}{{/}}` would render `inner` with `{ foo: { bar: 'inner' }, bar: 'outer' }`. That would require the context, wouldn't it?

martypdx: Github won't let me add a line comment that's not a changed line, so [here's](https://github.com/ractivejs/ractive/blob/dev/src/virtualdom/items/Section/prototype/setValue.js#L128-L141) the block I'm referring to. It only gets to the reevaAsConditional part _if_ the value is not an object.

evs-chris: Ah! I see it now. Fixed.

Rich-Harris:
> It looks like this PR is a little bit of a performance drag for array methods.

How much of a drag are we talking? I've poked around a bit to see if there's an alternative approach, but so far I've come up empty too

Rich-Harris:
> There's a lot of pushing the `newValue` bool around to various bits and bobs in the fragment tree

Have to confess I'm not quite sure what `newValue` is for - removing it doesn't seem to break any tests?

evs-chris:
> How much of a drag are we talking?

Looks like about 50% on my machine on the 'splice large array' perf tests in Chrome. control (dev when I originally branched) and 0.6.1 have an average duration of 17 (ms?) and this does 29. Oddly, this apparently speeds up 'box' by a bit. Time variances are a little wild though.

> removing it doesn't seem to break any tests?

`newValue === false` is a signal for rebind in sections not to tear down the fragments that are rebound, since they are just shifting their keypath instead of actually changing. Removing it probably would't fail any tests that don't specifically look for fragments not being torn down in a section downstream of the rebind/resolver from which you remove it.

There may be a better way to do that, but it looked like `rebind` is used in thing other than `smartUpdate`s, so wholesale assumption that only the keypath was shifting didn't seem like an option.

Rich-Harris:
I'm seeing similar results performance-wise. Eesh, that is a bit of a slowdown. Though I suppose if we were 'cheating' before (by not notifying all changed keypaths) then perhaps it's unavoidable

evs-chris:
@Rich-Harris I think I may be misunderstanding the purpose of the silent `set` in `smartUpdate`. I thought it was originally there to push a new array value in the case of a merge (or non-splice modification), so I followed the TODO comment and moved the updates to the upstream methods that didn't already update the array. Was I far off?

As far as performance, it looks like just `mark`ing the keypath if the array length changes is a little bit (~4ms) faster than marking `length` and changed indices individually for the splice test.

Rich-Harris:
@evs-chris not sure to be honest! I didn't actually mean to push that commit, my fingers must have been acting on muscle memory. Was just going through this PR to understand it - I didn't initially understand that specific change because I thought (mistakenly) that `viewmodel.smartUpdate()` was only called by `merge.js` and `processWrapper.js` (incidentally the latter case seems wrong, just like the magic mode adaptor interacting directly with the viewmodel is wrong).

Notwithstanding the long-standing mistake in `processWrapper.js`, I reckon that either `processWrapper.js` and `makeArrayMethod.js` should *both* call `.set()`, or *neither* should. Just not sure which yet!

> As far as performance, it looks like just marking the keypath if the array length changes is a little bit (~4ms) faster

Huh...

martypdx:
Just chiming in with validation of having gone down many paths mentioned above while looking at this. 

I did just do some basic DOM perf simulating adding one to the top while removing one from the bottom - comparing insertBefore/remove with changing the textContent of a single single text node. And yes, it is faster to shuffle the nodes than to reassign the values within the node, even at smaller numbers like 10 nodes.

So it does make sense to try and move the dom nodes on "shuffle" type operations, but reuse them on an array value set.

Rich-Harris:
It also means you can do stuff like [this](http://jsfiddle.net/rich_harris/bfdj68qg/), which is *really* hard if you're updating the text content of nodes to achieve the same end state

martypdx:
> It also means you can do stuff like this 

Put that in your dbmonster pipe and smoke it :sunglasses: 

So, as I understand it, the crux of the issue is for dependents that register directly against a static index value like `{{foo.0}}`. Is there an approach that recognizes those numeric ending keypaths - seperate than the registrations of the sections, which we could possibly detect by modifying sections use of Mustache and join calls somehow, - that we could notify outside of the shuffle operation?

Rich-Harris:
I think I might be onto something... leaving aside the issue about sections erroneously gaining a context during rebinding, the other three new failing tests can be fixed by applying @evs-chris' trick for correctly marking dirty keypaths (i.e. all of them in the case of `unshift`).

That caused another test to fail ('Array shuffling only adjusts context and doesn't tear stuff down to rebuild it'), but it turns out that's because of an entirely separate bug in `ReferenceExpressionResolver.js` - on rebind, its *members* are rebound but its *base* isn't. Fixing that fixes the newly failing test, and we don't need to pass any additional information around.

There's still a performance hit because we're checking all those extra keypaths, but it's a significantly smaller hit than before.

Still working on it, and hoping that I'm not being prematurely optimistic - will probably put this lot in a separate PR otherwise the history will be all over the place. Should have something to share soon

evs-chris:
> separate PR otherwise the history will be all over the place

It'd probably be safest to close this and re-PR on each of the 3 or 4 separate concerns that managed to wriggle in here from the various `unshift` issues.

Rich-Harris:
I've created two PRs, #1794 and #1795. Might not be the most logical division in the world but all the tests from this PR are in, and I've confirmed that they all pass when both of those branches are combined.

Sanity checks very much appreciated... 

I wish `git commit` let you distinguish between 'commit author' and 'person who did all the actual work'... sorry @evs-chris!

evs-chris:
I'm not entirely sure I'm sure I'm qualified for sanity checks, but those look good to me. So I'll close this. :+1:

Issue #1716, opened 2015-02-15
Investigate AppVeyor failures


Rich-Harris:
Have hooked Ractive up to AppVeyor (a bit like Travis for Windows) so that situations like #1706 don't arise again. For some reason, it's not working (even though it builds on Windows locally, at least for me in a VM), due to [this line](https://ci.appveyor.com/project/Rich-Harris/ractive/build/1.0.1/job/wpy4s3u2g6e2n6ni) (175):

![screen shot 2015-02-15 at 4 40 49 pm](https://cloud.githubusercontent.com/assets/1162160/6205122/75130ffa-b531-11e4-82ec-e80837d724b2.png)

Not sure why that's causing problems.

Rich-Harris:
Any Windows folk have any insight into this? I've tried being more explicit about which test file to run, but Mocha is still giving the same error. It works fine for me locally...

https://ci.appveyor.com/project/Rich-Harris/ractive/build/1.0.51/job/i3vmyvjw5rn9jbej (line 173)

codler:
What have you installed to run Bourne shell on windows? Do Appveyor have shell?

codler:
@Rich-Harris  Could it be that you forgot to run `build.sh` script before running `test.sh` script?

codler:
Im looking on [appveyor.yml](https://github.com/ractivejs/ractive/blob/dev/appveyor.yml) and the log and see it will only run `test.sh`. I compare it with [.travis.yml](https://github.com/ractivejs/ractive/blob/dev/.travis.yml) and see it will run `build.sh`. Looking closer inside `build.sh` I see `$MOD/gobble build tmp` runs before `test.sh`. I believe `$MOD/gobble build tmp` is missing in appveyor.yml

Rich-Harris:
@codler oops, good spot! The test files don't exist until `npm run build` happens. I have updated the appveyor.yml file and have the appropriate headwear prepared for when it successfully builds. Thanks :)

![duncecap](https://cloud.githubusercontent.com/assets/1162160/6308002/86309de4-b90e-11e4-93fe-5c7a965026fc.jpg)

evs-chris:
AppVeyor seems to be pretty happy these days...

PR #1754, opened 2015-02-27
check for Promise on window

bbudd:
We can’t just check for `typeof Promise === “function”` for reliance on native promises, because `Promise` is the declared name of the entire Class and therefore will be considered `undefined`. If we check against `self.Promise`, we’ll know that teh scope that we’re checking in is at least a worker or window scope, and we'll be able to utilize those native Promises.

FYI, this means that native promises were never really being used previously, and were always using to the Ractive version.

To test, put a breakpoint inside the ractive cdn latest at line 350, and check `typeof Promise` in the console there.

The other option for this is to change the name of the class from `Promise` to something else.

### DIFF
--- a/src/utils/Promise.js
+++ b/src/utils/Promise.js
@@ -3,9 +3,9 @@ var _Promise,
 	FULFILLED = {},
 	REJECTED = {};
 
-if ( typeof Promise === 'function' ) {
+if ( typeof self !== 'undefined' && typeof self.Promise === 'function' ) {
 	// use native Promise
-	_Promise = Promise;
+	_Promise = self.Promise;
 } else {
 	_Promise = function ( callback ) {
 		var fulfilledHandlers = [],
--- a/src/utils/Promise.js
+++ b/src/utils/Promise.js
@@ -1,3 +1,4 @@
+/*global self */
 var _Promise,
 	PENDING = {},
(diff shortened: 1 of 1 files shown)


bbudd:
I note the appveyor fail, but following the link doesn't give me a ton of insight as to _why_ it failed.

Rich-Harris:
@bbudd Thanks very much for the PR, though I'm fairly sure this is fixed in edge already? There was a bug introduced by the build process, whereby the internal `_Promise` name would be replaced with `Promise` (because of the module's filename), thereby shadowing the native `Promise`, but that's since been fixed. You can confirm this with `Ractive.Promise === window.Promise`: http://jsfiddle.net/5k0ha7t4/

You can safely ignore the AppVeyor thing - for some reason the build process isn't working on Windows (it works locally for me in a VM, but  ¯\\_ (ツ) _/¯ Windows). There's an open issue for it - #1716

bbudd:
Great, thanks @Rich-Harris  - any idea when it will be up in master?

Rich-Harris:
The million dollar question... I'm on record as saying 'fairly soon', which hopefully means next couple of weeks!

MartinKolarik:
Closing as this is fixed in `v0.7`

PR #1756, opened 2015-02-27
Support escaping mustaches in templates

### LINKED ISSUE #276: Escape brackets
Using [formatter.js](http://firstopinion.github.io/formatter.js/demos.html) you need to specify the validation pattern using `<input data-pattern='{{9999}}-{{9999}}-{{9999}}-{{9999}}' />` how can i escape the brackets, so that ractive does now swallow them?

`$('input').data('pattern')` -> `9999-9999-9999-9999` but should be `{{9999}}-{{9999}}-{{9999}}-{{9999}}`

/cc @Rich-Harris

evs-chris:
This fixes #276, which was closed as wontfix. Our docs currently use handlebars, which supports escaping mustaches with '\', so this does the same for Ractive.

Feel free to close this one wontfix too. I just needed a brief distraction to clear my head, and none of the current open issues aligned with my mood correctly.

### DIFF
--- a/src/parse/converters/readMustache.js
+++ b/src/parse/converters/readMustache.js
@@ -25,7 +25,11 @@ function readMustacheOfType ( parser, tag ) {
 
 	start = parser.pos;
 
-	if ( !parser.matchString( tag.open ) ) {
+	if ( parser.matchString( '\\' + tag.open ) ) {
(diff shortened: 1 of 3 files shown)


Rich-Harris:
I ran into this exact problem putting a presentation together the other day, I think this makes sense

PR #1757, opened 2015-02-28
Add error for extra chars after method call event handler

### LINKED ISSUE #1288: Parser fails badly if method invocation isn't correctly formed
http://jsfiddle.net/rich_harris/8sttznxp/

evs-chris:
This throws a parse error if there is extra, non-space stuff beyond the last `)` on a method call event handler with the offending chars included in the message. The pos is slightly off because it's already moved to the end of the attr, but it's pretty darn close. I added a parser param for `processDirective` so that the appropriate `ParseError` can be generated.

fixes #1288


Rich-Harris:
:+1:

Issue #1761, opened 2015-03-01
Data functions do not correctly roll-up in edge.


martypdx:
Discovered with #1760 (unless I've misunderstood the changes in edge).

This code:

```js
var View = Ractive.extend({
    data: function () {
        return {
            foo: 'foo',
            bar: 'bar'
        }
    }
})
var r = new View({
    data: {
        foo: 'qux',
        other: 'other'
    }
})
```

Should end up with quasi-data object:

```js
{
  foo: 'qux',
  bar: 'bar',
  other: 'other'
}
```

martypdx:
Fixed via #1762

PR #1762, opened 2015-03-02
Fix inherited data functions #1761

martypdx:
Cleaned up data option inheritance. Data is rolled-up, but no more `super` because we're not passing the data object through. So functions are run for their result and "inheritance" runs as expected from parent up to child. 

Only oddity is that I put in a preference to use objects returned from functions over object literals. This makes sense when you look at the two similar use cases that both, to me, suggest function preference is the way to go:

```js

// FIRST CASE
var View = Ractive.extend({
    data: function(){
        return new Model();
    }
});

var ractive = new View({
    data: {
        foo: 'bar'
    }
});

assert(ractive.get() instanceof Model);
asset(ractive.get('foo'), 'bar');

// SECOND (INVERSE) CASE
Ractive.prototype.data = {
    foo: 'bar'
}

var ractive = new Ractive({
    data: function(){
        return new Model();
    }
});

assert(ractive.get() instanceof Model);
asset(ractive.get('foo'), 'bar');
```

But admittedly a decision largely in response to existing tests.

Oh, and a render perf tests seems to have found it's way into this commit...


martypdx:
Though in the case where people use a function to return an instance of default data, it might be weird:

```js
var data = { ... some POJO ... };

Ractive.prototype.data = function(){
    return {
        foo: 'bar',
        bizz: 'buzz'
    }
}

var ractive = new Ractive({
    data: data
});

assert(ractive.get() === data ); // false!
```

martypdx:
Okay refined the logic to favor function returns _if_ they return non-POJO object.

Issue #1745, opened 2015-02-20
Component reset() does not reset arrays


sixfeetover:
If you initialize a ractive component with array data, `reset()` will not change array values back to their initial state. In other words:

```javascript
var C = Ractive.extend({
  data: {ary: []}
});
c = new C();
c.push('ary', 5)
c.reset()
c.get('ary') // =>  [5]
```

See the console output from [this fiddle](http://jsfiddle.net/jfraser/jL3o3nn5/) for a slightly more detailed example. Note that other non-array properties DO reset as expected.

I noticed this when attempting to override an instance of a Ractive component in my app, where the new instance inherited the array data value from the previous instance. That behavior is also represented in the fiddle. I then tried to use `reset` as a band-aid, but that didn't work either. For now, I have solved it by setting my data in `oninit` rather than as a property of the component.

I am unclear if this is a bug, the expected behavior for arrays, or simply not the correct usage of `reset`.

martypdx:
I was in the wrapper code yesterday, and I remember a test failed if the wrapper value was not recycled after teardown! I'll need to go back and have a look...

Rich-Harris:
I was going to say that a good solution to this problem is to use a function for the component's `data` property, so that it gets created anew for each instance. There really isn't a good way to reset non-primitive values (i.e. arrays and objects) back to some 'initial' value if you've modified the array itself - one of the reasons [immutable data](https://github.com/facebook/immutable-js) is so appealing.

But it turns out that there's a bug somewhere - I [forked the fiddle](http://jsfiddle.net/rich_harris/d0zscp0r/), replacing the data object with a data function, and it's throwing an error. Marking this accordingly...

Rich-Harris:
I've been looking at the [data configuration code](https://github.com/ractivejs/ractive/blob/dev/src/Ractive/config/custom/data.js) for a while now and I'm ashamed to say I'm struggling to understand it (and hence fix it). It seems we do different things depending on various factors - whether we're extending from a class that has default data, whether that default data is an object or a function, whether it's a POJO, and so on. 

I suppose this is our own fault for nobly trying to make everything Just Work, and I'm wondering how we might bring it under control. The alarming part is I wouldn't be able to explain how data gets inherited other than in non-exotic cases, which suggests either a) it's too complex or b) I'm too simple. Will have a bit of a poke and see what falls out.

martypdx:
[...]

PR #1767, opened 2015-03-03
gh-1745

Rich-Harris:
not ready for merge, just tracking progress on #1745

### DIFF
--- a/src/Ractive/config/custom/data.js
+++ b/src/Ractive/config/custom/data.js
@@ -1,4 +1,4 @@
-import { fatal, warn } from 'utils/log';
+import { fatal, warn, warnOnce } from 'utils/log';
 
 function validate ( data ) {
 	// Warn if userOptions.data is a non-POJO
@@ -60,11 +60,27 @@ function combine ( parentValue, childValue ) {
 	}
 
 	return function () {
-		let child = childIsFn ? childValue.call( this ) : childValue;
-		let parent = parentIsFn ? parentValue.call( this ) : parentValue;
+		let child = childIsFn ? callDataFunction( childValue, this ) : childValue;
+		let parent = parentIsFn ? callDataFunction( parentValue, this ) : parentValue;
 
 		return fromProperties( child, parent );
 	};
+}
+
+function callDataFunction ( fn, context ) {
+	let data = fn.call( context );
+
+	if ( !data ) return;
+
+	if ( typeof data !== 'object' ) {
+		fatal( 'Data function must return an object' );
+	}
+
+	if ( data.constructor !== Object ) {
+		warnOnce( 'Data function returned something other than a plain JavaScript object. This might work, but is strongly discouraged' );
+	}
+
+	return data;
 }
 
 function fromProperties ( primary, secondary ) {
--- a/src/Ractive/config/custom/data.js
+++ b/src/Ractive/config/custom/data.js
@@ -18,11 +18,8 @@ var dataConfigurator = {
 	reset: function ( ractive ) {
 		var result = this.init( ractive.constructor, ractive, ractive.viewmodel );
 
-		if ( result ) {
-			ractive.viewmodel.data = result;
-			ractive.viewmodel.clearCache( '' );
-			return true;
-		}
+		ractive.viewmodel.reset( result );
+		return true;
 	}
 };
 
--- a/src/Ractive/config/custom/data.js
+++ b/src/Ractive/config/custom/data.js
@@ -2,11 +2,11 @@ var dataConfigurator = {
 	name: 'data',
 
 	extend: ( Parent, proto, options ) => {
-		proto.data = combine( Parent, proto, options );
+		proto.data = combine( proto.data, options.data );
 	},
 
 	init: ( Parent, ractive, options ) => {
-		var result = combine( Parent, ractive, options );
+		var result = combine( Parent.prototype.data, options.data );
 
 		if ( typeof result === 'function' ) {
 			result = result.call( ractive );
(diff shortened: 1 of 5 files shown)


martypdx:
The addedKeys code handles this:

```js
var OurRactive = require('our-default-ractive'); // uses a data function
OurRactive.defaults.data.foo = 'bar';
var ractive = new OurRactive({});

ractive.foo === 'bar';
```

Individually, functions and adding props to `default.data` _are_ used in the wild. Does anyone run into this conflict using them together?  I dunno...

src/Ractive/config/custom/data.js
```diff
@@ -18,82 +18,60 @@ var dataConfigurator = {
...
-		parent = parentIsFn ? parent.call(this) : parent;
+	return function () {
+		let child = childIsFn ? childValue.call( this ) : childValue;
+		let parent = parentIsFn ? parentValue.call( this ) : parentValue;
 
 		// allow parent return value to take precedence if
 		// it is a function that returns non-POJO Model
 		// and child is either not a function or does not return non-POJO
 		if ( ( parentIsFn && parent.constructor !== Object ) && ( !childIsFn || child.constructor === Object ) ) {
-			return fromProperties( parent, child, true );
```
martypdx: If you take the force out, then the child keys won't override the model keys:

```js

function Model(){
    this.foo = 'qux';
}
var View = Ractive.extend({
    data: function(){
        return new Model();
    }
});

var ractive = new View({
    data: {
        foo: 'bar'
    }
});

// will only be 'bar' if child keys are forced onto parent
// because the retained object has been switched to the Model
ractive.get('foo') === 'bar'
```

Rich-Harris: Isn't it reckless to be overriding the keys of a `Model` anyway, given that it'll probably be storing its data in an `attributes` hash or similar?

No tests failed when I removed it, so I figured I'd take it out and see if you noticed :stuck_out_tongue_winking_eye:

Rich-Harris: Removing that whole clause and just having `fromProperties( child, parent )` breaks two tests - 'Data Initialisation: Data functions are inherited and Models are used as data object' and 'Miscellaneous: Instances of subclasses with non-POJO default models have the correct prototype' (the one you mentioned on the #1745 thread). If it's the case that we *don't* want to muck around with non-POJOs (for the reason I gave above) then we'd presumably want rid of both of those tests anyway

Rich-Harris:
> The addedKeys code handles this

Ah, I understand now. I thought it might just be dead code, since no tests depended on it. To my mind that's going above and beyond the call of duty, I honestly don't think it's something we need to support.

The more I think about it, the more I think the following model makes sense: if `ractive` is an instance of `B` which extends `A` which extends `Ractive`, then `ractive.viewmodel.data` is either a) the `data` option provided at initialisation, or b) a fresh `{}`. That object should have properties from `B`, `A` and `Ractive` copied, where later properties override earlier ones, and where 'properties from x' means either `x.prototype.data` or (if a function) `x.prototype.data.call(ractive)`.

That's pretty close to what we've got now, except it's totally predictable what overrides what.

Non-POJOs as root objects still throw a spanner in the works, but that's basically inevitable - I think we should generally discourage it.

A couple of other thoughts:

* It's probably not a bad idea to warn the user if they instantiate a component that has non-primitive default values, since the instance could potentially mutate them, causing havoc. The warning would presumably say 'use a data function - here's how'
* Performance-wise, it would probably be much better to use prototypal inheritance than property copying. The only real reason *not* to do that is that `ractive.viewmodel.data === data` is no longer true (meh), and for the same reason non-POJOs as root objects would break. But given that that seems like an accident-prone thing to do anyway, maybe we should just optimise for the common case.

Rich-Harris:
The last two commits change the behaviour as per the previous comment - child data always takes priority. I think this is less confusing. I've deleted the two relevant tests, so

:loudspeaker: **breaking change klaxon** :loudspeaker: 

though I can't really imagine this causing problems in the wild. Ractive will also now warn you if your root `data` object is a non-POJO - in the vast majority of cases where there isn't all sorts of inheritance going on, that will continue to work, but as I said above I think it's appropriate to discourage it.

@martypdx (and everyone else) what do you reckon?

> Performance-wise, it would probably be much better to use prototypal inheritance than property copying

I had a quick go at this, and it broke a few tests so I figured it's probably a separate piece of work we can look at if profiling reveals the current approach to be a bottleneck.

martypdx:
> Ractive will also now warn you if your root data object is a non-POJO

:+1: Exactly, the configuration should treat data as a registry. Any type of object can be passed as a key, or the root can be set explicitly with a model:

```js
// nothing wrong with this
data {
    model: new Model();
}
```

```js
// or this
onconfig: function() {
    this.set('', new Model( this.get('name'), this.get('age') ) );
}
```

The validation warning should probably happen on function return, not just options value.

> I had a quick go at this, and it broke a few tests so I figured it's probably a separate piece of work we can look at if profiling reveals the current approach to be a bottleneck.

Probably the tests for checking that, essentially `ractive.data === data` which doesn't need to be the case

martypdx:
> Performance-wise, it would probably be much better to use prototypal inheritance than property copying

Actually, for data I don't think that would be the case. It is most likely that a) there is no parent data to inherit or b) the amount of keys in parent data is far less than in the supplied data object.

Rich-Harris:
> The validation warning should probably happen on function return, not just options value.

Agree, I was being lazy. Fixed!

You might be right about the prototypal inheritance thing. We'd definitely need some benchmarks if we were going to look into it.

Will go ahead and merge this

PR #1775, opened 2015-03-06
Discussion: Pass thru keepExistingWrapper from adaptor reset?

martypdx:
I was trying to correct what I thought was an oversight, that adaptors are always torn down because the result of `reset` is never passed through to the `mark` method so on `clearCache` `teardown` always gets call.

This PR fixes that, but breaks magic adaptor because it [cheats and goes through the back door](https://github.com/ractivejs/ractive/blob/dev/src/Ractive/static/adaptors/magic.js#L87). And teardown can also selectively control whether it actually goes away by returning a non-false value. 

It seems `adaptor.teardown` has become the new `adaptor.reset` :)

I think the adaptor teardown should happen with the viewmodel teardown and be reserved solely for that purpose of clearing up used resources. And reset needs to serve the purpose of handling re-"`set`" of the keypath on which the wrapper is attached.

But given my level of confusion, I wanted to bring it up before trying to fix it up.

### DIFF
--- a/src/viewmodel/prototype/set.js
+++ b/src/viewmodel/prototype/set.js
@@ -45,7 +45,7 @@ export default function Viewmodel$set ( keypath, value, options = {} ) {
 	}
 
 	if ( !options.silent ) {
-		this.mark( keypath );
+		this.mark( keypath, { keepExistingWrapper: keepExistingWrapper } );
 	} else {
 		// We're setting a parent of the original target keypath (i.e.
 		// creating a fresh branch) - we need to clear the cache, but


evs-chris:
My level of confusion probably exceeds yours by a fair amount (so take whatever I say with a block of salt), but based on the backbone adapter, it looks to me like you're right. `adaptor.teardown` should be reserved for when a wrap-able value is entirely removed or the viewmodel is torn down. I don't see any reason that adaptors should be able to veto their teardown.

Further, is magic mode fully tested? Because commenting out<sup>[1](#footnote-1)</sup> the `mark` and `addRactive` lines you pointed to doesn't seem to have any negative effects in the tests or in a sandbox example. It looks like the `teardown` veto in magic mode happens because it originally got triggered from that `mark`. Can `reset` even get called from outside of the normal runloop, necessitating that `mark`?

<sup>1.</sup> <a name="footnote-1"></a> I'm using a branch with gh-1745 and gh-1729 merged in, so that may be affecting my results vs. edge.

martypdx:
> Because commenting out1 the mark and addRactive lines you pointed to doesn't seem to have any negative effects in the tests or in a sandbox example

Did you try combining that with the change in this commit? I tried that too, and there were a handful of specific fails, but curious if other changes "fixed" it.

evs-chris:
Adding in this change causes test failures on the same tests as Travis. It looks like an issue with the magic array wrapper. On `reset`, it doesn't update its own value, so the old value is passed through to everything beyond. Adding a `this.value = value` to [the magic array wrapper `reset`](https://github.com/ractivejs/ractive/blob/dev/src/Ractive/static/adaptors/magicArray.js#L34) turns everything green again, though it may need to be a little more complicated than that.

Rich-Harris:
> My level of confusion probably exceeds yours by a fair amount

Don't want to get all competitive but I reckon I could give you both a run for your money, despite it being my code. (Maybe 'despite' is the wrong word...)

If I wasn't already nodding in agreement with the proposals here, 'I don't see any reason that adaptors should be able to veto their teardown' would have sealed the deal. Are we talking about a breaking change for adaptor plugin authors? If so, I have an idea for how we might handle deprecation of old adaptors (it involves a more 'detectable' API change).

> This PR fixes that, but breaks magic adaptor because it cheats and goes through the back door

Facepalm. That's pretty bad... we'll need to fix that anyway if we're to split array/magic adaptors out as plugins (assuming that's possible, given that sometimes you need them in combination).

> Further, is magic mode fully tested?

Another good reason to split it out, I reckon

martypdx:
> Don't want to get all competitive but I reckon I could give you both a run for your money, despite it being my code. (Maybe 'despite' is the wrong word...)

@Rich-Harris Cool beans. Sounds like were on the same page.

> Are we talking about a breaking change for adaptor plugin authors? If so, I have an idea for how we might handle deprecation of old adaptors (it involves a more 'detectable' API change).

Hadn't gotten that far, mostly trying to shore up what we have. The current adaptor pattern _is_ hard to grok for newcomers.

martypdx:
Also @Rich-Harris checkout https://github.com/ractivejs/ractive-adaptors-ampersand/pull/2 sometime if you haven't seen it yet.

evs-chris:
Gonna punt this to #2512

Issue #1581, opened 2014-12-16
Mismatched template version! ...?


dagnelies:
Hi,

The full error is:
```
Mismatched template version! Please ensure you are using the latest version of Ractive.js in your build process as well as in your app
```
I encountered this error but I don't know the cause of it ...might someone shed some light on it? ...both with latest and edge.

TehShrike:
aaaugh I spent an hour trying to figure out where a second version of Ractive could be coming from

PR #1778, opened 2015-03-07
Fixing rabbit-trail-inducing error message

TehShrike:
Added a new error message when non-string templates are passed in without a parsed-Ractive-template version.

Fixes ractivejs/ractive#1581


martypdx:
:+1: Thx!

PR #1779, opened 2015-03-07
Change main from ractive.js to build/ractive.js

TehShrike:
I'm not sure this is an issue or not.

After I built the project locally, when I tried to use Ractive via npm link, I found that I wasn't able to require() the module because there was no ractive.js file in the root package directory.

Are there other steps that happen pre-publish?  I couldn't find another script that looked like it would move build/ractive.js to the root directory.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "description": "Next-generation DOM manipulation",
   "version": "0.7.0-edge",
   "homepage": "http://ractivejs.org",
-  "main": "ractive.js",
+  "main": "build/ractive.js",
   "keywords": [
     "template",
     "templating",


Rich-Harris:
Thanks, I think this is correct. The confusion comes from the fact that releases are tagged against the [build branch](https://github.com/ractivejs/ractive/tree/build), so that when you `npm install ractive` you only get the built files rather than having to download all the extra gubbins we need in development like PhantomJS. In that branch, `ractive.js` *is* in the root folder.

Happily, it has its own `package.json`, so changing this will fix `npm link` without affecting `npm install`.

PR #1781, opened 2015-03-09
Draggable is not a boolean attribute

### LINKED ISSUE #1780: bug: draggable='false' becomes true
Hi,

Here is the fiddle:
http://jsfiddle.net/2wgtqxqk/1/

When using `draggable='false'` in a template, it is not properly handled and the resulting DOM node becomes `draggable='true'`. This is illustrated by the fiddle and can be seen in the DOM inspector.

Suprisingly, using a variable instead of "false" works.

evs-chris:
According to the w3c [spec](http://www.w3.org/TR/2010/WD-html5-20101019/dnd.html#the-draggable-attribute), draggable is an enumerated attribute that can be 'true', 'false', or '' (empty/auto, which behaves differently depending on its element). Any objections?

Sorry, there's a little extra noise in the test file, as I swiffered while I was there.

Fixes #1780


evs-chris:
It seems a node module update somewhere is killing the build...

Rich-Harris:
> It seems a node module update somewhere is killing the build..

Seems gobble may be at fault. Downgrading to 0.7.3 fixes the build. Looking into it...

> draggable is an enumerated attribute

Gah. Technically that means it probably *shouldn't* be treated as a boolean, and should have some special case logic applied like `contenteditable` (another enumerated attribute), though I can't think of any situation where treating it as a boolean would cause problems, so :+1:

Rich-Harris:
Fixed the regression that gobble 0.7.5 introduced. Thanks for catching it

evs-chris:
Thanks for fixing it!

Issue #1783, opened 2015-03-10
ES6 destructuring & ractive.data


timoxley:
Now that ractive.data is deprecated, this a step backward for ractive ES6 friendliness, as one can no longer use destructuring to read and write data to the ractive instance. 

It also seems it's not possible to generate an object that can be destructured… I had high hopes for `ractive.get('')` but it seems to entirely ignore data passed in as properties on components:

```js
import Ractive from 'ractive'

const child = Ractive.extend({
  template: '<h1>{{name}}</h1>{{#items}}{{.}}<br />{{/items}}',
  data: {
    items: [1,2,3]
  },
  oncomplete: function() {
    let getName = this.get('name')
    let {name, items} = this.get('')
    console.log('name:', name)
    console.log('getName:', getName)
    console.log('items:', items)
    // Output:
    // name: undefined
    // getName: test
    // items: [1, 2, 3, _ractive: Object]
  }
})

new Ractive({
  el: document.body,
  template: '<child name="{{name}}"></child>',
  components: {child},
  data: {
    name: 'test'
  }
})
```

![image](https://cloud.githubusercontent.com/assets/43438/6573724/aab425f2-c758-11e4-986d-45a03f649e5c.png)

Rich-Harris:
I can't find the issue right now, but the plan is to include mappings (which includes non-static component attributes like `name='{{name}}`) and computations in `this.get()` (or `this.get('')`, so this *would* work:

```js
let { name, items } = this.get();
```

You can still use destructuring to write data:

```js
this.set({ name, items });
```

martypdx:
I've got a start for this on #1785. There are a couple of questions that need to be resolved:

* Should a root get from the template also be inclusive of computations and mappings? My incliniation is yes:

```html
{{JSON.stringify(.)}}
```

* Should non-computed property computations (expressions) be included? They are in the `computations` hash, but is that cool or confusing? (Leaning towards later):

```js
{
  "${JSON-stringify()}": "{\"foo\":\"foo\"}",
  "bar": "foobar",
  "foo": "foo"
}
```

timoxley:
> Should a root get from the template also be inclusive of computations and mappings? 

Vanilla data is easiest to rationalise. Including "rendered" computations means you need to be aware of any dependencies in the computation, e.g. consider this computation:

```js
new Ractive({
  data: {
    users: [new ValidUser('alice'), new InvalidUser('bob')]
  },
  computed: {
    validUsers(): {
[...]

PR #1785, opened 2015-03-10
Include mappings and computations on root get

martypdx:
~~Some unanswered questions remain, see #1783~~

Only mappings are included on root gets from `ractive.get()` and when used as an input to an expression, `{{ JSON.stringify(this) }}`.

### DIFF
--- a/src/viewmodel/prototype/get.js
+++ b/src/viewmodel/prototype/get.js
@@ -10,7 +10,8 @@ export default function Viewmodel$get ( keypath, options ) {
 		computation,
 		wrapped,
 		captureGroup,
-		keypathStr = keypath.str;
+		keypathStr = keypath.str,
+		key;
 
 	options = options || empty;
 
@@ -60,6 +61,17 @@ export default function Viewmodel$get ( keypath, options ) {
 
 	if ( !options.noUnwrap && ( wrapped = this.wrapped[ keypathStr ] ) ) {
 		value = wrapped.get();
+	}
+
+	if ( keypath.isRoot && options.fullRootGet ) {
+		// this.mappings.forEach( m => value[ m ] = m.getValue() );
+		for ( var key in this.mappings ) {
+			value[ key ] = this.mappings[ key ].getValue();
+		}
+
+		for ( var key in this.computations ) {
+			value[ key ] = this.computations[ key ].get();
+		}
 	}
 
 	return value === FAILED_LOOKUP ? void 0 : value;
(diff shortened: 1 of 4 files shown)


src/viewmodel/prototype/get.js
```diff
@@ -62,6 +63,12 @@ export default function Viewmodel$get ( keypath, options ) {
 		value = wrapped.get();
 	}
 
+	if ( keypath.isRoot && options.fullRootGet ) {
+		for ( key in this.mappings ) {
+			value[ key ] = this.mappings[ key ].getValue();
+		}
```
Rich-Harris: won't this mutate the `viewmodel.data` object? do we need to copy it first?

martypdx: Yes, good point. Does that mean it needs to happen prior to the wrapper so that this would work:

```js
var data = ractive.get();
data.foo = 'bar';
```

Or we expose as explicit option, or maybe it's own method and not guess about it:

```js
// how would an option work?
var data = ractive.get({ mappings: true });
var data = ractive.get( '', { mappings: true });

// what happens here?
var data = ractive.get( 'foo', { mappings: true });

// maybe just a dedicated method would be best: 
var data = ractive.getAll();
```

I'm leaning towards the last so we stay out of all the guess work. Thoughts?

Rich-Harris: Having a separate method would certainly make *our* lives easier! Not sure about users though.

The question for me is whether we should expect this to work or not:

```js
var data = ractive.get();
data.foo = 'bar'; // could argue that this is unsupported, shd be ractive.set('foo','bar')
```

But maybe that's a widespread pattern and we shouldn't break it. Not sure.

Rich-Harris: ...having said that, maybe mutating the data object isn't harmful. If we're discouraging people from interacting directly with it anyway, then perhaps it's okay, since `ractive.get('mappedKeypath')` will always be intercepted by the mapping anyway, there doesn't seem to be any risk of breaking anything else. Hmm

martypdx: Nothing great has come to me on this. We could modify the magic adaptor so it recognizes the `''` keypath and creates wrappers for mappings. Then everything works as expected for magic mode.

Non-magic mode is a headache because the non-object mapping values are stale once we return whatever we return:

```js
// mapping 'foo' currently 'bar'
var data = ractive.get();
data.foo; // 'bar'
ractive.set('foo','qux');
data.foo; // still 'bar' :(
```

Which is also why I don't think we can append the mapping keys to the data object and need to make a copy.

Issue #1592, opened 2014-12-18
'*' observer not firing on initialization for computed properties


martypdx:
See console at http://jsfiddle.net/cf6mmc6r/.

```js
var r = new Ractive({
    el: document.body,
    template: '#template',
    data: {
        foo: 'bar'
    },
    computed: {
        bar: '${foo} + "bar"'
    },
    oninit: function(){
        this.observe('*', function(n,o,k){
            console.log(k, n)
        })
    }
})

r.set('foo', 'qux')
```

PR #1790, opened 2015-03-10
test and fix for #1592

Rich-Harris:
@martypdx does this seem like a good fix for #1592?


martypdx:
:+1:  though I can't remember how and why I discovered it :)

PR #1794, opened 2015-03-11
fix various splice-related bugs

Rich-Harris:
see #1742 for background

### DIFF
--- a/src/shared/getNewIndices.js
+++ b/src/shared/getNewIndices.js
@@ -50,6 +50,13 @@ export default function getNewIndices ( array, methodName, args ) {
 		newIndices.push( i + balance );
 	}
 
+	// there is a net shift for the rest of the array starting with index + balance
+	if ( balance !== 0 ) {
+		newIndices.touchedFrom = spliceArguments[0];
+	} else {
+		newIndices.touchedFrom = array.length;
+	}
+
 	return newIndices;
 }
 
--- a/src/viewmodel/prototype/smartUpdate.js
+++ b/src/viewmodel/prototype/smartUpdate.js
@@ -23,7 +23,7 @@ export default function Viewmodel$smartUpdate ( keypath, array, newIndices ) {
 	if ( oldLength !== array.length ) {
 		this.mark( keypath.join( 'length' ), implicitOption );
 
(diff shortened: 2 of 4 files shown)


evs-chris:
:+1: LGTM!

martypdx:
Soo is the `{{foo.0}}` case fixed because of the `touchedFrom` or the `ReferenceExpressionResolver` bug (the later wouldn't make any sense to me, but just checking...)?

Rich-Harris:
Yep, `touchedFrom` - `foo.0` is being notified correctly now

evs-chris:
Should be the `touchedFrom` because without it, only the last index (or indices if you `unshift`ed multiple things) would get `mark`ed.

Rich-Harris:
Can either of you remember the history to `noCascade`? While looking into this whole thing just now, I removed it without any tests failing. Is it doing something important performance-wise that isn't represented in the test suite?

evs-chris:
I think it predates my time, but it looks like it was originally intended to keep computations from running multiple times (see `applyChanges`). I remember there being a comment speculating that it was no longer necessary since computations are lazy now?

Rich-Harris:
In which case I'm tempted to remove it... any thoughts @martypdx?

martypdx:
Introduced via #1282, `noCascade` keeps [`applyChanges`](https://github.com/ractivejs/ractive/blob/dev/src/viewmodel/prototype/applyChanges.js#L35) from cascading down into the computations dependent on the properties of the array item (because it was just an array modification).

Rich-Harris:
@martypdx thanks, good sleuthing. The test added in #1282 ('Array elements removed via *action* do not trigger updates in removed sections') is still green, so maybe @evs-chris was right about lazy computations?

martypdx:
> so maybe @evs-chris was right about lazy computations

Functionally yes. But lazy computations aren't completely free. What if it's an object with many expressions under it? I'd try the box update perf test with and without to get an idea of impact (if any) of one computation/expression.

Rich-Harris:
There's no difference between with `noCascade` and without in the case of the box test. Best thing is probably to leave it out of this PR and I'll create a separate issue, we can try and get a definitive answer later. Will merge this and #1795

PR #1795, opened 2015-03-11
Unwanted context

Rich-Harris:
See #1742 for background

### DIFF
--- a/test/__tests/arrays.js
+++ b/test/__tests/arrays.js
@@ -214,6 +214,69 @@ test( 'Event handlers in inside iterative sections should be rebound correctly',
 	t.equal( ractive.fragment.items[0].fragments[2].items[0].eventHandlers[0].keypaths.length, 1 );
 });
 
+test( "Nested sections don't grow a context on rebind during smart updates #1737", t => {
+	let ractive = new Ractive({
+		el: fixture,
+		template: '{{#each outer}}{{#each inner}}{{@keypath}} {{#if .foo || some.prop > 3}}<span>{{@keypath}}</span>{{/if}}<br/>{{/each}}{{/each}}',
+		data: { outer: [ { inner: [ { foo: true }, 1 ] } ], some: { prop: 10 } }
+	});
+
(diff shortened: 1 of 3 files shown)


evs-chris:
:+1: LGTM, but I'm a little surprised that the second test case passed... I could have sworn I saw the fragments being torn down during shuffle. I thought that was what was causing the various 'error unrendering the not-yet-rendered fragment' and 'attempted to remove a dependant that was no longer registered' errors. I did get several issues conflated there though, so that may be it.

Rich-Harris:
Actually I think I might have got them conflated - the test in question should probably belong with #1794, now that I think about it. It passes, in any case - the bug with the reference expression resolver not rebinding correctly looks like it was to blame. (I'm a little confused about how it worked in the first place, but I'm at the point where I'm happy to not ask too many questions)

test/__tests/arrays.js
```diff
@@ -214,6 +214,69 @@ test( 'Event handlers in inside iterative sections should be rebound correctly',
...
+	t.htmlEqual( fixture.innerHTML, 'foofoo_exprfoo<span ok class="foo">foo</span>' );
+
+	let iter = ractive.fragment.items[0].fragments[0],
+		ref = iter.items[0],
+		exp = iter.items[1],
+		mem = iter.items[2],
+		el = iter.items[3];
+
+	// make sure these little suckers don't get re-rendered
+	ref.node.textContent += 'a';
```
martypdx: Hang on, I don't think these tests are correct. By setting the textContent of the node, the textNode to which the interpolator is bound becomes irrelevant. these need to:

```js
ref.node.childNodes[0].textContent += 'a';
```

Rich-Harris: `ref`, `exp` and `mem` are all instances of `Interpolator` - the `node` in each case is just a text node which doesn't have any children, so it seems legit to me unless I'm missing something?

martypdx: Gotcha, I thought they were the `<span>`'s, but that's `el = iter.items[3]`.

PR #1806, opened 2015-03-13
fixes #1665

### LINKED ISSUE #1665: Calling pop on a multi select bound value fails
In this fiddle:
http://jsfiddle.net/sabob/5fd8a3q7/
is a Select set to multiple values and bound to the "selectedValues" array property.

If no value of the Select is picked and pop() is called on the empty selectedValues array, Ractive throws an error:
uncaught TypeError: Cannot read property 'length' of null

Could be a bug?

Rich-Harris:


martypdx:
I was just about to put in an issue for `unshift`, but this fixes it :)

```js
	let array = [];

	new Ractive({
		template: '{{#array}}{{>item}}{{/}}',
		data: { array },
		partials: { item: '{{this}}' },
		modifyArrays: true
	});

	expect( 0 );
	array.unshift(1);
```

Rich-Harris:
my timing is impeccable :sunglasses:

PR #1810, opened 2015-03-13
See if appveyor will succeed on the second run too

evs-chris:
Just curious...

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -20,7 +20,8 @@ test_script:
   - node --version
   - npm --version
   # run tests
-  #- npm run build
+  - npm run fakebuild
+  - npm run build
 
 # Don't actually build.
 build: off
--- a/package.json
+++ b/package.json
@@ -65,6 +65,7 @@
     "test": "sh ./scripts/test.sh",
     "start": "node_modules/.bin/gobble",
     "build": "sh ./scripts/build.sh",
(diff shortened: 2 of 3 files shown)


Rich-Harris:
looks like gobble is having a panic attack. Wonder if it's because a load of temporary files are being created or something, and if so why that doesn't seem to be the case when running in a VM locally.

evs-chris:
I think when the queue eventually flushes, that last build script change will work for appveyor. I don't think it's really gobble's fault... more like something whacky about the windows filesystem. I've *only* managed to reproduce what happens on appveyor once on my machine, just after a clean checkout and fresh `npm install`. Subsequent attempts were met with horrible success.

Rich-Harris:
Seeing the reports roll in from Appveyor and Travis has been like watching the world's nerdiest comedy. [Yakety Sax](https://www.youtube.com/watch?v=MK6TXMsvgQg) is playing in my head

evs-chris:
Yeah... I think I've been chased through every entry in the corridor now...

martypdx:
> Yakety Sax is playing in my head

Is Travis the white llama? or the black?

Rich-Harris:
:laughing: the white llama was the quicker of the two, so Appveyor is definitely black

evs-chris:
Hah! The llamas have been captured!

I hadn't seen that one yet, so thanks for that.

Issue #1793, opened 2015-03-11
Exceptions thrown in computed data are silenced


Skyrpex:
That's it.

Any error or exception thrown inside computed data is silently discarded. This means that the computed data is never updated, and you can't really know why...

http://jsfiddle.net/vvp4ug11/2/

Is there any workaround for this? Thanks.

martypdx:
add `debug: true` to your options, or `Ractive.defaults.debug = true;` And you will see the error on the console.

Rich-Harris:
I think it would make sense for `debug` to be `true` by default, then it's something you have to explicitly turn off in production (if that's what you want). Anyone object?

evs-chris:
I think that would probably be a good thing, but I'd definitely add a big bold statement to the release notes... Also, perhaps making the full exception available on the console would be nice? Maybe the `toString` of the failed function in the case of computations if the stack trace isn't particularly useful. Hmmmm... that actually sounds a little in-depth for a defaulted-to-on setting. Perhaps some sort of extra-detail debug flag would be useful? The defaulted on debug statements could refer to the extra flag to increase visibility?

I know I've had computations failing that I didn't notice until an embarrassing amount of time had elapsed.

martypdx:
> I think it would make sense for debug to be true by default, then it's something you have to explicitly turn off in production (if that's what you want). Anyone object?

That makes me nervous. Could we make `ractive.min.js` was `false` and `ractive.js` `true`, or is that weird?

martypdx:
> I know I've had computations failing that I didn't notice until an embarrassing amount of time had elapsed.

I think you could argue`debug: false` should be console error (which would be noticed by devs who have debug false) and throw the error if `debug: true`

Skyrpex:
@martypdx that sounds pretty good to me

evs-chris:
> Could we make ractive.min.js was false and ractive.js true,

@martypdx that's brilliant! I think using minification as flag is way better than an extra parameter.

How's this for minification detection?:

```js
Ractive.isMinified = (
  (function() {
    // do any minifiers leave single line comments?
  }).toString()
).indexOf('//') === -1;
```

That should work for code that gets minified outside of our build as well. Could we also use that in combination with the debug flag to turn down some chattiness automatically for 'production' mode?

aharpervc:
What is the rationale for swallowing exceptions in production? Exceptions should never be swallowed. If that means turning this option on in production, so be it.

[...]

PR #1812, opened 2015-03-13
encourage use of data functions with Ractive.extend()

Rich-Harris:
The idea here is to ensure that people don't run into weird bugs in which an instance of a component accidentally mutates that component's default data, by suggesting the use of data functions instead of data objects.

I'm trying something slightly different here - printing example code to the console (ES6 template strings ftw!). Takes up a few lines, but is maybe more useful than trying to explain things, or send people to an explainer page on the docs.

screen shot 2015-03-13 at 4 13 03 pm

Any thoughts?

### DIFF
--- a/src/Ractive/config/custom/data.js
+++ b/src/Ractive/config/custom/data.js
@@ -1,4 +1,5 @@
 import { fatal, warn, warnOnce } from 'utils/log';
+import { isObject, isArray } from 'utils/is';
 
 function validate ( data ) {
 	// Warn if userOptions.data is a non-POJO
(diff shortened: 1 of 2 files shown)


martypdx:
Love including the examples.

SInce the console functions no longer consider debug state (I think), would this mean if someone wanted to do this their app would forever console warn?

evs-chris:
:+1: I'm pretty sure I still do the wrong thing half of the time.

Would definitely prefer it didn't warn outside of debug (non-minified? :wink:) mode.

Rich-Harris:
Yeah, this seems like another good reason for `Ractive.DEBUG` (#1793). Maybe we implement that, then replace `warn` and `warnOnce` with `alwaysWarn`, `alwaysWarnOnce`, `warnIfDebug`, `warnOnceIfDebug`? (and maybe `fatalIfDebug` if we find we need it)

How does that sound? Suggestions for better names welcome...

evs-chris:
My $0.02:

To warn, no matter the mode: `warn`, `warnOnce`
To warn if only in DEBUG: `warnIfDebug`, `warnOnceIfDebug`
To warn or throw the error in DEBUG: `warnOrThrowIfDebug`

Having now repeated 'warn' a dozen or so times in my head, it seems a very strange word.

martypdx:
I still think we should add the ractive instance to the console wrapper functions, i.e.:

```js
warn( ractive, 'this is why I'm warning you' );
```

Two reasons:

1) I think it would reduce the number of methods (maybe not) because the debug state is implicit,
2) We can add richer log output by identifying info about the ractive instance. Certainly easy things like `ractive.component ? ractive.component.name : 'root'`, but also the `ractive.el` or `ractive.find('*')` which will cause the component to light-up in the browser on hover so you can see exactly where the problem is (assuming it made it to render).

Rich-Harris:
@evs-chris like it, will go with those. Though the more I think about it the more I wonder if there's a real use case for 'warn in prod, throw in dev' - part of me things that an error is an error and the only difference between dev and prod should be the verbosity of the feedback we give. Do you guys agree?

@martypdx yes, I think that's a really nice idea. Am currently poking around in this stuff, so I'll implement it and see what happens

constantx:
:+1:

martypdx:
> part of me things that an error is an error and the only difference between dev and prod should be the verbosity of the feedback we give. Do you guys agree?

Just go with it and try it out :)

PR #1813, opened 2015-03-13
replace ractive.debug with Ractive.DEBUG (#1793)

Rich-Harris:
**not ready for merge**

This PR replaces the per-instance (or per-component) `debug` option with a much simpler-to-understand global `Ractive.DEBUG` flag. See #1793 for the discussion that led here.

Remaining tasks:

* [x] Print a message to the console saying 'we're in debug mode, disable it with Ractive.DEBUG = false' OWTTE. (This can't happen immediately on load, because the app developer wouldn't have a chance to disable debug mode before that line was reached. We could `setTimeout` it or, perhaps better, print that on the occasion of the first `new Ractive()`. Suggestions welcome)
* [x] Decide whether it should be `true` by default, or `true` by default if unminified, or what. Right now it's `true` by default (personally I think disabling it should always be an explicit choice on the part of the developer, since many devs might download `ractive.min.js` from the CDN and put that in their `<script>` tag - because they've been told not to use unminified code - and would never see the helpful debugging messages. Though the alternative does imply either having a build step that injects environment variables into your bundle, or manually changing `Ractive.DEBUG` to `false` before each deployment)
* [x] Add a way to pass through a Ractive instance or a specific node to the various warning functions, so that we can print useful stuff to the console

### DIFF
--- a/gobblefile.js
+++ b/gobblefile.js
@@ -29,8 +29,8 @@ lib = (function () {
 		.replace( '${commitHash}', process.env.COMMIT_HASH || 'unknown' );
 
 	var bundleTransform = function ( src, path ) {
-		if ( /Ractive\.js$/.test( path ) ) {
-			return src.replace( '${version}', version );
+		if ( /(Ractive\.js|Ractive\/initialise\.js)$/.test( path ) ) {
+			return src.replace( '<@version@>', version );
 		}
 
 		return src;
--- a/gobblefile.js
+++ b/gobblefile.js
@@ -29,7 +29,7 @@ lib = (function () {
 		.replace( '${commitHash}', process.env.COMMIT_HASH || 'unknown' );
 
 	var bundleTransform = function ( src, path ) {
-		if ( /(Ractive\.js|Ractive\/initialise\.js)$/.test( path ) ) {
+		if ( /(Ractive\.js|utils\/log\.js)$/.test( path ) ) {
 			return src.replace( '<@version@>', version );
 		}
 
--- a/src/Ractive.js
+++ b/src/Ractive.js
@@ -22,6 +22,9 @@ Ractive = function ( options ) {
 // Ractive properties
 properties = {
 
+	// debug flag
+	DEBUG:         { value: true },
+
 	// static methods:
 	extend:        { value: extend },
 	getNodeInfo:   { value: getNodeInfo },
--- a/src/Ractive.js
+++ b/src/Ractive.js
@@ -38,7 +38,7 @@ properties = {
 	magic:         { value: magic },
 
 	// version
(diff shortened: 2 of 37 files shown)


Rich-Harris:
Actually now that I think about it, devs wouldn't need to be able to deal with environment variables in their build step, they could just use @evs-chris' trick in their own apps (and we could even show them how in the initial 'welcome' message):

```js
// app.js
import Ractive from 'ractive';

Ractive.DEBUG = (function () {
  // unminified
}).toString().indexOf( '//' ) !== -1;

// or even...
Ractive.DEBUG = Ractive.isMinified;
```

Rich-Harris:
Have found there isn't really a need for `warn` and `warnOnce` - every case I found was better replaced by `warnIfDebug`/`warnOnceIfDebug`. I'll leave them put for now, but it may be the case that we're better off only exposing the `...IfDebug` versions (in which case they probably don't need that suffix).

Rich-Harris:
The last couple of commits implement the initial 'welcome' message. Too chunky? The links are clickable, which is a nice touch on the part of devtools developers.

screen shot 2015-03-13 at 6 39 21 pm

evs-chris:
:+1: I like it.

martypdx:
:clap: 

surprised you didn't dip into the css enhancements :)

Rich-Harris:
> surprised you didn't dip into the css enhancements :)

I thought about it! But I decided that embedding the logo as a data URI probably wasn't the best use of bytes... also it comes out wonky in some browsers, and there's no way (that I know of) to detect support.

Anyway. You can now pass through an options object as the final argument to `warn[Once][IfDebug]()`, with `ractive` and `node` properties (both optional), and if possible the message will be augmented accordingly - it'll be preceded with `<component-name>` if there is one and we know what it is, and followed by `options.node` or `options.ractive.find('*')` if it's been rendered. It turns out there aren't too many situations where those are applicable (a lot of the warnings happen when instances are still going through their init cycle, for example), but maybe more will show up now that we have the ability.

@evs-chris what do you reckon to defaulting `Ractive.DEBUG` to `true` but keeping that code snippet visible? Seems like the best of both worlds - app developer choice but without convoluting build processes. Incidentally I found an even slimmer way to detect minification:

```js
Ractive.DEBUG = /unminified/.test(function(){/*unminified*/});
```

evs-chris:
Sounds good to me. I especially like your round of minification-detect golf!

anton-ryzhov:
Looks like you've made `Ractive.DEBUG` read-only property.

Rich-Harris:
/slaps forehead

fixed, thanks @anton-ryzhov

Issue #1788, opened 2015-03-10
Ractive Promise Error Handling


timoxley:
Chrome does a godawful job of rendering errors (note stack trace not visible)
![image](https://cloud.githubusercontent.com/assets/43438/6583799/93248b54-c7a1-11e4-8969-71c29ef67fdf.png)

You can improve things with `console.error(err.stack)`, but this isn't standard.
![image](https://cloud.githubusercontent.com/assets/43438/6583831/c819aaec-c7a1-11e4-9e4c-5d0c723308e9.png)

This sucks but most problematic about the current state of ractive error handling is the debugger is mostly useless (even with the `async` flag on, that feature apparently not perfect).

After removing all of ractive's promise error handling code: `%s/.then(null, consoleError)//gc `, now Chrome will handle unhandled promise rejections pretty much as if they were thrown exceptions:

![image](https://cloud.githubusercontent.com/assets/43438/6583907/49cfba86-c7a2-11e4-9046-998f0c2e5365.png)

if dev tools is open it will even pause on the correct line.

![image](https://cloud.githubusercontent.com/assets/43438/6583880/29b5f030-c7a2-11e4-9bb4-072e0c102144.png)

Prior to a very recent of Chrome this was not the case and you certainly needed to have some kind of manual catch-all. 

Suggest adding a flag to disable these. Ideally this is outside the promise flow so it doesn't mess with stack traces.

martypdx:
Thanks for the info @timoxley, we put those in because error was being completely swallowed. Since then we've also we fixed the fact that the native `Promise` implementation was not being used. 

Perhaps we should use the `.then(null, consoleError)` when we're not using the native Promise.

timoxley:
> we put those in because error was being completely swallowed.

Yeah I recently had a 2 day long wild goose chase trying to figure one of those out. Absolute horror. Very relieved when Chrome added logging.

> Perhaps we should use the .then(null, consoleError) when we're not using the native Promise.

Yes this would be great, though even in that scenario it'd be nice if there was an option to let it throw instead of logging e.g. dev mode

martypdx:
> Yes this would be great, though even in that scenario it'd be nice if there was an option to let it throw instead of logging e.g. dev mode

But without the Chrome intercept of a native promise, isn't the issue that our own-rolled Promises call `then` async from the `setTimeout` ethers, and so throwing went up the stack to where all the lost socks from dryer go, never to be heard from again.

Rich-Harris:
[...]

PR #1815, opened 2015-03-14
BREAKING CHANGE: change promise rejection handling behaviour (#1788)

Rich-Harris:
See #1788. With this PR, asynchronous errors that happen inside `.render()` and `.animate()` aren't dealt with by Ractive, but must be dealt with by a `.catch()` handler, regardless of `Ractive.DEBUG`.

The exception is an automatic render (i.e. where `new Ractive()` is passed an `el` property, or it's an inline component) because then there's no way for the developer to get hold of the promise. In that case, if `Ractive.DEBUG_PROMISES` is `true` (the default), a message/stack will be logged:

screen shot 2015-03-14 at 9 58 34 am


martypdx:
Are there browsers that have native implementations that Ractive would use to that don't handle errors sufficiently?

Rich-Harris:
AFAIK it's only Chrome that logs uncaught errors (so far) - in other browsers they'll get swallowed. So if you do this in Firefox or Safari...

```js
ractive.set( 'foo', 'somethingThatWillCauseAnErrorSomehow' );
```

...then in theory you'd need to add a `.catch()` handler. In practice though a) I'm *fairly* sure that anything that could cause an error inside those promise-returning methods would happen outside the promise itself, so the error wouldn't get swallowed, and b) we couldn't do anything in these situations anyway:

```js
ractive.set( 'foo', 'somethingOkay' )
  .then( throwAnError );
```

martypdx:
I just wonder if we could simplify the default logic to: `Ractive.DEBUG_PROMISE = Ractive.DEBUG && !hasNativePromises;`

Rich-Harris:
Surely that would just mean that Firefox et al wouldn't get the extra debugging help if initial render failed? If there was a way to detect whether the browser printed unhandled promise rejections, that'd be ideal, but AFAICT there isn't a way to do that, hence erring on the side of verbosity.

Rich-Harris:
will merge for the sake of 0.7 - we can revisit the details in future if we decide we got them wrong

PR #1817, opened 2015-03-14
dont create branches on new two-way bindings - fixes #1570

### LINKED ISSUE #1570: Ractive won't resolve references with manipulation of index, nor with aliasing (2-way binding or pass to component)
Here's a problem I ran into:
Reference like this `<myComponent fieldValue="{{data[multipleIndex+topRecordIndex][ref]}}” />`  or `<input value="{{data[multipleIndex+topRecordIndex][ref]}}" />` will not be updated on change of indexes when passed to component or for field 2-way binding.

http://jsfiddle.net/yesrmLqu/3/ works great until you add 
`<input value="{{ arr[arrIndex+otherIndex].name }}" />`
The result is: http://jsfiddle.net/yesrmLqu/4/ **Warning! WIll hang**

So I tried aliasing as a work-around but this wouldn't work either:

```
{{# {rowIndex: multipleIndex + topRecordIndex } }}
  <myComponent value="{{data[rowIndex][ref]}}" />
{{/}}
```

Also, I don't know if it's by design or not, but aliasing would not 2-way bind:

```
{{# {something: somethingelse} }}
<input value="{{something.field}}" />  Would show value but not update somethingelse
{{/}}
```

I would appreciate any hints!

Rich-Harris:
See #1570

### DIFF
--- a/src/viewmodel/prototype/set.js
+++ b/src/viewmodel/prototype/set.js
@@ -41,7 +41,7 @@ export default function Viewmodel$set ( keypath, value, options = {} ) {
 	}
 
 	if ( !computation && !keepExistingWrapper ) {
-		resolveSet( this, keypath, value );
+		resolveSet( this, keypath, value, options.createBranches !== false );
 	}
 
 	if ( !options.silent ) {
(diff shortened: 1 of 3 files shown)


Rich-Harris:
Closing based on discussion on #1570 - this isn't a very good idea

PR #1828, opened 2015-03-15
Collapse welcome message to minimize spaminess

martypdx:
The new debug welcome message take up a lot of real estate which gets tres annoying when your running live reload or otherwise refreshing browser. This reduces it to:

image

The full message is available if you expand the group (which works by clicking anywhere on that line).

image

`console.groupCollapse` has [good browser support](https://developer.mozilla.org/en-US/docs/Web/API/Console/groupCollapsed#Browser_compatibility) and this falls back to `log` if not.

We might consider this pattern for some of the other lengthy messages that have code snippets, etc. attached to them.

### DIFF
--- a/src/utils/log.js
+++ b/src/utils/log.js
@@ -6,6 +6,13 @@ import noop from 'utils/noop';
 var alreadyWarned = {}, log, printWarning, welcome;
 
 if ( hasConsole ) {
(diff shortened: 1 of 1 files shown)


Rich-Harris:
:+1: that's a really nice touch

PR #1829, opened 2015-03-15
Fixes #1805 - Inability to pass data to a Component's computed properties on instantiation

### LINKED ISSUE #1805: Cannot map to a computed property
Recently tried my project using `dev` and started receiving the new validation Error for `Cannot map to a computed property`.

I have some use-cases where this is a useful feature (my property is a getter and _setter_, so the data being passed in was previously being modified by the setter on the component's initialization before being stored under a different underscore prefixed key for later retrieval)... Curious what the rationale was behind this validation and whether there's a valid use-case here to support.

For now I can work around it with handlers in `oninit`, but I enjoyed the ease / cleanliness of having a computed setter ingesting data that needed some modification.

ElliotChong:
Fixes #1805 - Inability to pass data to a Component's computed properties on instantiation

### DIFF
--- a/test/__tests/components.js
+++ b/test/__tests/components.js
@@ -732,7 +732,9 @@ test( 'Mapping to a computed property without a setter is an error', function ( 
 	t.throws( function () {
 		new Ractive({
 			template: '<widget foo="{{bar}}"/>',
-			data: { bar: 'irrelevant' },
+			data: {
+				bar: 'irrelevant'
+			},
 			components: {
 				widget: Ractive.extend({
 					computed: {
@@ -746,16 +748,22 @@ test( 'Mapping to a computed property without a setter is an error', function ( 
 	}, /Cannot map to a computed property \('foo'\)/ );
 });
 
-test( 'Mapping to a computed property with a setter does not throw an error', function ( t ) {
-	new Ractive({
+test( 'Mapping to a computed property with a setter does not throw an error (#1805)', function ( t ) {
+	var ractive;
+
+	ractive = new Ractive({
+		el: fixture,
 		template: '<widget foo="{{bar}}"/>',
-		data: { bar: 'irrelevant' },
+		data: {
+			bar: 'irrelevant'
+		},
 		components: {
 			widget: Ractive.extend({
+				template: "<p>widget: {{foo}}</p>",
 				computed: {
 					foo: {
 						get: function () {
-							return "Win";
+							return this.get("_foo");
 						},
 						set: function (value) {
 							this.set("_foo", value);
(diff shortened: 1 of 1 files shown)


Skyrpex:
Will this be merged anytime soon?

Rich-Harris:
@Skyrpex unfortunately it's sort of blocked by #1740, which is a huge chunk of work that's going to change how a lot of this stuff works under the hood - I'm nervous about us adding anything like this until the dust settles a bit. @martypdx may have a different answer though!

Is this blocking you? If you can share details of your scenario there may be a workaround we can come up with

martypdx:
@Skyrpex Looking back on your example in #1805, the issue there is one of restricted references. You don't need to add the `@keypath`, just use `.valid`:  http://jsfiddle.net/mo0vvb92/.  (In `0.8` it looks like forced binding would be to the innermost context, so it would work without the `.`)

It's totally fine for a component to set and manipulate data in the parent. I believe what @ElliotChong is looking for in this PR is to pass an initial value to a computed property that has a setter. 

But are you also wanting the setter to continue to be called if the outside value changes?

What I think doesn't make sense is to expect the value computed by the getter to be propagated into the original parent value.

Skyrpex:
First of all, I want to thank you the proactivity of the Ractive team. It's awesome!

@Rich-Harris, nope! This is not blocking to me. I was just curious about this because I tried to pass an initial state to a computed property. The computed setter wasn't called, and the property passed to be a normal data variable.

@martypdx, in my example, I wanted to use a component to perform input validation. Here's an example of what I did: http://jsfiddle.net/skyrpex/rpLwme6x/1/. I suppose that computed properties aren't made for that purposes. Using observers is the right way to go: http://jsfiddle.net/skyrpex/rpLwme6x/2/

By the way, @Rich-Harris, can I add component examples to the wiki? It could be a nice place to explore best-practices!

fskreuz:
@Skyrpex There's the component-spec repo (https://github.com/ractivejs/component-spec/blob/master/authors.md) which has one. There's an open issue (https://github.com/ractivejs/docs.ractivejs.org/issues/179) to move relevant documentation into the repo. :D

PaulMaly:
Is this feature already in release?

martypdx:
@PaulMaly I don't believe so.

I go back and forth between thinking  we should just not allow Computed properties to shadow data properties, and thinking it's ok for a computed property with a setter to be initialized by a data property of the same name.

In either case, realize that if you just _use a different name_ you can do whatever you want, AFAICT this is all about a computed property and a data property _of the same name_.

PaulMaly:
Hello martypdx. 

I just try to bind my parent component property and computed property of child component. I write advanced form component which can have some fields like basic fields and some fields like more complex custom fields (sub-components). Like this:

```
{{#fields}}
<input value="{{.value}}">
....
<customfield value="{{.value}}"></customfield>
{{/fields}}
```

So, for basic fields two-way binding work properly. For custom field component value property is computed (specific function to get result value of custom field). But this code not working with an error: 

> Uncaught Error: Cannot map to a computed property ('value')

evs-chris:
The primary issue this is addressing is fixed as of 0.9, and #2933 is open for bindings on custom elements.

Issue #1824, opened 2015-03-14
Inline partials discrepancy


Rich-Harris:
Came across this while updating the [template spec](https://github.com/ractivejs/template-spec): http://jsfiddle.net/rich_harris/eh4kqxd6/

Presumably, inline partials should be attached to their parent component, rather than all sharing the top level, and there shouldn't be a difference between parsed-at-instantiation versus pre-parsed.

Possibly a 0.7.1 fix

evs-chris:
If you pass a Ractive instance to `parse` (`Ractive.parse(str, { ractive: myInstance })`), it will spit out a template object with the partials anchored to the right elements. It should definitely be possible to use `Ractive.defaults.components` though, and that isn't currently happening.

I think there is a better solution, though it adds a bit of expense to using partials, so it may not be worth it. That would be to just attach inline partials to the nearest element at parse time and then stick them to their components just before they are used. That way any pre-parsed template could be used with any set of components and the right thing should happen in each case.

Rich-Harris:
I think that's the right approach (attaching to nearest element at parse time). Just tinkering a bit with that now

PR #1835, opened 2015-03-16
Store inline partials at appropriate locations in parsed template

Rich-Harris:
See #1824. Decided this really ought to be in 0.7.0. It necessitates a few minor tweaks to how parsing happens, but the end result is that partials are store either at the top level, or as belonging to elements (by which we mean components, but there's no way to differentiate at parse time).

In addition, the old `<!-- {{>somePartial}} -->` syntax is deprecated in favour of `{{#partial somePartial}}`, and a warning will be printed (it still works, for now). Ractive will also throw an error if you have duplicate partials in the same 'scope'.


evs-chris:
Nice!

Issue #1838, opened 2015-03-16
Inline partial has gone missing


evs-chris:
http://jsfiddle.net/xmk8kod1/

```js
var ractive = new Ractive({
    el: 'body',
    template: '<div>{{#partial foo}}foo{{/partial}}</div>{{>foo}}'
});
```

I haven't had a chance to dig into it yet, I just wanted to make sure it was noted.

Rich-Harris:
This'll be related to #1824/#1835. Isn't that the desired behaviour though? If inline partials are 'scoped' to their components, then a partial that needs to be available throughout the template needs to be defined at the top level of the template: http://jsfiddle.net/rich_harris/cnffkqj0/

evs-chris:
> Isn't that the desired behaviour though?

I'm not sure... it is a bit different from the previous behavior though, which allowed inline partials anywhere and bubbled them up to the nearest component. I'll defer to your judgement on the desired behavior.

For reference, I have a CMS-like thing that assembles parts of a page with their chunks of template that may include partials for recursive things. I think I can make it work without having the partials at arbitrary levels.

martypdx:
> Isn't that the desired behaviour though?

It threw me at first, but makes sense. Does it means you can do:

<div>
  {{#partial foo}}one{{/partial}}
   <span>{{>foo}}</span>
</div>
<div>
  {{#partial foo}}twp{{/partial}}
   <span>{{>foo}}</span>
</div>

Needs to be document as _element_ scope right?

Rich-Harris:
ooh, good question. Right now, those partials would just get lost in the ether because they're not top-level, and `<div>` turns out not to be a component. But we could absolutely store them on `Element` items, and use them in `getPartialTemplate.js`. That would actually be pretty neat

martypdx:
I think you have to go all the way with this one. It doesn't make sense to say it "got lost" while still being component/instance scoped.

> That would actually be pretty neat

And in case we never understand the new mustache inheritance syntax, it would probably be swell to just do:

```html
{{#partial foo}}
Something happens {{>here}}
{{/partial}}

<div>
  {{#partial here}}one{{/partial}}
   <span>{{>foo}}</span>
</div>
<div>
  {{#partial here}}two{{/partial}}
   <span>{{>foo}}</span>
</div>
```

evs-chris:
Oh, I like that.

Rich-Harris:
> It doesn't make sense to say it "got lost"

[...]

PR #1841, opened 2015-03-17
Taking a swing at the inline partial shuffle

evs-chris:
### Part 1

~~Per #1838's discussion, this gets inline partials scoped back to their nearest Ractive instance. It does so by simply copying the partials from the element to the element root when the element is instantiated. This does have at least one potential issue that I can think of: non-rendered segments of the template can't contribute inline partials. I don't know if that's a bug or a feature. It kinda makes sense, but then the partial hangs around if the element is unrendered, which seems a little wrong. So, I could go either way on how to handle that:~~

1. ~~Read through the whole template when it is loaded and pull in any partials that don't exist down a component tree. This is roughly the same as the previous behavior.~~
2. ~~When the element is torn down, reset any partials that it contributed at init back to what they were if they haven't been updated in the interim. For partials unique to the element, this would remove them, and for partials that override, the overridden partial would be restored.~~

~~I'm leaning toward option 2 at the moment. It seems to track better with part 2 of this.~~

After discussion, it seems that just leaving partials attached to their elements and looking for partials up the fragment hierarchy is a more consistent approach.

### Part 2

Per @martypdx's suggestion, this also lets `getPartialTemplate` look up the element hierarchy for a partial before checking the usual places. This is pretty neat for doing partial overrides outside of components.

~~Something to consider here is how this should interact with `resetPartial`. Should all instances get touched, or should partials that were originally rendered from inlines end up with their inline again, or something else?~~

The current behavior is, upon reset, partials with inline overrides reset to their nearest inline. That seems to be most consistent with inline partials as full-fledged members of a template. Inline partials at root levels can still be reset, as they are treated as normal partials.

### DIFF
--- a/src/virtualdom/items/Element/prototype/init.js
+++ b/src/virtualdom/items/Element/prototype/init.js
@@ -97,11 +97,4 @@ export default function Element$init ( options ) {
 	// create transitions
 	this.intro = template.t0 || template.t1;
 	this.outro = template.t0 || template.t2;
-
-	// if there are any partials bound here, copy them to root
-	if ( template.p ) {
-		for ( let k in template.p ) {
-			this.root.partials[k] = template.p[k];
-		}
-	}
 }
--- a/src/virtualdom/items/Element/prototype/init.js
+++ b/src/virtualdom/items/Element/prototype/init.js
@@ -97,4 +97,11 @@ export default function Element$init ( options ) {
 	// create transitions
 	this.intro = template.t0 || template.t1;
 	this.outro = template.t0 || template.t2;
+
+	// if there are any partials bound here, copy them to root
+	if ( template.p ) {
+		for ( let k in template.p ) {
+			this.root.partials[k] = template.p[k];
+		}
+	}
 }
--- a/src/virtualdom/items/Partial/_Partial.js
+++ b/src/virtualdom/items/Partial/_Partial.js
@@ -29,7 +29,7 @@ let Partial = function ( options ) {
 	// (i.e. `{{>foo}}` means 'use the foo partial', not 'use the partial
 	// whose name is the value of `foo`')
 	if ( !this.keypath ) {
-		if ( template = getPartialTemplate( this.root, this.name ) ) {
+		if ( template = getPartialTemplate( this.root, this.name, parentFragment ) ) {
 			unbind.call( this ); // prevent any further changes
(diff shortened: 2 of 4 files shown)


evs-chris:
Having pondered it a bit more, I favor dropping the instance scoping all together for inline partials. It makes more sense to me to have them be scoped only to elements since they are part of the virtual dom.

In that context, it makes sense that `resetPartial` would not touch inline partials unless they are attached to an instance, which, coincidentally, is the current behavior in this PR.

TL;DR (for the whole PR) - I think inline partials at a root should behave as normal partials; inline partials at an element should only affect fragments further down the hierarchy and should be left otherwise unmolested

martypdx:
If I understand this, element partials get copied to root so they are generally available. However, a partial use will look up the vdom elements and favor partials of closest ancestor. Best of both worlds?

evs-chris:
That's what the PR does as it stands, but there are some potentially weird issues with copying the partials. For instance
```js
new Ractive({
  el: 'main',
  template: '{{#foo}}<div>{{#partial foo}}foo{{/partial}}</div>{{/}}<button on-click="toggle(\'foo\')">Go</button>{{>foo}}'
});
```
'foo' does not appear until you click the button the first time, but it remains after you click it again. So if there was already a 'foo' partial, it would be overwritten and remain so if the element that overwrote it disappeared. This could have particularly strange implications for elements in iterative sections. I suppose another solution to that would be to keep element partials from overwriting existing root partials?

martypdx:
Or we could respect "scope", which is a fairly standard model.

```html
{{#foo}}
  <div>
    {{#partial foo}}foo{{/partial}}
    {{>foo}} <!-- this find foo -->
  </div>
{{/}}
<button on-click="toggle(\'foo\')">Go</button>
{{>foo}}  <!-- causes warning because no foo here -->
```

Then no need to "transfer" to root, either you find the partial, or not, based on scope.

martypdx:
Just to add, I can't think of a case where I nested a partial in a sibling of where it was used. Maybe it's just me, but it would bug me to see it nested outside of where it was used.

evs-chris:
> Or we could respect "scope", which is a fairly standard model.

That's what I was thinking too. I've updated the PR.

Rich-Harris:
101379-jim-carrey-i-like-it-alot-gif-gflm

Issue #1833, opened 2015-03-15
Dynamic yields


evs-chris:
We currently have dynamic partials and named yields. Would it make sense to allow other partials, including those generated on-the-fly, to also be the target of a yield?

This is somewhat related to #1832, where part of the template in a component may be generated on-the-fly from data passed in and should really be owned by the container for most purposes. I think removing the inline content restriction on yields and setting them up as dynamic would allow that. If that's acceptable, it might make sense to merge yield and partial and just make yield a special case.

As a further question, related to #1379, how would you feel about another form of partial that moved a bit closer to a lightweight component? I have a bunch of components that would probably work better as a souped-up partial, like date and tab components. A component is pretty heavy for something as simple as a three-state checkbox that has 50 or 60 instances. Something that could be a target for events, host additional functions, and get involved in the reference resolution process for its own internal state or have an additional provided mapped/linked namespace would be great.

PR #1849, opened 2015-03-18
Look for methods for method-call events in parents - RFC

### LINKED ISSUE #1832: Walk up-hierarchy to find method for method-call events?
I have a tab component that is only used to control the display of chunks of template. I also have a template builder that uses said tab component and other components and form controls to build a template from a declarative chunk of JSON. When mixing those two, it's pretty easy to overlook the fact that method-call events only work on their immediate host Ractive.

Would it make sense to let those bubble (or search up-hierarchy) to the nearest Ractive with the right method instead of bailing when the tab component doesn't have it?

After I facepalm every time it happens, I work around it by using proxy events and bubbling (thanks again for that @martypdx, it's great!).

evs-chris:
I don't know if this is something that generally makes sense for most components, but I keep bumping my head on it. I thought I'd see what the general consensus is.

I see this as the equivalent of proxy event bubbling for method-call events. If an `EventHandler` root doesn't have the appropriate method to handle the event, it looks on parents until it runs out or hits an isolation boundary. The original target instance is supplied to the event as `event.component`.

Fixes #1832

### DIFF
--- a/src/virtualdom/items/Element/EventHandler/prototype/init.js
+++ b/src/virtualdom/items/Element/EventHandler/prototype/init.js
@@ -85,7 +85,7 @@ function fireMethodCall ( event ) {
 	var ractive, values, args;
 
 	ractive = findMethodHost( this.root, this.method );
-	event.origin = this.root;
+	event.component = this.root;
 
 	if ( !ractive || typeof ractive[ this.method ] !== 'function' ) {
 		throw new Error( 'Attempted to call a non-existent method ("' + this.method + '")' );
@@ -141,7 +141,7 @@ function fireEventWithDynamicParams ( event ) {
 function findMethodHost ( ractive, method ) {
 	if ( typeof ractive[ method ] === 'function' ) {
 		return ractive;
-	} else if ( ractive.parent ) {
+	} else if ( !ractive.isolated && ractive.parent ) {
 		return findMethodHost( ractive.parent, method );
 	}
 }
--- a/src/virtualdom/items/Element/EventHandler/prototype/init.js
+++ b/src/virtualdom/items/Element/EventHandler/prototype/init.js
@@ -81,13 +81,13 @@ export default function EventHandler$init ( element, name, template ) {
 	}
 }
 
-
 function fireMethodCall ( event ) {
 	var ractive, values, args;
 
-	ractive = this.root;
+	ractive = findMethodHost( this.root, this.method );
+	event.origin = this.root;
 
-	if ( typeof ractive[ this.method ] !== 'function' ) {
+	if ( !ractive || typeof ractive[ this.method ] !== 'function' ) {
 		throw new Error( 'Attempted to call a non-existent method ("' + this.method + '")' );
 	}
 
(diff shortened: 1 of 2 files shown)


Rich-Harris:
It makes me a tiny bit nervous... it seems like it could make the behaviour of apps slightly harder to predict, and means that control is flipped round - rather than parents choosing to act on the actions of their children, the children effectively control the parent, which seems dangerous.

Consider that a strong opinion weakly held though - I'd be more interested in knowing what other people think about this than in my own initial reaction

Madgvox:
Yeah, I'm not really sure I like the idea of "bubbling" method calls. To me at least, it seems like events are built to be able to take input from anywhere. Method calls feel a lot more specific and in my mind, a bubbling method call is more likely to break something.

Also, events can bubble through multiple handlers, whereas method calls can't -- or shouldn't. It feels like a break of expectations to have a function call act like an event.

evs-chris:
That's fair, and I pretty much agree. I just have a bunch of components that are mostly just for layout (tab box, popover, and the like), and my default for event handlers is method calls. It seems to be a little rough spot that I've been rubbing against a lot lately.

How about an opt-in flag on the component? (`Ractive.extend({ bubbleCalls: true, ... })`, or something along those lines)

Alternatively, I think allowing yields on non-inlined partials would help. I have a component that gets most of its content generated during its lifespan rather than at init, and it can make some interaction a little weird. I have a separate related question open for that at #1833.

Also, if anyone has any architectural/behavioral guidance for layout-y components, I'd love to hear it!

martypdx:
Can you just use event bubbling? Parent can redirect to method pretty easy:

```js
this.on('child.eventName', this.methodName.bind(this));
```

I think maybe exposing method calls on the component tag declaration would be the way to go:

```html
<child on-select='makeCall()'/>
```

martypdx:
> Alternatively, I think allowing yields on non-inlined partials would help. 

But that seems to work fine: http://jsfiddle.net/38dje3bt/.  Or am I missing something?

evs-chris:
This close to one of the scenarios I have: http://jsfiddle.net/10x5ohcx/

I do end up using bubbled proxy events, but usually after I scratch my head for a minute wondering why I'm getting method missing errors from events (I mean, gosh, they're clearly there... oh). Here's the tab component that's a good example of what I'm talkin' 'bout: https://github.com/evs-chris/ractive-tab-box

Madgvox:
This is totally unrelated, but the link to gobble-giblet in your README is broken. I think it's supposed to be gobble-giblets.

martypdx:
> This close to one of the scenarios I have: http://jsfiddle.net/10x5ohcx/

Not sure what the objective is on this one. Is the child dynamically choosing a partial for the parent???

evs-chris:
@Madgvox derp. Thanks, fixed.

> Is the child dynamically choosing a partial for the parent???

Yep. The parent hands the component a few chunks of template and says 'here, display these as appropriate'. The component then shuffles them around with user interaction. The parent may add and remove chunks of template or allow the component to remove them based on user input.

evs-chris:
It looks like this one turned out to be a bad idea, which was kinda what I felt to begin with. I think I have an idea for a Better Way<sup>**TM**</sup> that I'll try out when I get a few spare cycles.

Issue #1850, opened 2015-03-19
ractivejs bind array problems


blueforest:
The demo is [here](http://jsfiddle.net/84xp4jan/1/),a nested array bind a template,when i click"Add Tab",and then click "Add Item",it not work,why?If the "arr" does not add or delete operation,the "Add Item" work good!

Madgvox:
It looks like there's a bug of some kind with wrapped arrays. If you add `modifyArrays: false` to the config, and call `ractive.update()` when you modify the arrays, it works: http://jsfiddle.net/84xp4jan/2/

Alternatively, you can save yourself the hassle of parsing the keypath from `event.keypath` every time by passing in the item you want to modify as arguments, and then using the built-in ractive array methods to modify the arrays: http://jsfiddle.net/84xp4jan/3/

Isinlor:
I'm facing with the same issue - as for now I isolated the bug: http://jsfiddle.net/wa4hLnyx/1/ Maybe will save someone a little bit of time.

I will make a pull request if I manage to solve it.

Isinlor:
Update - it is an issue with `clearCache` and ` wrapper.teardown()`. 
It seems like this code crated a lot of confusion: https://github.com/ractivejs/ractive/blob/dev/src/viewmodel/prototype/clearCache.js

If I understand well `clearCache` is recursive and this check here:
https://github.com/ractivejs/ractive/blob/dev/src/Ractive/static/adaptors/array/index.js#L66
should stop the teardown of all nested wrappers, but it won't because of `setting` variable is set only on first level:
https://github.com/ractivejs/ractive/blob/dev/src/Ractive/static/adaptors/array/patch.js#L27

Setting `setting` to true on all nested wrappers would be an appropriate solution?

Madgvox:
@martypdx This looks like it's #1775 related.

Isinlor:
Seems like it's fixed on edge, cool :+1: .

Madgvox:
That it does.

PR #1854, opened 2015-03-20
Fix teardown of wrapper in clearCache #1850

Isinlor:
I added test case and fixed the #1850 issue, however my fix creates regression on two tests and I'm not fully sure if performance of the fix is OK. Is it worth investing more time on fixing failing test cases?


evs-chris:
This is now fixed in edge after a pretty massive refactor that no longer includes the same form of caching. Thanks!

Isinlor:
What do you think about making a pull request of only the test case? It would help assure that there won't be any regression on this one and the test is ready anyway, so it would be pity to not make use of it.

Issue #1853, opened 2015-03-19
noIntro should apply to child components


Rich-Harris:
Demo: http://jsfiddle.net/rich_harris/rhdaezxm/

martypdx:
Why not set option on that component? http://jsfiddle.net/rhdaezxm/2/

Madgvox:
Yeah, I second @martypdx. There could be situations where you would want to preserve child intros.

martypdx:
I think, lize `lazy` and `twoway`, it would be helpful to have element-level attributes. Though the name might need to change because `<div noIntro='true' intro='fade'>.` is a head scratcher.  While were at it, I've needed a `noOutro` as well...

Madgvox:
It seems like this should be a larger discussion than just intros. It would be nice to address the nested transitions issue from #1397 as well.

martypdx:
Wanted to see if you could dynamicatlly control noIntro from the parent (you can), so leaving [this link](http://jsfiddle.net/h24j7mzt/) here.

Rich-Harris:
@Madgvox is right, there is a need for a larger discussion about transitions. Their main virtue at the moment is their simplicity (and the fact that they're not restricted to CSS transitions), but there is a cost associated with that in terms of expressiveness. Handling nested transitions is one part of that discussion; so are...

* how to have better control over the timing of when elements are removed from the DOM
* how to set parameters in a non-hacky way based on the states we're transitioning from and to (e.g. if you're transitioning from the `settings > account` route to `settings`, the current page should fly out to the right, but if you're transitions to `settings > account > name`, it should fly out to the left
* in a similar vein, how to think about transitioning between states full stop. `intro` and `outro` are fairly crude. I watched @ef4's [EmberConf talk](https://www.youtube.com/watch?v=p8aF-7-_cE8&list=PLE7tQUdRKcyacwiUPs0CjPYt6tJub4xXU&index=15) about [liquid fire](https://github.com/ef4/liquid-fire) with some jealousy - there's a really great moment when speaker headshots remain on the page even as you move from one page to a completely different one. That kind of object constancy is tricky (though not impossible) to achieve with Ractive transitions as they are now.
* how to handle things like spring dynamics (a la Facebook's [Rebound](http://facebook.github.io/rebound-js/examples/))
[...]

PR #1857, opened 2015-03-21
Warn and ignore functions supplied for standard options

martypdx:
See discussion [starting here](https://github.com/ractivejs/ractive/issues/1853#issuecomment-84411137)


martypdx:
Feel down the phantomjs hole for a bit on this one.

Turns out `typeof fixture.querySelectorAll('div')` returns `function`

Rich-Harris:
Good lord. That's even more messed up than the `typeof node.parentNode === 'unknown'` I discovered in IE8 the other day

Issue #1865, opened 2015-03-23
Observers are not removed after teardown


blackgate:
0.7.1:
http://jsfiddle.net/z7ytqLhc/

0.6.1:
http://jsfiddle.net/5xmhL0g3/

As you can see in the console of the examples above, the 0.7.1 version keeps calling the observers from the old instances.

blackgate:
My last opened issue(#1861) can also be related to this.

constantx:
confirmed with 0.7.1
![screen_shot_2015-03-25_at_11_17_01](https://cloud.githubusercontent.com/assets/113969/6832005/f973cdee-d2e0-11e4-8d22-71f13f22c698.png)

constantx:
and child `r-14` in my case observe still stuck around even though the view is torn down.
![screen shot 2015-03-25 at 11 21 29](https://cloud.githubusercontent.com/assets/113969/6832040/27345cc6-d2e1-11e4-96e8-ee0ce507248a.png)

Rich-Harris:
Fixed in 0.7.2. Thanks for the test case @blackgate - you'll find that the observer is called immediately before teardown (`true -> false`), but after that it's neutered.

PR #1894, opened 2015-04-02
cancel all observers when instance is torn down (#1865)

Rich-Harris:
#1865


constantx:
@Rich-Harris will this be in 0.7.2?

[+1](https://medium.com/@Rich_Harris/this-doesn-t-have-to-be-the-front-end-cf187e558a14) :)

Rich-Harris:
> +1 :)

Thanks :)

Yep, this is included with 0.7.2, which was just released.

PR #1895, opened 2015-04-02
0.7.2

Rich-Harris:
We managed to sneak a couple of regressions into 0.7.1, so I figured we should release 0.7.2 sooner rather than later.

(Generally, frequent small releases are self-evidently better, though that's tricky with some of the changes on the roadmap - server-side encapsulated CSS rendering, progressive enhancement, #1740 and so on. So if there are any obvious major bugs or regressions that need to be dealt with, we should try and identify them for a 0.7.3 release before 0.8.x work begins in earnest.)

Have I missed anything that needs to be in this release?


martypdx:
LGTM

codler:
:+1: I like frequent small releases especially after a major release.

Rich-Harris:
This is now merged and published (http://cdn.ractivejs.org/0.7.2/ractive.js, http://cdn.ractivejs.org/latest/ractive.js etc, plus [npm](https://www.npmjs.com/package/ractive)).

guilhermeaiolfi:
ractivejs.org needs desperately a blog post. The last one is from may/2014 (almost an year ago). People trying ractive can see that as a bad sign. 0.7 was a major release and i've seem not even a tweet about it :(

Rich-Harris:
@guilhermeaiolfi you're totally right. I was trying to bring all the websites up-to-date before writing an announcement (a lot of stuff is/was as out of date as blog.ractivejs.org, e.g. the main website was running on Ractive 0.4.0...) but haven't managed to get that done yet. It is definitely on the todo list. Keep prodding us if it doesn't happen soon!

Issue #1897, opened 2015-04-02
XSS


Rich-Harris:
Via the [mailing list](https://groups.google.com/forum/#!topic/ractive-js/Uf8Y0VEk2FI): the [mustache-security](https://code.google.com/p/mustache-security/) project does research into potential XSS vulnerabilities in various front end frameworks. [Here's the report for Ractive.](https://code.google.com/p/mustache-security/wiki/RactiveJS).

They use the following qualifiers:
- **{}SEC-A** Are template expressions executed without using eval or Function? (yes = pass)
- **{}SEC-B** Is the the execution scope well isolated or sand-boxed? (yes = pass)
- **{}SEC-C** Can only script elements serve as template containers? (yes = pass)
- **{}SEC-D** Does the framework allow, encourage or even enforce separation of code and content? (yes = pass)
- **{}SEC-E** Does the framework maintainer have a security response program? (yes = pass)
- **{}SEC-F** Does the Framework allow or encourage safe CSP rules to be used (yes = pass)

I'll address each in turn.
### {}SEC-A

Currently, expressions are parsed into a compact AST, but then 'flattened' to a string representation (e.g. `{{a+b}}` and `{{c  +  d}}` both get represented as `_0+_1`) which is turned into a function with `new Function(...)`. This is both fewer bytes to send down the wire (if templates are precompiled), and is most likely more performant than the alternative.

But the alternative (preserving the AST, and evaluating the expression without code generation) does exist. It might be worth doing some research into the performance implications (and seeing how much code it would take).
### {}SEC-B

Functions run in the global context, but only a [subset](https://github.com/ractivejs/ractive/blob/7b8d9bf7cff9201dae64a00817bbf8e5039296a9/src/parse/converters/expressions/primary/readReference.js#L9) of built-ins can be used within an expression. An attacker could do `window.Math = somethingNaughty` and we would have no protection. We could mitigate that _slightly_ by storing references to those globals when Ractive first executes, and somehow using those instead of the (potentially compromised) globals, but that's no protection against them being compromised _before_ Ractive loads.

Maybe we could have a `ractive-secure.js` version for the ultra-paranoid that included our own implementations of those globals, and banned the ones we can't implement ourselves.
### {}SEC-C

We get a fail for that in the report, though more recent versions of Ractive _do_ only allow script elements to contain templates.
### {}SEC-D

[...]

PR #1900, opened 2015-04-02
add notes on reporting security vulns

Rich-Harris:
We got a black mark on a security report (see #1897) for not having a security response program. As a starting point, we should make it clear that people are encouraged to report security vulnerabilities, and by asking people to add `[security]` to issue titles we can ensure they're appropriately prioritised.

One thing this *doesn't* give us is a way for people to report vulnerabilities privately. Is that something we also need to consider?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,6 +15,11 @@ If you're having trouble getting Ractive to do what you want, there are a couple
 
 Of course, if you've encountered a bug, then the best course of action is to raise an issue (if no-one else has!).
 
+Reporting security vulnerabilities
+----------------------------------
+
+If you think you've found a security vulnerability, please raise an issue, prefixing the issue title with `[security]`.
+
 
 Raising issues
 --------------
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,7 +18,7 @@ Of course, if you've encountered a bug, then the best course of action is to rai
 Reporting security vulnerabilities
 ----------------------------------
 
-If you think you've found a security vulnerability, please raise an issue, prefixing the issue title with `[security]`.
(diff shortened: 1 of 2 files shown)


MartinKolarik:
I think vulnerabilities (although I don't expect there to be many) should be always reported privately.

Rich-Harris:
@MartinKolarik yeah, I guess you're right.

Any opinions on how we make that work? Does one person volunteer to be the point-of-contact for these reports, with their email address in CONTRIBUTING.md, or do we set up an email address for the core team? Bearing in mind that reports are likely to be infrequent.

Madgvox:
If they're likely to be infrequent, I don't think it'd be too big a hassle to have one person take the initial message and then loop everyone else into the thread manually. That said, if creating a mailing list would be easy to set up, that'd be the better option for sure.

MartinKolarik:
Having only one person as the point-of-contact could be a problem if they become temporarily unavailable for any reason. I can set up an email address with forwarding to the members of the core team.

Rich-Harris:
@MartinKolarik that'd be great, thank you. You're right on all counts

MartinKolarik:
Turns out Google groups can also be private, so I set up one instead of a regular email account. @ractivejs/team-ractive members, please request an invitation [here](https://groups.google.com/forum/#!forum/ractive-js-security).

Only members can see the messages sent to the group. Note that when somebody sends a report, you need to cc them when responding.

Rich-Harris:
@MartinKolarik thanks :email: :closed_lock_with_key: :+1:

Rich-Harris:
Have amended [README.md](https://github.com/ractivejs/ractive/blob/security/README.md#reporting-security-vulnerabilities) and [CONTRIBUTING.md](https://github.com/ractivejs/ractive/blob/security/CONTRIBUTING.md#reporting-security-vulnerabilities) with the email address - have I missed anything?

MartinKolarik:
@Rich-Harris looks good.

Issue #1927, opened 2015-04-20
TypeError: Cannot call method 'add' of null (global_css)


fabien:
I just updated from 0.61 and there appears to be a change when trying to build a component in Node. It seems to be related to https://github.com/ractivejs/ractive/blob/7741ceb6e42f2033b98e9ba678d0efa44c6da208/src/global/css.js#L13 and the fact that global_css will not be initialized correctly in this situation.

Removing any `<style>` tags sidesteps the issue, but I would still like to be able to share the exact same component code on both server and clientside.

Rich-Harris:
Thanks - the fix is in, it'll be released in 0.7.3. When I say 'fixed', I mean that it will no longer throw an error - there's a big ol' TODO around how encapsulated CSS is actually handled with server-rendered HTML, which will probably be one of the major areas of focus for 0.9. For now, 'adding' and 'applying' encapsulated CSS is a noop in node.

PR #1931, opened 2015-04-20
0.7.3

Rich-Harris:
Just a few small fixes - most notably, should make `<script>` tag templates work in IE8 again.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,9 @@
 # 0.7.3
 
 * Fixed reading templates from `<script>` tags in IE8 (#1908)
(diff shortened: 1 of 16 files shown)


Rich-Harris:
Before merging this I'd like to fix #1927 as well since it's an annoying regression

Rich-Harris:
So #1927 and #1933 are also fixed in 0.7.3 - anything else that's urgent, or do we reckon we're good to go?

evs-chris:
Not that I've been 'round much, but for works me. And there's always 0.7.4 :grinning:

Issue #1933, opened 2015-04-20
contenteditable two-way binding fails with lazy: true


Rich-Harris:
via [twitter](https://twitter.com/rreddi/status/590274831813337088): http://jsfiddle.net/rich_harris/6b9syyvf/

PR #1935, opened 2015-04-21
Fix contenteditable binding laziness

evs-chris:
This bases `ContentEditableBinding` on `GenericBinding` so that it also gets working lazy and twoway directive handling. I'm not sure why it wasn't based on it to begin with... anyone else know why?

### DIFF
--- a/test/__tests/twoway.js
+++ b/test/__tests/twoway.js
@@ -749,3 +749,17 @@ if ( hasUsableConsole ) {
 		console.warn = warn;
 	});
 }
+
+test( 'Contenteditable works with lazy: true (#1933)', t => {
+	const ractive = new Ractive({
+		el: fixture,
+		template: '<div contenteditable="true" value="{{value}}"></div>',
+		lazy: true
+	});
+
+	const div = ractive.find( 'div' );
+	div.innerHTML = 'foo';
+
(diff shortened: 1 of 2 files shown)


evs-chris:
Oh yeah, fixes #1933

Rich-Harris:
-23 lines, and fixes a bug in time for 0.7.3 - aa69764fd1669eec56001ffca44db116a6ddfc61 is my kind of commit! I added a test to guard against regressions

evs-chris:
> I added a test to guard against regressions

Gah! Forgot a test.

testing

test/__tests/twoway.js
```diff
@@ -749,3 +749,21 @@ if ( hasUsableConsole ) {
...
+test( 'Contenteditable works with lazy: true (#1933)', t => {
+	const ractive = new Ractive({
+		el: fixture,
+		template: '<div contenteditable="true" value="{{value}}"></div>',
+		lazy: true
+	});
+
+	const div = ractive.find( 'div' );
+	div.innerHTML = 'foo';
+
```
martypdx: sorry late to the party, but shouldn't we test `t.ok( !ractive.get( 'value' ) )` here, otherwise test would work lazy true or false.

PR #1938, opened 2015-04-21
Fixes #1937 - "radio" inputs conflicts

### LINKED ISSUE #1937: "radio" inputs causes conflicts between components
The method described in http://docs.ractivejs.org/latest/two-way-binding#radios, causes conflicts between components, if they use the same name, as you can see in this jsfiddle:

http://jsfiddle.net/noy7o6u1/

blackgate:


Madgvox:
As nice as it would be to namespace the radio buttons and be done with it, this would lock down the possibility of someone wanting to share radio buttons across multiple ractives (i.e. the exact opposite of what you're trying to do). Unfortunately, we're going to have to find another solution to the problem.

blackgate:
You're right. Makes sense if you want to create a custom radio component for example.

Issue #1982, opened 2015-06-02
event.index is undefined on custom proxy events


bensinober:
When using ractive-events-keys.js and probably any other custom proxy events, the event.index object is left undefined.

I've left a small jsfiddle to show the issue:
http://jsfiddle.net/u1fh11fd/

ISNIT0:
@bensinober Do you mean like this? 
http://jsfiddle.net/u1fh11fd/1/

bensinober:
Actually, that's a workaround, but will not work e.g. in nested templates.
I believe event.index should be an object containing all active indexes, at least it works this way for on-click.

I've demonstrated this with a nested fiddle here: http://jsfiddle.net/h25aytwk/
using on-click proxy event, you can see that event.index object holds both outer and inner index, whereas keypath only gives current path.

Using on-enter will leave event.index undefined, which is the bug I've submitted a pull request for.
Benjamin

heavyk:
this is fixed in edge.

Madgvox:
:+1:

PR #1983, opened 2015-06-02
Add index refs to getCustomHandler

bensinober:
Resolves issue #1982 where event.index was missing/undefined on custom events, e.g. from keypress enter in ractive-event-keys.js

### DIFF
--- a/src/virtualdom/items/Element/EventHandler/prototype/listen.js
+++ b/src/virtualdom/items/Element/EventHandler/prototype/listen.js
@@ -42,13 +42,22 @@ export default function EventHandler$listen () {
 function getCustomHandler ( name ) {
 	if ( !customHandlers[ name ] ) {
 		customHandlers[ name ] = function ( event ) {
-			var storage = event.node._ractive;
+		  	var storage,
+  			    handler,
+  			    indices,
+  	  		  index = {};
+  			storage = event.node._ractive;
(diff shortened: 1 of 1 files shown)


martypdx:
@bensinober Thanks for the PR. 

It looks like we now have pretty much same code as [here](https://github.com/ractivejs/ractive/blob/dev/src/virtualdom/items/Element/EventHandler/shared/genericHandler.js) and should put that into `getEventObject` module that can go into [EventHandler/shared folder](https://github.com/ractivejs/ractive/tree/dev/src/virtualdom/items/Element/EventHandler/shared).

Also need to add test for this similar to [this one](https://github.com/ractivejs/ractive/blob/dev/test/__tests/events.js#L128-L148).

bensinober:
@martypdx, thanks for quick followup!
Can I leave it for the experts? Don't want to sound lazy, but I've probably not the best toolbox around.
But of course, I'll give my best, if need be.

Madgvox:
Of course! Any contribution is always welcome, but please don't feel obligated to do anything.

sisou:
I believe the missing index object on custom events is also the reason why the step 10 of the tutorial (todo app) does not work. Any way to fix this?

evs-chris:
This is fixed in 0.8 :smile:

Issue #1856, opened 2015-03-21
Change {{#with obj}} to not render if obj is falsey


martypdx:
@Rich-Harris brought this up in a comment [here](https://github.com/ractivejs/ractive/issues/1776#issuecomment-77954227) and I agree that `{{#with obj}}` should not render if `obj` is falsey.

Madgvox:
Shouldn't it also have an `{{else}}

martypdx:
It does already, added it last release. :) I

t's just kind of weird in that else works when falsey, but if you don't have an else block it runs the with with the empty object. Another reason to fix it up.

Madgvox:
Oh, sweet! I totally missed that addition. Time to go edit some templates...

ISNIT0:
https://github.com/ractivejs/ractive/pull/1984

Madgvox:
This behavior still occurs in 0.8-edge: http://jsfiddle.net/jwk54ah9/

skeptic35:
So... am I reading too much into this or does it mean the answer to #2024 should be "an empty object ist falsey"?

MartinKolarik:
@skeptic35 yes, for explicit `{{#with object}}` it's now falsey. For `{{#object}}`, it's still truthy, but I think we should change that as well.

dagnelies:
Damn, I'm always sooo late. I know it's long past, and I don't expect it to change, just wanted to add my grain of salt.

IMHO making `#with` work similar to `#if` was a step backwards. 

1) It broke all templates when upgrading from 07 to 0.8
2) It behaves now differently than handlebars, which is counter-intuitive to many
3) Before, a "conditional "with" could easily be achieved by combining "#if" and "#with". Now, chunks like:

```
{{#with just_let_the_user_fill_this_object}}
<div>
  Foo: <input value="{{.foo}}" />
  Bar: <input value="{{.bar}}" />
  More: <input value="{{.more}}" />
</div>
{{/with}}
```

Must always now be written more verbosely:

```
<div>
  Foo: <input value="{{just_let_the_user_fill_this_object.foo}}" />
  Bar: <input value="{{just_let_the_user_fill_this_object.bar}}" />
  More: <input value="{{just_let_the_user_fill_this_object.more}}" />
</div>
```

evs-chris:
[...]

PR #1984, opened 2015-06-02
Re: Bug 1856 {{#with obj}} should not render if obj is falsey

ISNIT0:
My first Ractive pull request, there may well be a better way of doing this.

https://github.com/ractivejs/ractive/issues/1856

### DIFF
--- a/src/virtualdom/Fragment/prototype/render.js
+++ b/src/virtualdom/Fragment/prototype/render.js
@@ -1,14 +1,14 @@
 export default function Fragment$render () {
 	var result;
-
+	
 	if ( this.items.length === 1 ) {
 		result = this.items[0].render();
 	} else {
 		result = document.createDocumentFragment();
-
-		this.items.forEach( item => {
-			result.appendChild( item.render() );
-		});
+		if(!this.parent||this.owner.value)
+			this.items.forEach( item => {
+				result.appendChild( item.render() );
+			});
 	}
 
 	this.rendered = true;


ISNIT0:
Pretty sure the Travis Build failing was nothing to do with me.

Madgvox:
Unfortunately, that change causes many tests to fail, so it would be unsafe to merge this PR. However, I think this is more indicative of the importance of one of our long-term goals of making the ractive codebase easier to read and modify to users outside of the project. Thank you very much for your contribution regardless!

PS: In the future, you can make sure your changes don't break anything by running `npm start` in the command line, then navigating to `localhost:4567/test` in your browser as outlined in [this section](https://github.com/ractivejs/ractive#developing-and-building) of the readme. That will run the entire suite of unit tests on the library to ensure that everything still works like it's supposed to.

ISNIT0:
Gotcha, I'll check them all out :)

martypdx:
@ISNIT0 Thanks for the PR. The right place to do this is in the [Section$setValue}(https://github.com/ractivejs/ractive/blob/dev/src/virtualdom/items/Section/prototype/setValue.js). Need to add tests to `render` tests, and change the breaking tests that assume old behavior.

All that being said, I already changed this in #1740, so this should be the behavior for `0.8`.

ISNIT0:
@martypdx Thanks :) if there are any pull requests you reckon are 'beginner level' then please do send them my way. I am just starting to use Ractive at my work and really would like to get into the community and learn the platform!

Issue #1986, opened 2015-06-05
Functions shouldn't get re-evaluated on outro


markisus:
Fiddle: http://jsfiddle.net/rdgs7rv2/3/

I have a list with some strings inside of it. I want Ractive to turn each string into an image through a function called get_url.

The relevant part of the template looks like 
```
 {{#game.cards}}
        <img intro-outro="slideh" src="{{get_url(.)}}" width="100em" height="100em">
    {{/cards}}
````

When the list is emptied, the images should slide out. But somehow the get_url function is getting called again with an empty argument after the list is emptied and the images are replaced with a default image (in this case a bee instead of a chair).

I'm pretty sure this is not the desired behavior, is it?

zenflow:
I also noticed this behavior and found it undesirable.

For the record, this behavior occurs with nodes being removed, even without the outro directive. In the above fiddle, try removing the intro-outro attribute and adding a call to `console.log` or `alert` inside the `get_url` function. You will see that `get_url` is called 4 times (once for each image) immediately before the images are removed.

For my own purposes I used a work around, which is sufficient, but I do think changing it to only compute/render within *existing* nodes (including nodes being "intro"d, but not nodes being "outro"d or simply removed) would be an improvement, although I currently have no idea how hard this would be to implement.

Agree? Disagree?

Madgvox:
If I'm understanding the changes from #1740 correctly, this issue should become fixed with 0.8. The new logic is a lot more choosy about when to attempt to evaluate a keypath, which helps improve performance, but should also prevent `get_url` from being evaluated again unnecessarily.

heavyk:
this is fixed in edge.

updated the fiddle.
http://jsfiddle.net/9sy4upj8/

Madgvox:
Fantastic! I will close the issue.

Issue #1776, opened 2015-03-06
Revisit forced resolution on two-way bindings


martypdx:
Currently two-way bindings force resolution of the keypath on which they are bound, and current binding rules specify that an ambiguous reference resolves to the instance root. It can lead to situations like this:

``` js
{{#with foo}}<input value='{{bar}}'>{{/with}}
```

In the above, if no `foo` in `data`, the input is bound to `data.bar`, **not** `data.foo.bar`.

So @Rich-Harris added #1765, which will helpfully let you know that you should probably initialize data.

The problem I'm seeing is that with asynchronous data retrievals, particularly when used with an adaptor, it's not a viable option to sketch out an empty schema. Which means you either have to only use restricted references:

``` js
{{#with foo}}<input value='{{.bar}}'>{{/with}}
```

or use an additional `if` block:

``` js
{{#if foo}}{{#with foo}}<input value='{{bar}}'>{{/with}}{{/if}}
```

Neither of which seem developer friendly, IMO. And the `if` block approach as the added disadvantage of not rendering the DOM while waiting for the data to load anyway!

Two possible options I can think of:
1. Delay the timing of forced resolution until the bound control is actually used
2. Redefine forced resolutions rules so that closest context, not root, is used.

Not a `0.7` thing, but I think we're only going to see more async data use as things roll forward.

evs-chris:
Option 2 sounds like the way to go to me, given that there is no `data.bar` in existence when the binding is created.

Option 1 sounds like the best possible scenario, but I think it might be troublesome with controls that don't have a clear not-yet-used state and instances where you *want* to let the control initialize the data.

Rich-Harris:
I like option 2 as well. Not initialising the values of `<select>` elements, or a group of checkboxes where `checked` is initially true in some cases, could be problematic. The combination of defaulting to local context and printing warning messages is probably enough.

Re `{{#if foo}}{{#with foo}}`... I've occasionally wondered if `{{#with}}` shouldn't be both `with` and `if`. That's actually how it is in Handlebars, and it's what `with` becomes if it has an associated `else`/`elseif` block. (In other words, it's an `if` that also creates a new context.)

arxpoetica:
:+1: on the `{{#with}}` equals `{{#if}}` part.

[...]

PR #2022, opened 2015-07-02
The road to Ractive 0.8 is tree-lined

Rich-Harris:
Brace yourselves: a tsunami of text approaches. I've been squirreling away on this branch for a while now without sharing the progress, which, although necessary – due to a) the magnitude of the changes and b) uncertainty over whether they'd pan out – isn't how I like to work. This PR (which isn't ready for merge) is intended to bring the changes into the open and explain how 0.8 will most likely differ from 0.7.

Those of you who follow Ractive's development will be aware of #1740, an epic pull request by @martypdx. Marty had a Eureka moment a while back relating to how data changes propagate through Ractive instances, and #1740 applied that insight to the existing codebase. Unfortunately, as I contributed to #1740, I came to realise that the technical debt in the codebase (incurred when Ractive was my solo project, I hasten to add) made such large-scale changes extremely difficult, and I began to wish we could rewrite it as though Marty had designed it in the first place.

This is that rewrite. It's based on #1740, and aims to achieve the following:

* Replace the 0.7 (and earlier) viewmodel concept with a lighter tree-like structure (as per #1740)
* Apply the same thinking to the way DOM updates are applied. More on this below
* Make the codebase lighter and more readable

The first of these is all about performance. While Ractive comfortably bests most of its competitors in typical scenarios when it comes to DOM update performance, templates with hundreds of components are sluggish on initial render, which is often when performance is most critical. This is largely due to some convoluted caching logic that becomes unnecessary when you model things the right way.

The second is also performance related, but is also a big win for simplicity. As with the data side of things, thinking about DOM updates in terms of trees makes a lot of the indirection in the current codebase unnecessary. Tricky issues like #1986 should be much easier to resolve.

The third bullet point is perhaps the most important. Time and again, people tell us that they'd love to contribute, but that the codebase is too overwhelming to get a handle on. That's a great shame, and left unchecked will slow Ractive's development down. This PR won't fix that in one go, but I'm confident that it's a major step in the right direction.

### Fewer modules

The current codebase is characterised by lots of tiny modules. In many cases that's unnecessary, because those modules are only used in one place (e.g. they define a method on an internal class). In this PR, you're more likely to see one file per class than one file per method. This is partly an aesthetic choice (I think it's easier to read), but largely motivated by the use of ES6 classes, which have a more streamlined syntax that encourages this form. (The `class` keyword is much maligned, but is actually very well suited to Ractive's ontology, at least internally.)

Note: our build process will need to be updated to accommodate classes – at present, Babel's `classCallCheck` helper is included many times when it only needs to be there once.

### viewmodel -> model, virtualdom -> view

Strictly speaking it doesn't make sense to say that these folders have been renamed, since their contents are all new. But the responsibilities of the things formerly found in `src/viewmodel` now reside in `src/model`, and likewise with `src/virtualdom` and `src/view`.

Partly this is because even though Ractive isn't an MVC library in the traditional sense, those terms are well-understood and the division of responsibilities is easy to grok. But mostly it's because I like to be able to use bash autocomplete after typing the first couple of letters of a directory name, and 'v, i, TAB, nope, e, whoops, BACKSPACE, r, TAB' was driving me mad.

### Trees

As alluded to, 0.8 is all about trees. Each value in your `data` object is represented by a `Model`. If you have this...

```js
var ractive = new Ractive({
  data: {
    foo: {
      bar: 42
    },
    baz: 'childless'
  }
});
```

...then the root node has two children with the keys `foo` and `baz`. `foo` is a branch (with a child, `bar`), `baz` is a leaf. Each node is aware of its immediate parent-child relationships, and of things that depend on its value. Models are created 'lazily' (i.e. as and when they become necessary for Ractive to track).

An instance's 'root model' (the equivalent of `ractive.viewmodel` in 0.7 – will most like become `ractive.model`, I suppose) is just another node, with a few minor differences; ditto computed values.

As soon as a model changes value (internally, `model.set(newValue)`), each of its dependencies are notified of the change (via their `handleChange` methods). There's no longer any need to distinguish between different types of dependant (e.g. computation, observer, virtual DOM node). When a virtual DOM node (hereafter, 'view') is notified in this way, it marks itself as 'dirty' and 'bubbles' the notification. Parent nodes (fragments containing the view, and elements/sections etc containing those fragments) are also marked as dirty and continue propagating the notification (unless they were already dirty).

The end result of this is that a top-level fragment is marked as dirty. Once all the data changes have been propagated, and observers etc have been called, we can 'clean' it. Because we're thinking in trees, we can disregard already-clean branches (the bits that don't need to change), and ignore any children of things that no longer need to exist.

### (Almost) no breaking changes

I'm hopeful that the only breaking changes will be desirable ones (e.g. #1776). If there *are* any breaking changes as a result of this rewrite, they'll be signposted in the documentation, but so far it looks good. (About 5% of the tests are currently failing, but most of them look like they'll be remedied reasonably straightforwardly.)

------

If you read this far, congratulations on your stamina. As ever, feedback and questions are warmly welcomed. Thanks!


Skyrpex:
Those are great news!!

I'm actually in the process of choosing a MVC-esque JS framework to integrate in my company. I really like Ractive and would really love it to become more mature, free of bugs and easier to understand.

I hope to dive in the Ractive sources and contribute a little bit the future.

May the force be with us!

rstacruz:
:+1: just chiming in to say @Rich-Harris and @martypdx you guys are doing awesome work :)

Madgvox:
:sunglasses: Awesome! I've got to checkout this branch and take a gander.

smallhadroncollider:
Sounds good to me. Good work guys!

guilhermeaiolfi:
NICE! That's a LOT of work. I can't wait to see some benchmarks after all those commits.

martypdx:
Hey @Rich-Harris - kudos to you for the picking up the ball and scoring a wicket, um a try? a goal? Well whatever :)

I did a quick once over, but not a pull and try it out. Will do when I get a chance.

One thing I wanted to check, because we ran into this today with `0.7`, was whether `splice` was still doing DOM rearranging versus reusing the nodes. We just had to use `merge` to get that behavior to work around some issues. Like I said, I'll check it out myself with this branch but thought I'd throw it out there.

Also, if I notice parts of #1740 that didn't get pulled over but should - do you want me to wait or push on this branch?

One thing I did notice is that we should make sure we aren't adding (or removing - but that's doesn't really happen in ractive classes) properties to classes outside of the constructor. It's an optimizer thing.

Rich-Harris:
@martypdx Oh, you must be referring to croquet - we call it a 'hoop'.

Splicing and merging both use the same mechanism ('shuffling'). It is rearranging the DOM. Definitely a major source of pain as it inevitably blurs at least one conceptual boundary (and I flip-flopped on the implementation a couple of times), but it seems to be solid now. 

> Also, if I notice parts of #1740 that didn't get pulled over but should

It should have everything - I branched after the most recent commit on that branch. But if there is anything missing then please, dive in. 

> It's an optimizer thing

Yeah - agree it's worth doing, there are definitely some places where I forgot to. Basically impossible to enforce through static analysis so I guess we just have to be vigilant.

Speaking of optimisation, another gotcha - block scoping in for/while loops. I forgot what a headache this caused before, so there's a few places with IIFEs inside loops, probably in performance sensitive areas. Will have to clear those out.

aharpervc:
Philosophically I'm disappointed with the move away from the `viewmodel` jargon, as I feel like that is the correct word and most precise description of responsibilities. I don't really consider a Ractive an "MVC" library so I don't see why it should have to move to that domain's jargon.

Anyway, looking forward to migrating to 0.8 (still on 0.4 :dancer: )

parhelium:
@Rich-Harris @martypdx  Great work ! Thank you !

martypdx:
> Philosophically I'm disappointed with the move away from the viewmodel jargon, as I feel like that is the correct word and most precise description of responsibilities. I don't really consider a Ractive an "MVC" library so I don't see why it should have to move to that domain's jargon.

@aharpervc - As someone who's spent quite a bit of time in the Ractive code, I have to second @Rich-Harris observation that the similarity between `viewmodel` and `virtualdom` in the project is a cognitive PITA.

But developer conveniences aside, I still like the name changes. I think component architectures will replace MVC as the predominant architectural paradigm for browser apps. CA are networks of components (trees!) that deal with data from one or more stores, either independently or inherited from their parents. Typically this means you can access the store directly without the need for some "other" model (though nothing's stopping you from using ampersand or backbone or whatever if that's what you like). 

So conversely I think the term `viewmodel` should be abandoned precisely because it implies you should have some other `model` representation on top of it. From Ractive's perspective the `model` classes are exactly what they imply: descriptions of the underlying source data for the purpose of binding to view items.

I like `view` for two reasons: 
  1. IMO `virtualdom` has become associated with `dom-diffing` which I find inferior to declarative templates. `0.8` is the start of some changes to continue to exploit that view update operations in Ractive are O(1) versus O(n) or higher in dom-diffing.
  2. I'd like to see `view` _implementations_ in Ractive become more specific. Meaning that you'd have a DOM renderer, a pure text-render (what you'd use on the server or for `.toHTML()`), maybe a Native renderer for mobile, etc.

aharpervc:
> Typically this means you can access the store directly without the need for some "other" model (though nothing's stopping you from using ampersand or backbone or whatever if that's what you like).

>  From Ractive's perspective the **model classes** are exactly what they imply: descriptions of the underlying source data *for the purpose of binding to view items.*

First, gross, I'm never writing any backbone code again in my life if I can swing it.

Second, you probably still do have models in your app, just not models which are javascript "classes" (excuse the fuzziness of that word in this stack). They're probably your DTOs (if youre talking to a server), or otherwise coming from your app's data access layer, which get fed into your ractive viewmodels for binding & state management. Some of those viewmodels may have UI capabilities, some not. But the capability of directly referencing & managing a related UI state is what makes a viewmodel a viewmodel. For whatever reason yall are seeing connections to views as justification for calling the ractive data things `models` which is bizarre.

Rich-Harris:
@aharpervc you're reading way too much into this! `Model` is just a convenient name - easy to type, and has broadly the right connotations (i.e. it's a class for making values observable). It doesn't imply any philosophical shift, and it won't have any impact on the public API - it's purely an internal implementation detail that allows us to worry about fewer things at a time and propagate changes through the system more efficiently.

We leave arguments about the precise semantic boundaries of MV* jargon to people who care; we're too busy making tools that are actually useful :-)

Madgvox:
> We leave arguments about the precise semantic boundaries of MV* jargon to people who care; we're too busy making tools that are actually useful

#sickburn

aharpervc:
>  It doesn't imply any philosophical shift

Good, this is what I was concerned about. Occasionally in the English language people use words to describe their thoughts, so when words change the implication is that the thinking has changed. 

> We leave arguments about the precise semantic boundaries of MV* jargon to people who care; we're too busy making tools that are actually useful :-)

Nice burn, but being too casual about naming, and/or naming things contrary to common patterns only increases the friction for people to use Ractive and get involved in the codebase. If you name things "model" and "view", people who haven't been following the project for very long could make presumptions that are incorrect. Generally speaking the goal should be to pick names which reflect the thinking, which previously to now seemed to be the case.

Rich-Harris:
> Generally speaking the goal should be to pick names which reflect the thinking

I agree, but I'm not sure if there's a commonly accepted term for what `Model` (as it's currently named) does - `Observable` is probably closest, though that's quite strongly associated with [Rx](http://reactive-extensions.github.io/RxJS/) and implies an interface that `Model` doesn't satisfy. I don't think reverting to `Viewmodel` makes sense as a) the ergonomics suck, as mentioned, b) it probably makes too strong an association to MVVM, and c) the new `Model` (responsible for an individual node in the data tree) has a very different set of responsibilities to the old `Viewmodel` (responsible for the whole tree, plus computed values and inter-component mappings). Definitely open to suggestions though.

MartinKolarik:
1. Great work guys! Can't wait to check this out.
2. :+1: for `model` and `view`

evs-chris:
I finally got round to looking at this, and I must say that it is a phenomenal set of changes! I can actually almost fit everything into my tiny set of working memory! Between the implosion of files and conceptual simplification, this should lower barrier to entry significantly. Awesome work, @Rich-Harris and @martypdx!

FWIW, I really like the model/view change. Patterns, acronyms, and whatnot aside, it seems to fit my concept of which parts do what in Ractive nicely.

barjo:
Neat!

FWIW +1 for the model/view change! 
I suggest `RactiveAwesomeDynamicModel` .. or `SmartModel` or `IModel` or `JanusModel` or `AutonomicModel` :laughing: just in case you ever have to write a marketing/research paper.

> Whatever we conceive well we express clearly, and words flow with ease.
(Nicolas Boileau)

JonDum:
How goes this? Still tree-lined? :evergreen_tree:

Rich-Harris:
@JonDum There's a few tests that still don't pass – edge cases around computed properties and the like.   My plan is to merge this into the dev branch (so that people can start trying it out as the edge build) as soon as they're in place – unfortunately my schedule the last few weeks hasn't been particularly supportive of those ambitions but I'm anxious to get back on it! Sorry for the long delay. It *is* happening.

JonDum:
@Rich-Harris No need to apologize, man—we all have work and life responsibilities. Looking forward to that merge so I can test it out and contribute what I can.

evs-chris:
@Rich-Harris @martypdx I'm taking a swing at a few of the failing tests to familiarize myself the epic changes here. I resolved the out-of-order failure by letting computations contribute to capture and notify deps on init. That seems not to jive with the lazy computation test, which was already failing, because it triggers computation updates as new dependees arrive. My question is, are computations meant to remain entirely lazy? It would seem the computation models do not want to be as lazy as the old methods, and to get them to be so would require moving value access to a getter like `.value()` instead of `.value`. How far in left field am I wandering?

Also, what should I do with fixes? I have created an evschris-arboretum branch for now, though the changes there are pretty tiny.

evs-chris:
D'oh... looks like @Rich-Harris is waaay ahead of me (as usual) on the computation getter issue. I just noticed the arboretum-getter branch, which probably solves other problems that I've yet to stumble across as well.

Rich-Harris:
@evs-chris ah, yep, sorry for not signposting that branch a bit better! was one of those 'this might break everything so I should do it off to one side' things. Basically, yeah, I think it's better if computations *are* always lazy (otherwise you might get a console full of warnings when you try to retrieve the undefinedth member of undefined...), I just wasn't smart enough to plan it that way at the start of this refactor. It's *probably* better to push changes to `arboretum-getters`, then soon we can merge that back into `arboretum`.

Rich-Harris:
Tests passing... no merge conflicts... YOLO

bryansray:
Awesome. Can't wait to give it a go! Thanks, Rich. #yolo

martypdx:
http://25.media.tumblr.com/tumblr_lrech7SQPj1qiz3j8o1_500.gif

NikhilVerma:
This is amazing! I've been waiting for this to be merged for a long time :) Thank you !

JonDum:
its happening!

JonDum:
Tried Dev in one of my ractive apps...

Hundreds and hundreds of `Failed to compute`'s, and a yummy `An error happened during rendering` before poor ol' ractive finally gave up haha :D 

@Rich-Harris Should we start new issues for this stuff? Or do you want to hold off until it stabilizes more?

fskreuz:
@JonDum It maybe handy to mention what version of Ractive you came from.

Based on your observations, I assume your app previously used a version prior to 0.7. That's because starting with 0.7, Ractive has `debug` globally set to `true`. One indication is that computed properties will show a lot of logging, which is normally hidden in prior versions.

The `An error happened during rendering` would be something though. Never seen those before.

Rich-Harris:
I'd say err on the side of reporting issues – if nothing else, it'll help other people find the answers if it does turn out to be a migration headache as opposed to a bug proper. And if it *is* a bug, we definitely need to know!

The good news is it ought to be easier to address these issues now that there's no longer a massive refactor looming over the `dev` branch.

JonDum:
@fskreuz I was on `edge`, debug was already enabled so they were all errors being thrown in 0.8 that were not happening in `edge`.

@rich-harris will do. Is there any sort of migration guide or should there be no breaking changes in the api?

fskreuz:
@JonDum Not sure I understand. [edge *is* 0.8](https://github.com/ractivejs/ractive/commit/38a5fee1bec0d7b74312a357a5fa263b69f2efa0).

fskreuz:
>  they were all errors being thrown in 0.8 that were not happening in edge

@JonDum Not sure I understand. [`edge` is 0.8](https://github.com/ractivejs/ractive/commit/38a5fee1bec0d7b74312a357a5fa263b69f2efa0)

JonDum:
@fskreuz Yes, *now* they are. You understood what I meant.

Rich-Harris:
@JonDum We've tried to minimise breaking changes – to be honest I'll need to scour the commit history to figure out which ones exist. It's all fairly random stuff like `foo` in `{{#each foo}}` can't change between an array and an object, which is a mythical use case that caused code bloat. There'll be a migration guide [here](http://docs.ractivejs.org/latest/migrating) when 0.8 is stable, until then shout if you find breaking changes so we don't forget about them!

Issue #2019, opened 2015-06-29
Expression references that utilize data functions are not called with ractive instance as `this` when used as arguments of event or expression methods.


JonDum:
Ran into more fun stuff this evening. If you call a function within another function, the inner function will get called under global scope instead of the instance's scope. Happens in events and in regular expressions:

Example:
```
**stuff**
 <li on-click='set("data."+virtualToReal(r)+"."+columns[c]+".type", this)'
**more stuff**
```
What I love about ractive is how crazy you can get in templates, which is super powerful compared to Meteor, Ember, etc. This is actually super limiting and would require you to numerous helper functions to get anything done (one my biggest annoyances with other frameworks) instead of just a couple `sets()` here and there.

Here's an isolated running example: http://plnkr.co/edit/QH2AlBk1PmEKiBPilcyj?p=preview

JonDum:
Just read the title I wrote lol.

Obligatory http://inception.davepedu.com/inception.mp3

martypdx:
@JonDum took me a minute to follow along - but if I understand correctly, problem could be defined as:

expression references that utilize data functions are not called with ractive instance as `this` when used inside event methods as parameters

JonDum:
@martypdx Correct! That is more aptly said in ractive terms than I could muster.

dagnelies:
1) should inner functions be allowed in "method-calls"? Or even arbitrary javascript? Or just a warning requesting simple arguments only?

2) where should these hook on:
- root like for the method itself?
- data like for expressions?

fskreuz:
Dropping in 2 cents:

In plain HTML, inlined JS runs in the context of `window`. In Ractive's case, it's the instance. It might be more intuitive if everything called inline was called in the context of the instance. Besides, aren't components supposed to be isolated from the world?

As for expressions, I believe they're created using `new Function`. And [MDN says](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function);

> Note: Functions created with the Function constructor do not create closures to their creation contexts; they always are created in the global scope. When running them, they will only be able to access their own local variables and global ones, not the ones from the scope in which the Function constructor was called. This is different from using eval with code for a function expression.

Just out of curiousity, stepped through the debugger and found the following code. I assume this is the result of parsing expressions using `new Function`.

```js
// new Function('_0', 'return ["scope",_0("name")]');

(function(_0
/**/) {
return(["scope",_0("name")])
})
```

[...]

PR #2029, opened 2015-07-07
Fix event expression Function scope when called

JonDum:
A la https://github.com/ractivejs/ractive/issues/2019

Just adds a simple check to see if the item is a function and if so, binds it to the instance.

### DIFF
--- a/src/virtualdom/items/Element/EventHandler/prototype/init.js
+++ b/src/virtualdom/items/Element/EventHandler/prototype/init.js
@@ -112,7 +112,7 @@ function fireMethodCall ( event ) {
 			value = ractive.viewmodel.get( keypath );
 		}
 
-        if( typeof value == 'function' ) {
+        if( typeof value === 'function' ) {
             value = value.bind( ractive );
         }
 
--- a/src/virtualdom/items/Element/EventHandler/prototype/init.js
+++ b/src/virtualdom/items/Element/EventHandler/prototype/init.js
@@ -112,6 +112,10 @@ function fireMethodCall ( event ) {
 			value = ractive.viewmodel.get( keypath );
 		}
 
+        if( typeof value == 'function' ) {
+            value = value.bind( ractive );
+        }
+
 		return value;
 	});
 


fskreuz:
Not to be nitpicky but... `===`?

JonDum:
@fskreuz not needed with typeof. Always returns a string and when comparing with another string no conversion will occur

fskreuz:
Nearby code a few lines down uses `===` for `typeof` so... uniformity :D

martypdx:
Thanks for the PR @JonDum!

IMO `==` means "something special happening here" (like `foo == null` ), so agree with @fskreuz that if vanilla string compare, use `===`.  

Also, need to add a test in [events](https://github.com/ractivejs/ractive/blob/dev/test/__tests/events.js)

fskreuz:
IIRC, one should use tabs instead of spaces for indents. More on https://github.com/ractivejs/ractive/blob/master/CONTRIBUTING.md

JonDum:
Something must have been screwed up with .editorConfig in my vim. It looked right in Vim, but not in GH I guess. Should be tab characters now.

Rich-Harris:
@JonDum Thanks for this PR. Tabs vs spaces aside (and [GitHub really should get their act together on that...](https://github.com/isaacs/github/issues/170) probably my most commented issue ever), this fix will unfortunately be swept aside by #2022, so it's probably best we *don't* merge it, so that it stays open and we don't forget about it.

>  I'm not sure where/how to put a test case for this though.

The same applies to any tests as to the PR itself. [This'll do for now](http://jsfiddle.net/36e3wycs/) - we'll just have to ensure that it goes into the test suite before #2022 is merged. I'm tied up with a gazillion other things at the moment but if anyone wants to submit an equivalent PR against the arboretum branch it would be gladly received!

Rich-Harris:
We stumbled upon a better way of approaching this problem – binding functions on init/set (#2115) – so I'll close this PR. The test from the fiddle above is included via #2111

Issue #2027, opened 2015-07-04
Elements in `if` and `else` blocks visible at the same time


MartinKolarik:
See [this demo](http://jsfiddle.net/gc230cj9/) and click the button twice. Text nodes are removed immediately, but it seems that elements are waiting for the transition (which really has nothing to do with the if/else block) to complete.

YRathod:
Not able to reproduce with test html page but able to reproduce on jsFiddle only.

evs-chris:
Fixed in edge 😃

PR #2037, opened 2015-07-16
Bug #2027

ISNIT0:
We need to remove any non-animating elements before transitions start (for some reason they are blocking the un-rendering).
I have no idea if this is the right way to do it, what do you guys think?
#2027

Joe


evs-chris:
This is fixed in 0.8 :smile:

Issue #2038, opened 2015-07-16
Namespace issues when working with certain SVGs


ISNIT0:
At work, I need to render SVGs which have been exported from Microsoft Visio.

Exported SVGs have some wacky namespaces. I don't know if there is any way to disable the namespace checks, but I have just been editing the following on my local copy of ractive: 

Added a property here:
https://github.com/ractivejs/ractive/blob/db7a5d61730ad72857ae355597838b3d77f5bc28/src/config/environment.js#L23

v:'http://schemas.microsoft.com/visio/2003/SVGExtensions/'

This fixes the error (Uncaught Unknown namespace ("v")).
I'm not sure what the best way to avoid this is, but I would like there to be some native way to do so within Ractive.

ISNIT0:
Problem with the above pull request is that it only fixes a very specific problem when there may be other namespaces required in the future.

Madgvox:
Have you tried running your svgs through an svg optimizer like [svgo](https://www.npmjs.com/package/svgo)? They remove all the guff from exported svgs, as well as normalize the namespaces.

ISNIT0:
The problem with that is that they are user uploaded, plus I need to use
some of the properties.

Joe

Rich-Harris:
@ISNIT0 Could you provide some example code, or a JSFiddle (ideally based on http://jsfiddle.net/rich_harris/va6jU/, since that uses the latest edge version)? That 'Unknown namespace' error doesn't exist any more, and I'm left wondering if that's a good thing or an oversight! Thanks

ISNIT0:
Here we go: http://jsfiddle.net/puppr9gm/2/

We get: Uncaught Unknown namespace ("v")

Rich-Harris:
@ISNIT0 sorry for the slow response on this – please see #2304

ISNIT0:
@Rich-Harris Nice :) is there anything else to do with this issue that I can help with?

Rich-Harris:
It's merged now and will filter through to http://cdn.ractivejs.org/edge/ractive.js momentarily – if you could confirm that it behaves as expected that'd be great! Thanks

PR #2039, opened 2015-07-16
Potential fix for #2038

ISNIT0:
Not sure of the best way, but this is one possible solution to the problem at #2038


ISNIT0:
Replaced by https://github.com/ractivejs/ractive/pull/2040

PR #2040, opened 2015-07-16
Bug #2038

ISNIT0:
We now extend namespaces with a namespaces object that is supplied to the Ractive instance on init.


ISNIT0:
#2038

ISNIT0:
Odd that check are failing, all tests pass fine on my machine.

Madgvox:
From what I can see, Travis isn't getting to the tests. The linter is throwing up errors. There's a few formatting errors in your changes.

```
src/virtualdom/items/Element/Attribute/helpers/determineNameAndNamespace.js: line 7, col 5, 'utils_object__extend' is not defined.

src/virtualdom/items/Element/prototype/unrender.js: line 82, col 18, Missing semicolon.

src/virtualdom/items/Element/prototype/unrender.js: line 21, col 15, 'global_runloop' is not defined.
```

Rich-Harris:
@ISNIT0 Thanks for submitting this – see my response on #2038. The reason this isn't passing is that you're referring to function names that are generated as part of the build process, e.g. calling `utils_object__extend`. The correct thing to do is to import the function into the module and *then* call it:

```js
import { extend } from '../../utils/object';
const obj = extend(...);
```

Otherwise you're assumed to be referencing a non-existent global, and `extend` gets assigned to a different name to avoid conflicting with said global.

In any case: unfortunately it's not possible to merge this, because the current dev branch differs greatly from the dev branch at the time of the PR. Would like to get to the bottom of this though – if you could supply a fiddle (or even better, a failing test case) it seems like something we can fix.

Rich-Harris:
Closing as fixed via #2305

Issue #2043, opened 2015-07-21
Add support for Custom Element typeExtensions


paulie4:
The draft spec on Web Components Custom Elements (http://w3c.github.io/webcomponents/spec/custom/#extensions-to-document-interface-to-instantiate) describes a new HTML and DOM API enhancement that allows for a custom element to extend a native element. For example, you could extend the `<img>` element to add a caption feature to it. This new feature has worked for a while in standard Chrome/Chromium.

On the Javascript-side, they added an argument to document.createElement() to pass the typeExtension. They also added a special DOM node attribute, named "is", whose value is the typeExtension that's been registered in Javascript.

For example:
```
<img is="img-ext" src="http://lorempixel.com/400/200/nature" caption="This is my custom caption!" />
```
The typeExtension here is named "img-ext".  The JSFiddle link below shows this working (please run in Chrome/Chromium).  It also shows the same extended tag NOT working correctly when created by Ractive, since Ractive doesn't know about the special "is" attribute nor the new typeExtension argument to document.createElement().
http://jsfiddle.net/3c2pfhgL/1/

Rich-Harris:
Closing – #2044 has been ported to the current dev branch and will be included in 0.8

PR #2044, opened 2015-07-21
add support for custom element typeExtensions

paulie4:
In reference to https://github.com/ractivejs/ractive/issues/2043

This code change assumes the user will put a static string value for the "is" attribute. I don't think it makes sense to allow dynamic mustaches for "is", since the whole node would need to be deleted and a new node created with the different typeExtension. Maybe Attribute$init() could just print an error and treat a dynamic mustache as a static mustache to avoid that complication.

Note: this (https://github.com/ractivejs/ractive/blob/master/CONTRIBUTING.md#pull-requests) says to run `grunt test`, but when I try that I get "Fatal error: Unable to find Gruntfile."

### DIFF
--- a/src/utils/dom.js
+++ b/src/utils/dom.js
@@ -4,20 +4,29 @@ var createElement, matches, div, methodNames, unprefixed, prefixed, i, j, makeFu
 
 // Test for SVG support
 if ( !svg ) {
-	createElement = function ( type, ns ) {
+	createElement = function ( type, ns, extend ) {
 		if ( ns && ns !== namespaces.html ) {
 			throw 'This browser does not support namespaces other than http://www.w3.org/1999/xhtml. The most likely cause of this error is that you\'re trying to render SVG in an older browser. See http://docs.ractivejs.org/latest/svg-and-older-browsers for more information';
 		}
 
-		return document.createElement( type );
+		if ( extend )
+			return document.createElement( type, extend );
+		else
+			return document.createElement( type );
 	};
 } else {
-	createElement = function ( type, ns ) {
+	createElement = function ( type, ns, extend ) {
 		if ( !ns || ns === namespaces.html ) {
-			return document.createElement( type );
+			if ( extend )
+				return document.createElement( type, extend );
+			else
+				return document.createElement( type );
 		}
 
-		return document.createElementNS( ns, type );
+		if ( extend )
+			return document.createElementNS( ns, type, extend );
+		else
+			return document.createElementNS( ns, type );
 	};
 }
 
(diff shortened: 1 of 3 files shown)


martypdx:
Thanka! Seems reasonable to me. 

We use the `dev` branch for dev, so that CONTRIBUTING is update to date with right build/test instructions: https://github.com/ractivejs/ractive/blob/dev/CONTRIBUTING.md

Also need to add a test, probably to [elements](https://github.com/ractivejs/ractive/blob/dev/test/__tests/elements.js). They're pretty easy-to-write API level tests.

paulie4:
I finally found some time to write the test code. Sorry for my inexperience, but I'm not sure how to run tests besides via ``npm run build`` (which in turn uses PhantomJS, but I'm not sure how to run that in a browser), and since only [Chrome, Opera](http://caniuse.com/#feat=custom-elements), and browsers polyfilled via [webcomponents](https://github.com/webcomponents/webcomponentsjs) can register new elements, I can't really run this test. This is the test I wrote, but I haven't committed it, since I don't know how to really test it:
```javascript
if ( 'registerElement' in document ) {
	test( '"is" attribute is handled correctly for custom elements (#2043)', t => {
        let XFoo = document.registerElement('x-foo', {
            prototype: Object.create(HTMLParagraphElement.prototype, {
                testMember: { value: true }
            }),
            extends: 'p'
        });
		let ractive = new Ractive({
			el: fixture,
			template: '<p is="x-foo"></p>'
		});

		let p = ractive.find( 'p' );
		t.isTrue( 'testMember' in p );
	});
}
```

Madgvox:
Build the project using the command `gobble`, which will create a server at `localhost:4567`. There you can navigate to the test folder, which will run the tests in the browser.

paulie4:
OK, thanks. I've committed the new test.

paulie4:
When I run "npm run build" on my computer, PhantomJS runs and all tests pass. Does anyone know what could be causing Travis CI build's PhantomJS to fail to run? This is what Travis CI's log shows:
```
PhantomJS timed out after 3000ms
Unsafe JavaScript attempt to access frame with URL about:blank from frame with URL file://scripts/phantom-runner.js. Domains, protocols and ports must match.
```

I only modified these three files, so I'm pretty sure it's not something I caused:
```
 src/utils/dom.js                                 | 19 ++++++++++++++-----
 src/virtualdom/items/Element/prototype/render.js |  5 +++--
 test/__tests/elements.js                         | 18 ++++++++++++++++++
```

martypdx:
@paulie4 I restarted Travis and it passed. Go figure :)

Rich-Harris:
This is really cool – thanks @paulie4!

Issue #2013, opened 2015-06-25
Ability to turn off Dynamic Partials


joesepi:
Currently using slashes `/` inside a partial tag leads Ractive to try to evaluate it as an expression

It would be great to have a way to not do that, so that partials can be stated as paths.

http://jsfiddle.net/joesepi/2g4pf4m6/

Madgvox:
Do you have a real world example of where this is tripping you up? In your fiddle it works because you're passing in a string, what situation would you require not to have the quotes?

Rich-Harris:
I was actually in a meeting with @joesepi when he raised this issue :laughing: IIRC it was problematic because the templates need to be used with different Mustache engines (e.g. Hogan) for reasons that are frighteningly complicated but make total sense. In traditional Mustache engines (and in @joesepi's case), the partial name is often a filename (e.g. `{{>foo}}` loads `foo.mustache`, `{{>foo/bar}}` loads `foo/bar.mustache`), so they're having to do some crazy string manipulation hacks before feeding the template to Ractive.

joesepi:
This is correct. @Aintaer can go into the frighteningly complicated detail if need be. Thanks!

Madgvox:
Ha, I'll take your word for it!

martypdx:
Adding a note because it's tangentially related: we should change the algorithm to resolve partials as named first, then try resolving as a dynamic partial (a reference). Reason being that it's rather intensive to look up a non-existent reference and it ends up being an unresolved reference. Seems to me that named partials are the more common case and shouldn't be penalized checking if the name is a resolvable reference.

I know there's still an issue of the "name" getting parsed as an expression, but maybe at parse time we parse both "full name" and as a reference. End result being that this issue's use case would work as expected.

ISNIT0:
Does PR #2067 Mean this issue can be closed? The original jsfiddle seems to be working too.

evs-chris:
Feel free to re-open if the fix in edge is not sufficient.

PR #2067, opened 2015-08-13
Take care of a few arboretum-getters test failures

evs-chris:
There are a few things in here that I'm not sure about... I'll throw comments inline.


evs-chris:
The partial tests now pass, and I added support for slashes (solidi?) in partial names per #2013 because it was super simple and I was there.

Rich-Harris:
:metal:

PR #2071, opened 2015-08-17
Implement makeKey static method

narthur157:
I ended up implementing this in a project I'm doing, so I figured I'd make a PR for it

### DIFF
--- a/src/Ractive/static/makeKey.js
+++ b/src/Ractive/static/makeKey.js
@@ -2,16 +2,6 @@
  * @param {Array} argument - series of keys
  * @return {string} keypath
  */
-export default function Ractive$makeKey() {
-	var keypath = "";
-
-	for ( var i = 0; i < arguments.length; i++ ) {
-		keypath += arguments[i];
-
-		if ( i !== arguments.length - 1 ) {
-			keypath += ".";
-		}
-	}
-
-	return keypath;
+export default function Ractive$makeKey(...keys) {
+	return keys.join('.');
 }
--- a/src/Ractive/static/makeKey.js
+++ b/src/Ractive/static/makeKey.js
@@ -1,0 +1,17 @@
+/**
+ * @param {Array} argument - series of keys
+ * @return {string} keypath
+ */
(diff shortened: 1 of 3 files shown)


test/__tests/makeKey.js
```diff
@@ -0,0 +1,10 @@
+/*global test, module */
+module( 'makeKey()' );
+
+
+test( 'Returns keypath from a series of keys', t => {
+	var keys = ['a', 'b', 'c', 'd'],
+			expected = 'a.b.c.d';
```
narthur157: Will fix this

src/Ractive/static/makeKey.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * @param {Array} argument - series of keys
+ * @return {string} keypath
+ */
+export default function Ractive$makeKey() {
+	var keypath = "";
```
martypdx: Ractive is ES6ified, so you could do something like:

```js
Ractive$makeKey ( ...keys ) { return keys.join('.'); }
```

narthur157: Oh nice. This came from a non ES6 project, next commit does this now instead

Rich-Harris:
Thanks for the PR @narthur157 - am curious as to what the use case is? Not doubting you, I'm just not seeing what the difference is between these two!

```js
Ractive.makeKey( a, b, c );
[ a, b, c ].join( '.' );
```

The second one is shorter (and to be rigorous, we'd have to call it `Ractive.makeKeypath()` :-) and doesn't require any API knowledge. Am I missing something?

narthur157:
Use case for this would just be a convenience method over using join. Making keys is a fairly common thing. I made the function in my code partially because I was considering that I might need to sanitize values in my keypaths because some come from input, and just because the array syntax got a bit cluttery at times.

Though in my code I had makeKey globally defined, so having to prepend that with `Ractive` makes it a bit more cluttery.

narthur157:
I'll close this, I agree with you

PR #2080, opened 2015-08-27
Fix bugs with unless sections and undefined models

NikhilVerma:
I've fixed a couple of bugs where an undefined unless block won't work. And an undefined model would throw a script error.

### DIFF
--- a/src/view/items/Section.js
+++ b/src/view/items/Section.js
@@ -76,6 +76,12 @@ export default class Section extends Mustache {
 
 			this.fragment = fragment;
 		}
+		else if (this.sectionType && this.sectionType === SECTION_UNLESS) {
+			this.fragment = new Fragment({
+				owner: this,
+				template: this.template.f
+			}).bind();
+		}
 	}
 
 	detach () {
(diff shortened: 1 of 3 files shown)


heavyk:
:+1: I also get undefined model errors

cheers

NikhilVerma:
Yeah this basically fixes some of the issues I was having with ifelse blocks and some undefined values.

martypdx:
@NikhilVerma thanks for the PR!

Need to add test for the `{{#unless}}` block, which should be easy to add to the render test samples [here](https://github.com/ractivejs/ractive/blob/dev/test/testdeps/samples/render.js).

For the missing model, I'm not sure ignoring the undefined is the way to go. Could be, but not familiar enough with the new code. A test would certainly help understand the use.

NikhilVerma:
@martypdx I've added test for the #if/else block. That test current fails on the dev branch and the fix is in this pull request. I've removed my undefined model change because I can't replicate the situation which caused me to do that. (This is why I should write down things :P) Let me know if any other changes are needed !

NikhilVerma:
@martypdx Great news, I found the other bug as well :) Added it's test and fixed it. You can run these two tests in the dev branch and they will fail. The model being undefined one will throw a script error

Rich-Harris:
Thanks @NikhilVerma!

PR #2086, opened 2015-08-29
Better build

Rich-Harris:
Esperanto is now an [inactive project](https://github.com/esperantojs/esperanto/pull/191), so it makes sense to move to [Rollup](https://github.com/rollup/rollup) which does a better job of bundling ES6 modules.

A couple of other things go along with this:

* Module paths should always be relative to the importing module, *not* to the `src` folder
* During development, it's good to run the source code through Babel as individual modules, because then incremental rebuilds are much faster than if the entire library had to be transpiled. But during production we want to avoid that, since helpers like `classCallCheck` get included repeatedly

The build is a good bit leaner than the current edge build, but sourcemaps are currently kaput. Might be better to merge this anyway since it'll cause later merge conflict headaches otherwise (because all the `import` statements have been relativised) – anyone object?


martypdx:
👍

evs-chris:
+1

Issue #1172, opened 2014-09-01
Discussion: should we deprecate the old proxy event syntax?


Rich-Harris:
This follows on from the discussion on #1146 - see in particular feedback from @thomsbg (https://github.com/ractivejs/ractive/pull/1146#issuecomment-53526330) and @unity (https://github.com/ractivejs/ractive/pull/1146#issuecomment-54030188).

As of 0.5.6, we have two ways of dealing with events:

### Direct method call (new style)

```html
<button on-click='set("selected",foo)'>select foo</button>
```

### Triggering a proxy event, and handling it separately (old style)

```html
<button on-click='select:{{foo}}'>select foo</button>
```

```js
ractive.on( 'select', function ( event, selected ) {
  this.set( 'selected', selected );
});
```

These two are *functionally* identical, but *semantically* very different - the new syntax is far more concise and convenient, but introduces a coupling (between the event and the action) that wasn't there previously. For example, if you were using a Flux-style architecture, you wouldn't want to `set()` data directly.

## Option 1: deprecate

If we *did* deprecate the old syntax entirely, you *can* still maintain that decoupling:

```html
<!-- this... -->
<button on-click='fire("select",event,foo)'>select foo</button>

<!-- is identical to this: -->
<button on-click='select:{{foo}}'>select foo</button>
```

You can also call custom methods so that you're not blindly `set`ting data, but calling the same API that you'd call programmatically:

```js
var TodoList = Ractive.extend({
  add: function ( description ) {
    // any validation etc goes here...
    this.push( 'todos', { done: false, description: description });
  }
});
```

**Pros**:

* eliminates code from the codebase - makes it slimmer, more maintainable
* reduces the amount of stuff newcomers have to learn
* removes potential confusion
* we don't lose any expressive power

**Cons**:

* `on-click='fire("select",event,foo)'` is more verbose than `on-click='select:{{foo}}'`
* 'Proxy events' (i.e. 'click means select') are second-class citizens - some people prefer the proxy event style

## Option 2: Keep both syntaxes

**Pros**:

* Keeps existing users happy, means no breaking changes down the line
* Keeps proxy events as first-class citizens

**Cons**:

* Codebase size and maintainability
* Potentially confusing for newcomers

## Option 3: Keep old syntax, but disallow arguments

One idea we briefly tossed around was keeping `on-click='foo'` as a short version of `on-click='fire("foo",event)'`, but disallowing additional arguments. This alleviates the problem of situations like

```html
[...]

PR #2095, opened 2015-09-01
Component events as first-class citizens

martypdx:
Refactored out the event source (`DOMEvent`, `CustomEvent`) from `EventHandler` and added a `RactiveEvent` used to proxy the component instance events.

This gives full event handling capability to component events including method calls.

Right now, original instance event args are tacked on after component event arg (but original `event` arg is still first if it exists). Need to implement argument repackaging per [#1172 comment](https://github.com/ractivejs/ractive/issues/1172/#issuecomment-136159638).

### DIFF
--- a/src/events/fireEvent.js
+++ b/src/events/fireEvent.js
@@ -51,7 +51,7 @@ function fireEventAs  ( ractive, eventNames, event, args, initialFire = false ) 
 			let fullName = ractive.component.name + '.' + eventNames[ eventNames.length-1 ];
 			eventNames = getWildcardNames( fullName );
 
-			if( event ) {
+			if( event && !event.component ) {
 				event.component = ractive;
 			}
 		}
--- a/src/parse/converters/element/processDirective.js
+++ b/src/parse/converters/element/processDirective.js
@@ -33,10 +33,18 @@ export default function processDirective ( tokens, parentParser ) {
 			}
 
 			result = { m: match[1] };
-			args = '[' + tokens.slice( result.m.length + 1, end ) + ']';
+			const sliced = tokens.slice( result.m.length + 1, end )
 
-			parser = new ExpressionParser( args );
-			result.a = flattenExpression( parser.result[0] );
+			if ( sliced === '...arguments' ) {
+				// TODO: what the heck should this be???
+				// maybe ExpressionParser should understand ES6???
+				result.g = true;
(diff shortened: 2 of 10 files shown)


fskreuz:
+1

thomsbg:
Wahoo! Love this

JonDum:
Awesome, Marty!

martypdx:
Okay, implemented argument substitution on method calls via both `arguments[n]` (0 based) and `$n` (counting-number based):

```html
<component on-foo="foo( arguments[2], \'qux\', arguments[0] )" on-bar="bar( arguments[0], 100 )"/>
```

or

```html
<component on-foo="foo( $3, \'qux\', $1 )" on-bar="bar( $1, 100 )"/>
```

Not attached either way, `arguments` seems consistent with `...arguments`, `$1` has nice brevity. Let me know if you think one or the other, or like both.

Event proxies don't support substitution.

martypdx:
Oh, didn't mention it explicitly, but `<component on-foo="foo( ...arguments )"/>` also supported.

martypdx:
Seems good enough to try out, we can update the semantics if needed...

JonDum:
Hey been on flights for a day, didn't respond to this in time. 

As per the other discussion, if we're getting rid of the `this.on()` syntax, would it be possible to also have just the method name a la `on-click="foo"` act the same as `on-click=foo(...arguments) `?

martypdx:
@JonDum Hmm, I don't know that we've reached that conclusion. I still think `foo` should be short for `fire( "foo", event )`. Discussion continues... :)

Also, I didn't do DOM level registration for components, so not sure where `on-click=foo(...arguments)` would be relevant, fyi...

Rich-Harris:
### YES :metal: 

Re `...arguments`: I think this is the right approach for now – long term maybe we should think about ES6 syntax more generally as it sort of looks like our expression parser supports the spread operator when in reality it doesn't. Besides that I *think* it's just template literals that we need to worry about, off the top of my head. (Whether we just parse them to say 'yep, this is valid JavaScript, it uses these refs, carry on' or transform them to equivalent ES5 is something we'll have to worry about eventually – since these expressions would bypass Babel in many build pipelines, I suspect the answer may turn out to be that we should. But we can cross that bridge when we come to it.)

Issue #2103, opened 2015-09-03
[BUG] Updating a template with iterated items again breaks the flow of the content (0.8.0)


NikhilVerma:
Example: http://jsbin.com/basapeqale/2/edit?html,js,output

What you expect to see after 2 seconds is

```
1
2
3
4
Here is some foo
```

But what you see is

```
1
2
Here is some foo
3
4
```

@Rich-Harris The error seems to lie in Fragment.prototype.findNextNode in this bit of code where it returns null instead of the correct node.

```javascript
  		if (this.isRoot) {
  			if (this.ractive.component) {
  				return this.ractive.component.parentFragment.findNextNode(this.ractive.component);
  			}

  			// TODO possible edge case with other content
  			// appended to this.ractive.el?
  			return null; 
  		}
```

NikhilVerma:
Fixed here: https://github.com/ractivejs/ractive/pull/2104

PR #2104, opened 2015-09-03
#2103 - Fix bug with repeated fragment updating the wrong node

NikhilVerma:
Ref: http://jsbin.com/basapeqale/2/edit?html,js,output

This relates to #2103


Rich-Harris:
Thanks @NikhilVerma!

PR #2110, opened 2015-09-03
prevent babel generating iifes in hot code paths, etc

Rich-Harris:
One of the unfortunate consequences of transpiling from ES6 is that Babel will sometimes create IIFEs inside (e.g.) for loops to simulate block scoping. I *think* this PR catches all of those cases. Not sure how best to prevent them creeping back in.

Also, it removes a case where a destructured array was causing a helper to be included unnecessarily.

### DIFF
--- a/src/view/items/Component.js
+++ b/src/view/items/Component.js
@@ -96,6 +96,8 @@ export default class Component extends Item {
 		if ( this.template.a ) {
 			Object.keys( this.template.a ).forEach( localKey => {
 				const template = this.template.a[ localKey ];
+				let model;
+				let fragment;
 
 				if ( template === 0 ) {
 					// empty attributes are `true`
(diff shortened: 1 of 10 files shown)


src/view/items/Component.js
```diff
@@ -214,11 +216,12 @@ export default class Component extends Item {
...
-			const viewmodel = this.instance.viewmodel;
+		const viewmodel = this.instance.viewmodel;
 
+		if ( this.template.a ) {
 			Object.keys( this.template.a ).forEach( localKey => {
 				const template = this.template.a[ localKey ];
+				let model;
 
 				if ( isArray( template ) && template.length === 1 && template[0].t === INTERPOLATOR ) {
 					let model = resolve( this.parentFragment, template[0] );
```
martypdx: You still have a `let` here I don't think you wanted

Rich-Harris: good catch, fixed

PR #2111, opened 2015-09-03
Model fn bindings

heavyk:
ran into this trying to get our code to run on 0.8

cheers

### DIFF
--- a/src/view/items/shared/EventDirective.js
+++ b/src/view/items/shared/EventDirective.js
@@ -168,7 +168,8 @@ export default class EventDirective {
 						return model.wrapper.value;
 					}
 
-					return model.get();
+					let value = model.get();
+					return typeof value === 'function' ? value.bind(this.ractive) : value;
 				});
 
 				args = this.argsFn.apply( null, values );
--- a/test/__tests/events.js
+++ b/test/__tests/events.js
@@ -1785,18 +1785,18 @@ try {
 
 	test( 'correct function scope for this.bar() in template', t => {
 		let ractive = new Ractive({
-	    el: fixture,
-	    template: `
-			<button on-click='set("foo",bar())'>click me</button>
-			<p>foo: {{foo}}</p>
+			el: fixture,
+			template: `
+				<button on-click='set("foo",bar())'>click me</button>
+				<p>foo: {{foo}}</p>
 			`,
-	    data: {
-	        foo: 'nope',
-	        bar: function () {
-	            return this.get( 'answer' );
-	        },
-	        answer: 42
-	    }
+			data: {
+				foo: 'nope',
+				bar: function () {
+					return this.get( 'answer' );
+				},
+				answer: 42
+			}
 		});
 
 		simulant.fire( ractive.find( 'button' ), 'click' );
--- a/test/__tests/events.js
+++ b/test/__tests/events.js
@@ -1783,6 +1783,27 @@ try {
 		t.ok( event2Fired );
 	});
 
+	test( 'correct function scope for this.bar() in template', t => {
+		let ractive = new Ractive({
+	    el: fixture,
+	    template: `
+			<button on-click='set("foo",bar())'>click me</button>
+			<p>foo: {{foo}}</p>
+			`,
+	    data: {
+	        foo: 'nope',
+	        bar: function () {
+	            return this.get( 'answer' );
+	        },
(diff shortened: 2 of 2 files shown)


Rich-Harris:
thanks :+1:

heavyk:
actually, I'm not sure this is the right fix because test 'hierarchy order: onunrender' is failing (and a few others I think). 

this is my first time modifying ractive. I definitely want to use `this` in the function. what would be a better way to bind it? would it be better to bind it on the model parse?

Rich-Harris:
Oh, whoops – I thought Travis had given it a green light. Elsewhere (functions in expressions), the [wrapping happens once](https://github.com/ractivejs/ractive/blob/2de4047bad6f90fcc50a26c0cae95dadfa5e1a53/src/view/resolvers/ExpressionProxy.js#L8-L36) per function/instance to avoid situations where it would otherwise happen hundreds of times, but in the context of an event simply binding it on demand seems like a saner approach. So the PR looked good to me. Let me check it out locally and see if I can figure out why those tests are failing

martypdx:
It needs to be bound to the model's Ractive instance, not the events. 

However it may make more sense that model gets on functions always self-bind before returning value.

Rich-Harris:
> It needs to be bound to the model's Ractive instance, not the events.

Not sure I follow – aren't models instance-agnostic?

So the good news is also the bad news: for some unfathomable (by me at least) reason, the test in this PR is causing all those others to fail. Comment it out, and we get a full bank of green lights. Bit confused as to why that would happen, the test seems innocent enough

Rich-Harris:
Adding `ractive.teardown()` to the end of the test fixes everything. Am thoroughly confused.

martypdx:
> Not sure I follow – aren't models instance-agnostic?

Not if `bar` in `<button on-click='set("foo",bar())'>click me</button>` was mapped from a parent view. It needs to be called bound to the instance in which it was defined.

heavyk:
the test was pretty much copied and pasted from the other pr. the test doesn't do anything extraordinary.

@martypdx, it is bound to the instance (`this.ractive`) before being called -- otherwise it wouldn't find `bar` (at least that was my understanding)

martypdx:
Wait, that's not the whole story. It should be bound to it's parent model value, _unless_ it's model parent is root, in which case it should be bound to Ractive instance of root.

Rich-Harris:
Okay, I think I've got my head round this (the ownership issue, not the mysterious failures) – the function should be bound upon initialisation...

```js
ractive = new Ractive({
  data: () => ({
    // this is bound to ractive upon initialisation
    foo () {
      return this.get( 'answer' );
    }
  })
});
```

...or `ractive.set(...)`:

```js
ractive.set( 'bar', someOtherFunction ); // is immediately bound
```

That way, the model doesn't need to know anything about which instance it belongs to, and we only need to bind functions occasionally. Does that seem right?

heavyk:
I think initialisation.
omg, wish I had I have known about using `this.set` ... lol

martypdx:
Yes, but also consider `foo.bar()` which needs to still be called with `foo` as the `this` context

martypdx:
Or does that get parsed as `_0.bar()`?

Rich-Harris:
> Or does that get parsed as `_0.bar()`?

[Bingo](https://github.com/ractivejs/ractive/blob/4a1e0c1a15a6c25e00e8fdaf502d7f3443f87770/src/parse/converters/expressions/primary/readReference.js#L59-L68) :grinning:

heavyk:
right, but I fucked it up... I can take my hand at binding on init and writing tests for foo.bar

I'm willing to help out as much as I can cause I really wanna use 0.8 :) I posted 3 fiddles last night of other bugs. I still think there are still quite a few more (it's pretty broken)... btw, if ya'll give me time (and if it would help), I can put the code up on github. the ractive 0.7.3 test version I'm working on is here: ractive.affinaty.com:1155

martypdx:
And you submitted a test. That's like a dinner guest showing up with a fine bottle of wine or liquor, you know they'll be invited back :)

heavyk:
lol cool :)

I'll make the other fiddles into tests and pr them later :)

I'd submit a fix as well, but when I tried last night, I had some difficulty. I still haven't quite wrapped my head around the code (yet).

PR #2115, opened 2015-09-04
Bind functions on init/set

Rich-Harris:
This is a follow-up to #2111. Rather than binding functions to the current instance (or reusing existing bound function copies) in situations like `{{foo()}}` and `on-click='set("bar",foo())'` (where `foo` depends on `this`) at the point of *getting* them, we bind them at the point of *setting* them (or upon initialisation). This is cheaper (it only needs to happen once) and cleaner (we don't need to track bound copies and clean them up later).

The only situation where it might be a problem is if someone is using a non-POJO as their root `data` object, which is already strongly discouraged (and warned about in debug mode).


martypdx:
image :grin:

heavyk:
nice!

good call on checking if `this` is found somewhere in the source.
will be lots of false positives, but still better than naively binding like I did

PR #2116, opened 2015-09-04
Simplify event code and perf test events

martypdx:
I had a chance to investigate something I was curious about: the existing use of routing of DOM event handling through a common function and then calling a handler stored on the event node. 

Via the perf tests it didn't seem to make a difference, though @Rich-Harris  maybe you know something I don't, or it's an IE thing, or something like that? (Side note: Our perf tests don't run on firefox, due to number of interaction issues between iframe and main window)  

Removing it simplifies the code and we don't have to store state on the DOM nodes for events.

For event plugins, we have a closure anyway, so I added default inclusion of the `node` so plugin authors don't need to add it. Even if we need to keep the common `defaultHandler` we should keep this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,5 +19,6 @@ ractive.js
 !src/Ractive.js
 .gobble-build
 perf/control/*
+perf/control/1740
 !perf/control/.empty
 npm-debug.log
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,6 @@ ractive.js
 !src/Ractive.js
 .gobble-build
 perf/control/*
-perf/control/1740
+perf/1740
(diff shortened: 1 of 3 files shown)


Rich-Harris:
:+1: for evidence-based perf decisions. If it simplifies the code without any adverse effects then I'm all for it. Presumably we can lose [this line](https://github.com/ractivejs/ractive/blob/e3f83aee5de37d7b83f89c641d45ed3baa3cd78f/src/view/items/Element.js#L224) as well.

The FF thing is something we'll have to look into...

martypdx:
> Presumably we can lose this line as well

done

PR #2117, opened 2015-09-04
Computed parent

heavyk:
turns out that we have to check if it's a computed property when getting (not on the model.joinPath) and lazily compute the value (not when joining it) - otherwise this.value is always undefined (because it never even computes) ...

this is because the computed property returns an object.

just thought of something: it could potentially not resolve if the computed property returns a deep object (and the template references at least 2 levels inside). I dunno

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -2,3 +2,8 @@ root = true
 
 [*.js]
 indent_style = tab
+indent_size = 2
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
--- a/src/model/Model.js
+++ b/src/model/Model.js
@@ -6,6 +6,7 @@ import getPrefixer from './helpers/getPrefixer';
 import { isArray, isObject } from '../utils/is';
 import KeyModel from './specials/KeyModel';
 import KeypathModel from './specials/KeypathModel';
+import Computation from './Computation';
 
 const hasProp = Object.prototype.hasOwnProperty;
 
@@ -166,6 +167,10 @@ export default class Model {
 
 	get ( shouldCapture ) {
 		if ( shouldCapture ) capture( this );
+		if ( this.value === void 0 && this.parent instanceof Computation ) {
+			this.parent.get();
+			return this.retrieve();
+		}
 		return this.value;
 	}
 
--- a/src/model/Model.js
+++ b/src/model/Model.js
@@ -167,10 +167,23 @@ export default class Model {
 
 	get ( shouldCapture ) {
 		if ( shouldCapture ) capture( this );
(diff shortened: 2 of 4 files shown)


heavyk:
oh yeah and my editor was confused with the tabs, so I cleaned them.. dunno if that's unkosher :/ there were already 8-spaces mixed with tabs...

heavyk:
just a sec before merging... gonna add one more thing

heavyk:
ok, now we're good :+1:

heavyk:
shit! looking at the diff, it appears I made the tabs worse. one sec. will fix

heavyk:
no shit. the tabs appear as 2 spaces now on github (or maybe it was cause I added that line (`indent_size = 2`) on the `.editorconfig` -- cause I always remember github having 8 space tabs... weird

anyway, I verified that they are indeed tabs... sorry bout that

Rich-Harris:
Thanks for this @heavyk 

> just thought of something: it could potentially not resolve if the computed property returns a deep object (and the template references at least 2 levels inside). I dunno

Good catch – just tested this and it *doesn't* work:

```js
ractive = new Ractive({
  el: fixture,
  template: '{{foo.bar.baz}}',
  data: {
    answer: 42
  },
  computed: {
    foo () {
      var answer = this.get( 'answer' );
      return {
        bar: {
          baz: answer
        }
      };
    }
  }
});

ractive.toHTML(); // ''
ractive.set( 'answer', 99 );
ractive.toHTML(); // ''
```

Which makes me think we should probably hold off on merging this until we've figured out a more general solution... sorry!

heavyk:
I think I know the way... one moment :)

martypdx:
> he tabs appear as 2 spaces now on github (or maybe it was cause I added that line (indent_size = 2) on the .editorconfig -- cause I always remember github having 8 space tabs

@heavyk github is the honey badger of indentation

heavyk:
nailed it!

there is one thing I don't understand. when the instance inits, then it will call the computed property, then again on render, then again after set (a total of three times).

computed! 42
computed! 42
computed! 99

first, I don't understand why it's set in init, but I'm pretty sure that's a whole nother issue because it didn't cover the code path I was modifying.

this gives me the impression (also) that if if must run on init, the value could be saved until dirty (or maybe it's doing that and keeps getting dirty)... dunno. I'd like to look at it at another time.

JonDum:
> there is one thing I don't understand. when the instance inits, then it will call the computed property, then again on render, then again after set (a total of three times).

I noticed that too when I was optimizing computed props in 0.7. Also really curious why.

heavyk:
@martypdx feel free to change that number to 12 if you feel 2 spaces is hard to distinguish indentation levels :)

PR #2118, opened 2015-09-04
Unrender all section fragments if type changes (#2054)

### LINKED ISSUE #2054: unexpected behavior when setting a keypath to an array, then an object, then back to an array
This pen:
http://codepen.io/jmellicker/pen/eNPrpJ?editors=101

- sets the 'products' keypath to an array (works great)
- then sets it to an object (works great)
- then sets it back to an array, but at this point the DOM only shows the last product (does not iterate through the whole list as it did initially).

My guess is that Ractive is caching in some way that prevents it from rendering properly when the keypath is set back to an array, in which case I thought you would want to know (since you all are heroically committed to finding and fixing every last anomaly in this wonderful framework).

martypdx:
Fixes #2054

### DIFF
--- a/src/view/RepeatedFragment.js
+++ b/src/view/RepeatedFragment.js
@@ -29,13 +29,21 @@ export default class RepeatedFragment {
 		this.owner = options.owner;
 		this.ractive = this.parent.ractive;
 
+		this.context = null;
+		this.rendered = false;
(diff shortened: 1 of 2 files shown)


MartinKolarik:
:+1:

PR #2121, opened 2015-09-05
fix #2069 + amplify test suite

### LINKED ISSUE #2069: Setting array to [] still cause calls from template
Hi,

Below is a complete example of a thing that I don't understand in ractive. I have an array of items which is first rendered correctly. However, when I empty that array (ractive.set('items', []);), unnecessary calls are then being made (func: {{ func(this) }}). My expectations was that those calls should not happen since items is an empty array and the template should not trigger calls.

To run the example, just copy the code into an html file and run in a browser. Watch the console where func logs 'undefined' three times when emptying the items array.

Can someone explain why that happens? I think it is a defect.

I use ractive 0.7.3, but have tried older versions as well.

```
<!doctype html>
<html lang='en-GB'>
<head>
  <meta charset='utf-8'>
  <title>Ractive test</title>
</head>

<body>
  <h1>Ractive test</h1>

  <div id='container'></div>

  <script id='template' type='text/ractive'>
    <p>Hello, {{name}}!</p>

    <ul>
      {{#items}}
        <li>
          {{#a}}
            func: {{ func(this) }}
            f: {{ this.f() }}
            d: {{ this.d }}
          {{/a}}
          b: {{ b }}
        </li>
      {{/items}}
    </ul>
  </script>

  <script src='http://cdn.ractivejs.org/latest/ractive.min.js'></script>

  <script>
    var ractive = new Ractive({
      el: '#container',

      template: '#template',

      data: function () {
        return {
          name: 'world',

          items: [
            {
              a: {
                f: function() { console.log('f1'); return 'f1';},
                d: '22'
              },
              b: '2'
            },
            {
              a: {
                f: function() { console.log('f2'); return 'f2';},
                d: '44'
              },
              b: '4'
            },
            {
              a: {
                f: function() { console.log('f3'); return 'f3';},
                d: '66'
              },
              b: '6'
            }
          ],

          func: function(obj) { console.log(obj); return obj;}
        };
      }
    });

    setTimeout(function() {
      ractive.set('items', []);
    }, 2000);
    
  </script>
</body>
</html>
```

heavyk:
I started out trying to fix some things with multiple components, only to find out that it was working properly. instead of deleting the test, I committed it. (2nd commit) 

then, I set out to add the fiddle in #2069 (3rd commit). I realized it was pretty easy, so I fixed it (4th). better test coverage :)

I'm totally stoked. a lot of things are working now... I just have the really weird intermittent errors left...

sweeet :)

### DIFF
--- a/src/model/Model.js
+++ b/src/model/Model.js
@@ -227,7 +227,7 @@ export default class Model {
 		if ( !isEqual( value, this.value ) ) {
 			this.value = value;
 
-			this.children.forEach( mark );
+			if ( value ) this.children.forEach( mark );
 
 			this.deps.forEach( handleChange );
 			this.clearUnresolveds();
--- a/test/__tests/arrays.js
+++ b/test/__tests/arrays.js
@@ -270,6 +270,63 @@ test( 'Array updates cause sections to shuffle with correct results', t => {
 	t.htmlEqual( fixture.innerHTML, 'threeoneAtwoBC' );
 });
 
+test( 'Setting an Array to [] does not recompute removed values', t => {
+	// t.expect(2)
+	let called = { func: 0, f1: 0, f2: 0, f3: 0 };
+	let ractive = new Ractive({
+		el: fixture,
+		template: `
+		<p>Hello, {{name}}!</p>
+		<ul>
+			{{#items}}
+			<li>
+				{{#a}}
+					func: {{ func(this) }}
+					f: {{ this.f() }}
+					d: {{ this.d }}
+				{{/a}}
+				b: {{ b }}
+			</li>
+			{{/items}}
+		</ul>
+		`,
+		data: function () {
+			return {
+				name: 'World',
+				items: [
+					{
+						a: {
+							f: function() { called.f1++; return 'f1'; },
+							d: '22'
+						},
+						b: '2'
+					},
+					{
+						a: {
+							f: function() { called.f2++; return 'f2'; },
+							d: '44'
+						},
+						b: '4'
+					},
+					{
+						a: {
+							f: function() { called.f3++; return 'f3'; },
+							d: '66'
+						},
+						b: '6'
+					}
+				],
+
+				func: function(obj) { called.func++; return obj; }
+			};
+		}
+	});
+
+	t.deepEqual( called, { func: 3, f1: 1, f2: 1, f3: 1 })
+	ractive.set('items', []);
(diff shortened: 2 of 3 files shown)


martypdx:
I think #2069 is a bit tricker (but let me know if you figure it out), I spent some time looking at it today and tried a few things.

Basically, it needs to not notify dependents beyond the end of the new array. So in this particular test case, checking `if ( value )` works, but try it by setting a new array of say length `1` and make sure the first item _does_ update.

Unfortunately `this.children` in `Model` has both properties (`length` specifically) _and_ array members. I'm not sure the order is always correct (after shuffles for example). I think what needs to happen is we need to store children with numeric keys (aka indices) in a separate array so we can compare lengths and only update those that are still within the length of the new array.

heavyk:
I see what you're saying. I thought it was gonna be just removing the value, but I see how this can start to complicate..

it also worked with our code, because I always set the array to 0 before putting things back into it. there's this super cool bug in 0.7 that goes something like this: if data: { list } is set, then ractive.set('list', []), then list.push({}) will modify the dataset (or something really horrible like that).

do you have the tests that you were using today somewhere, where I can get them to verify the fix.

I'll update this tomorrow. it'll be good to go at it fresh again. I'm gonna sleep now. g'nite :)

martypdx:
The other approach to this would be for block dependents to register separately from "value" dependents.

I went the route of tracking an array of array members in #1740 and it introduced another level of overhead that had to be maintained. If instead sections were just notified first, in theory they would unbind the unneeded fragments, causing the children to be unregistered. Then when we notified Model children they would presumably have been removed.

@heavyk here's the test:

```js
if ( hasUsableConsole ) {
	test( 'Expressions are not triggered when section iterations are removed', t => {
		const o = { f(){} };
		const warn = console.warn;

		expect(1);

		console.warn = function( msg ) {
			t.ok( false, 'unexpected warning triggered: ' + msg );
		};

		var ractive = new Ractive({
			el: fixture,
			template: '{{#items:i}}{{ this.f() }}{{/items}}',
		    data: {
		    	items: [ o, o ]
		    }
		});

		ractive.set( 'items', [ o ] );
		
		t.ok( true );

		console.warn = warn;
	});
}
```

Rich-Harris:
@heavyk @martypdx I just submitted #2124 which addresses #2069 from a different angle – basically reusing the lazy propagation mechanism in #2122.

Rich-Harris:
Have gone ahead and merged #2124, so I'll close this. Both the tests in this PR are now in the suite – thanks!

PR #2124, opened 2015-09-05
Alternative #2069 fix

Rich-Harris:
I believe this is the correct fix for #2069 – the `ExpressionProxy` class, which acts as an interface between the view and a `Computation`, should behave similarly to the `ComputationChild` in #2122 (this branch is based off that PR's branch). This ensures that `fn` in `{{#each items}}{{fn(this)}}{{/each}}` is never called with no-longer-extant memnbers of `items`.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -2,3 +2,8 @@ root = true
 
 [*.js]
 indent_style = tab
+indent_size = 2
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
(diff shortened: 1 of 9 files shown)


Rich-Harris:
Should have said: not ready for merge, it breaks another test. Looking into it.

heavyk:
nice! you combined my previous work to fix it too. that's cool! it certainly *looks* more like the correct fix as well, heheh :)

glad the tests made it in .. cheers

JonDum:
> This ensures that fn in {{#each items}}{{fn(this)}}{{/each}} is never
called with no-longer-extant memnbers of items.

Oh *fuck* yes. That's been a thorn in my side for the longest time. If it's
really fixed I'm buying you a beer, Rich.

PR #2127, opened 2015-09-05
Better tests

Rich-Harris:
**not ready for merge, work in progress**

Our test suite is very woolly – it's sometimes hard to tell which file a particular test is in, and the organisation of files is somewhat arbitrary. The tests themselves are a bit messy and inconsistent in terms of code style, lending a 'here be dragons' feel to the whole thing.

This PR does a few things:

* Ditches Esperanto (which is abandoned) in favour of Rollup
* Names test modules based on the filename, making them easier to find. (This doesn't happen quite everywhere yet, because some modules have custom setup/teardown needs.)
* Moves some test files around, e.g. tests of `Ractive.prototype` methods sit in a `methods` folder

Plenty still to do, just thought I'd do it in the open so everyone gets a heads-up...


heavyk:
:+1:  less dragons :)

yeah, so when I first opened the project for the first time last week, I was surprised to find the tests in a directory named `__tests` - it just seemed to be almost as if they were trying to be hidden or something... heh

that and I still haven't really figured out why no matter how many more assertions (or tests) I add, it stays at 302 assertions... come to think of it, in localhost:4567/tests/ there are like 30k assertions... dunno, you must be hiding more tests somewhere else :)

Rich-Harris:
Bit more progress:

* the folder structure makes more sense – inside `tests/` we have `browser-tests` and `node-tests`. `__support` contains everything else we need to run them. (@heavyk you were right – it was very confusing before)
* there's only one module per file. This allows the automatic module naming to work, making it easier to find the source for a test, and eliminates the confusion that arises from having multiple modules with different `beforeEach`/`afterEach` hooks in the same file
* `asyncTest` is no longer used – instead we use `const done = t.async() ... done()` as per the QUnit docs
* there's an `npm run lint-tests` task
* there's a README, with instructions on running the test suite and adding new tests. Hopefully the whole thing will be a bit less bewildering

The organisation of the tests themselves is still slightly haphazard (e.g. there's still a `misc.js` file with several dozen tests that belong elsewhere), but we can address that gradually.

Any objections to merging this?

evs-chris:
:+1:

Issue #1636, opened 2015-01-14
Legacy polyfills addition, Array.isArray, Array.prototype.some etc.


akhoury:
I noticed that the legacy ES5+ polyfills are not complete. Would it be too much to ask for these to be added or to be completely removed from ractive source?

I understand that Ractive core may not actually be using all of them, however, IMO, I think you should either should include them all or **none**, in the latter case, legacy support could be achieved by including something like the [es5-shim](https://github.com/es-shims/es5-shim) as a hard requirement for legacy support. 
I think that's a better idea than having to load ractive-legacy along with another es5-shim to complete it, but still partially redundant.

Rich-Harris:
I do like the idea of not having a legacy version, and letting people bring their own shims. Unfortunately es5-shim doesn't include `addEventListener` and `getComputedStyle`, which Ractive needs, so it's not all that practical.

On the flip side, adding in all the missing pieces that es5-shim includes would bulk things out a fair bit - it's a [pretty big file](https://github.com/es-shims/es5-shim/blob/master/es5-shim.js), compared to the 200-odd lines for the [bits we currently include](https://github.com/ractivejs/ractive/blob/dev/src/legacy.js#L16-L256).

Not quite sure where I land on this. Would welcome other opinions.

Madgvox:
It would be extra work, but you could split the polyfills out into their own library, a la `ractive-es5-shim`. That way the shims aren't in the main library, and if you need the shims you can easily import a shim library that includes `addEventListener` and `getComputedStyle`.

akhoury:
@Rich-Harris, I would 
* drop the legacy code
* move the `addEventListener ` and `getComputedStyle` to the main ractive src (meaning not the legacy), I am assuming these 2 won't execute on good browsers. It is true that the LOC number will increase a bit for everyone, but for a good cause. Also, it doesn't look like your `getComputedStyle` function is getting added to the `window` anyways, according to [this block](https://github.com/ractivejs/ractive/blob/dev/src/legacy.js#L348-L353), so that's another reason why you can't count on an external lib to shim it for you.
* ship 1 "type" of ractive from now on, no more `ractive-legacy`
* require everyone who wants ie8 support to include the es5-shim

evs-chris:
The legacy polyfill is moving to its own file, so you will be able to shim/sham before it or wherever you like. See #2801

PR #2131, opened 2015-09-07
Remove legacy builds

Rich-Harris:
Not ready for merge. Thought this might be a 20 minute fix, I was wrong. Just going to leave it here for now.

https://github.com/ractivejs/ractive/issues/1636 raised the suggestion of removing the legacy builds and letting people bring their own polyfills. This appeals, but is slightly tricky. A few takeaways from my brief experiments just now:
- Babel doesn't produce IE8-safe code – it uses `Object.create` (which can't be polyfilled) for classes
- We should avoid `import *` as it necessitates the use of ES5 getters
- @Madgvox was right, we should probably create a ractive-polyfills.js or similar. Otherwise people who need to support IE8 have to faff around with a bunch of different scripts

(The reason this PR is so large is that it's based on #2127)


ghost:
Superseded by https://github.com/ractivejs/ractive/pull/2677

PR #2137, opened 2015-09-08
fix #2065

### LINKED ISSUE #2065: get('array-of-objects').splice(x,y) removes to much from the dom.
This is the fiddle:
https://jsfiddle.net/ztzqdop6/22/

Removing any but the last element of an array (via `Array.splice()`) results in broken DOM. When iterating over an Array in a template, not only the missing item is removed but also parts of the following item.

heavyk:
**WIP**

### DIFF
--- a/test/browser-tests/methods/splice.js
+++ b/test/browser-tests/methods/splice.js
@@ -83,3 +83,32 @@ test( 'splice with one argument (#1943)', t => {
 
 	t.htmlEqual( fixture.innerHTML, '1' );
 });
+
+test( 'splice on array removes correct elements (#2065)', t => {
+	const Item = Ractive.extend({
+		template: '{{value}}'
+	});
+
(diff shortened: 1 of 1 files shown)


heavyk:
hmmm, I thought I could fix this easily, however I appear to be overlooking something.

I'm gonna leave it for the moment.
anyone else wanna have a go at it?

heavyk:
ok, I give up. the problem happens here: https://github.com/ractivejs/ractive/blob/dev/src/model/Model.js#L369-L371

if I comment those lines, the test will pass but other tests fail (obviously).
it looks like it's updating already updated deps

I ran into this bug this morning on 0.7.3. it also affects 0.8.0... does anyone have a workaround?

heavyk:
fixed by https://github.com/ractivejs/ractive/pull/2170

Issue #2151, opened 2015-09-20
Adjacent components, only the first is rendered


lazylester:
HTML:
```
<div id='container'/>

<script id="bandz_template">
    <h3>Bandz</h3>
    <beatles>
    <stones>
</script>
```
javascript:
```javascript
Beatles = Ractive.extend({
    template : "{{#beatles}}<p>{{name}}</p>{{/beatles}}"
})

Stones = Ractive.extend({
    template : "{{#stones}}<p>{{name}}</p>{{/stones}}"
})


new Ractive({
  el : "container",
  template : '#bandz_template',
  data : {stones : [{name : "mick"},{name : "keith"}], beatles : [{name: "ringo"},{name:"george"}]},
  components :{
    beatles : Beatles,
    stones : Stones
  }
})
```
(this is on jsfiddle at http://jsfiddle.net/fdf9cdxe/)

In the top-level template there is a \<beatles\> component followed by a \<stones\> component. However if you look at the jsfiddle, only the first is rendered. As an experiment, on jsfiddle, reverse the two components so it's \<stones\> followed by \<beatles\>. Then only the \<stones\> component is rendered.

The only way to get them both on the page seems to be to encapsulate each of them in an html element, e.g. div.

martypdx:
You need to close your component tags:

```
<beatles/>
<stones/>
```

http://jsfiddle.net/fdf9cdxe/2/

Rich-Harris:
Have raised #2154, which adds an error if the end of the template is reached with missing end tags.

PR #2154, opened 2015-09-22
error on missing end tags (#2151)

Rich-Harris:
If the end of a template is reached, and an element is still open (i.e. there was a non-void, non-self-closing opening tag with no closing counterpart), an error is thrown. This avoids confusion in cases like #2151 (it will also fix the cryptic errors that happen in a single-file component if there's an unclosed element before the `<script>` block, causing the code to be executed in the global scope).

It *does* mean that in certain quasi-legitimate situations where someone is relying on end tags being optional (see the [ludicrously complex rules governing tag omission here](https://developers.whatwg.org/syntax.html#syntax-tag-omission)) they'll get an unexpected error. But I think that's a very small price to pay for the reduction in confusion (and the error message tells you exactly what you need to add).


evs-chris:
FWIW, I'm :+1: on this since it seems to be a somewhat common problem for people starting Ractive. I also hail from the time of XHTML when skipping closing tags was a mortal sin, so there's that.

heavyk:
:+1:

PR #2160, opened 2015-09-24
Restore oninit behavior to be after bind so virtualDOM is complete

martypdx:
Restores behavior of `oninit` to be fully flushed out virtual DOM bound to data.

Along the way, had to fix test(s) that exposed issue in twoway binding with forced reference resolution.


test/browser-tests/twoway.js
```diff
@@ -603,7 +603,11 @@ test( 'Changes made in oninit are reflected on render (#1390)', t => {
 	new Ractive({
 		el: fixture,
 		template: '{{#each items}}<input type="checkbox" name="{{selected}}" value="{{this}}">{{/each}}',
-		data: { items: [ 'a', 'b', 'c' ] },
+		data: {
+			items: [ 'a', 'b', 'c' ],
+			// will fail without this
+			selected: null
+		},
```
heavyk: I don't think this is intended behaviour. there should be no reason to set `selected`.
I think this is a bug...

src/Ractive/initialise.js
```diff
@@ -36,23 +38,20 @@ export default function initialise ( ractive, userOptions, options ) {
...
-		ractive.fragment = new Fragment({
+		ractive.fragment = fragment = new Fragment({
 			owner: ractive,
 			template: ractive.template,
 			cssIds,
 			indexRefs: options.indexRefs || {},
 			keyRefs: options.keyRefs || {}
 		});
+
+		fragment.bind( ractive.viewmodel );
```
Rich-Harris: `.bind(...)` returns `this`, so you can – if you want – do `fragment = new Fragment(...).bind(...)`

test/browser-tests/twoway.js
martypdx: Agree, but that would have sent me into some serious yak shaving.

Two things happened: 1) the new code immediately forces binding to happen AND 2) forced resolution now resolves to context scope, not instance scope (a good thing). 

Except that for multi-select twoways it will end up trying to bind to the first instance in the loop when probably some higher context was intended.

FWIW, it does warn about the ambiguity. 

Ideally I think bindings would wait till the last responsible moment to force bind, then it would work as long as set happened before some other action like user clicking on a checkbox or a checkbox having a `checked` attribute.

In any case, I added an additional test that demonstrated the issue outside of change to `oninit` to show it wasn't strictly related to this change - and hence wanted to divorce the two things and avoid:

image

heavyk: LOL! I see what you're saying there: yak shaving makes more sense for a PR relevant to that specific error :)

PR #2167, opened 2015-09-25
new ractive.transition() method RFC WIP

martypdx:
#### tl;dr
* Ractive prototype type method `.transition( name, node, params )`
* Can be used as API and in event handlers
* Proposed part 2 change transitions api from `intro='fade'` to `on-render='transition("fade")'`

An idea I'd been thinking about for past few weeks, isolates the transition logic into runnable api method.

`name` parameter can be a registered transition name _or_ a function so you can do both:

```js
this.transition( 'fade', node );
```

and

```js
this.transition( t => t.animateStyle( 'opacity', 0, { duration: 200 } ), node );
```

( Notice I also enhanced transition plugin function to allow a `Promise` return for calling `t.complete()` ) 

`node` parameter is optional and will default to `this.event.node` if missing, whole point being that you can use them in event handlers:

```html
<div on-click='transition("fade")'></div>
```

I considered overloading `node` to take a string that would act as a query selector via `this.find(query)` but it becomes ambiguous in the case of `on-click='transition("fade", ".myClass")'` versus `on-click='transition("fade", "fast")'`. 

Perhaps it's more powerful to animate other nodes than provide the duration shortcut?

The method returns a promise:

```js
this.transition( 'fade', node ).then( () => {
    // called when transition complete...
})
```

It changes the purely `isIntro` semantics of transition plugins, so those will need to be rethought. As-is it seems to work fine in testing. I suspect it will evolve into an `t.eventName` parameter. We also may want to borrow the 'fillMode' animation semantics for how to deal with reset of styles.

It may also make sense to allow transition plugin authors to supply aliases which represent default parameters:

```js
function fade ( t, params ) {
    // ...
}

// this would also register fadeIn and fadeOut as transitions
fade.alias = {
    fadeIn: { start: 0, opacity: 1 },
    fadeOut: { start: 1, opacity: 0 },
}

export default fade;
```

#### Phase 2
Eventually, I think `intro-outro` should be deprecated for two reasons:
* Specific lifecycle events can be introduced. Instead of `noIntro` true or false, use specific events: `on-render`, `on-unrender`, `on-add`, `on-remove`. Capture array modification events, eg `on-shuffle`
* Consistency of API and deprecation of `:` style attributes. ( Not sure how we deal with decorators which  are a whole breed unto themselves. While at first glance they can be setup via `on-render`, unlike events they imply a data-binding to the supplied parameters)


Rich-Harris:
This is very interesting. I'm very much in favour of unifying APIs, and extremely keen on ditching anything that follows `:`. I have to wonder what the value of decoupling transitions from render/unrender is though – isn't the value of data-bound transitions that you can't get into a state where the view is desynced from the model, which is almost guaranteed to happen if they can be triggered programmatically?

It could very well be that I'm just not thinking big enough though. (It'd be great to have an easy-to-use way of transitioning from one node to another node, a la [ramjet](http://www.rich-harris.co.uk/ramjet/) or [liquid fire](http://ef4.github.io/liquid-fire/#/transition-map/dom-constraints)'s 'matching', for example.)

How would you account for `noIntro` (notwithstanding the fact that it's currently broken across component boundaries – #1853)?

Also, playing devil's advocate, adding a third type of event to elements (i.e. in addition to native DOM events and custom events like `tap`) adds a bit of complexity (also, what's the difference between `render` and `add`?), and `on-render='transition("fade")' on-unrender='transition("fade")'` is a lot more to write than `intro-outro='fade'`...

Decorators *are* tricky. I think we could tackle that separately though – my main gripe with them is the proliferation of curlies, I reckon we could keep the existing functionality with a slightly tweaked syntax:

```js
<!-- ...instead of this... -->
<div decorator='foo:{color:{{color}},size:{{5*i}}}'>...</div>

<!-- ...we do e.g. this: -->
<div decorator='foo {color, size: 5 * i}'>...</div>
```

evs-chris:
I'm not in a position to make any useful comments right now, but I will say :+1: to being able to trigger transitions outside of the render/unrender cycle.

As to the duration vs selector API, my preference would be to make both explicit in an options argument. It's a shame JS doesn't have sugary named arguments. My second choice would be to always push the duration out so that the node position is consistent.

martypdx:
> I have to wonder what the value of decoupling transitions from render/unrender is though – isn't the value of data-bound transitions that you can't get into a state where the view is desynced from the model, which is almost guaranteed to happen if they can be triggered programmatically?

Think I might need an example, not really following concern. How is it different than current `intro` where (if I understand you) data could diverge during the time the transition takes?

> It'd be great to have an easy-to-use way of transitioning from one node to another node, a la ramjet or liquid fire's 'matching', for example.

Not sure how we get there :grin:, but agree 100%

> How would you account for noIntro?
> adding a third type of event to elements (i.e. in addition to native DOM events and custom events like tap) adds a bit of complexity (also, what's the difference between render and add?)

One would be absolute, the other would only be triggered if `ractive[.fragment].rendered` is `true`.  `render` === `intro`, `add` === `intro` plus `noIntro: true`. (we could use `intro` instead of `render`, or `add` or whatever).

More broadly these are just new element-level out-of-the-box element events available via ractive that don't necessarily have anything to do with transitions per se:

```html
{{#each items}}
    <!-- fired any time element is rendered whether on ractive instance render or data added to items -->
    <div on-render='foo()'></div>

    <!-- fired when element is create because data added to items, after initial ractive instance render -->
    <div on-add='foo()'></div>

    <!-- fired when element is removed due to data removed from items, but not ractive instance unrender -->
    <div on-remove='foo()'></div>

    <!-- fired any time element is removed whether on instance unrender or data removed from to items -->
    <div on-unrender='foo()'></div>

    <!-- fired when element is moved via splice, splice-equivelent, or merge -->
    <div on-move='foo()'></div>
{{/each}}
```

> and on-render='transition("fade")' on-unrender='transition("fade")' is a lot more to write than intro-outro='fade'

Same as any other event handler: `on-render-unrender='transition("fade")'`, or if `intro` instead of `render`: `on-intro-outro='transition("fade")'`

> Decorators are tricky. I think we could tackle that separately though – my main gripe with them is the proliferation of curlies, I reckon we could keep the existing functionality with a slightly tweaked syntax

What if we use the moniker `as-` and pull the name out to the attribute?

```html
<!-- ...instead of this... -->
<div decorator='tooltip:{color:{{color}},size:{{5*i}}}'>...</div>

<!-- ...we do e.g. this: -->
<div as-tooltip='{ color, size: 5 * i }'>...</div>
```

Then we just have `on-foo` and `as-bar`

Rich-Harris:
> Same as any other event handler: `on-render-unrender='transition("fade")'`

Oh, right! Duh. 

> What if we use the moniker as- and pull the name out to the attribute?

sobeautiful

A thousand times yes! That would neatly solve the ongoing 'how to use multiple decorators' problem too.

> Think I might need an example, not really following concern. How is it different than current `intro` where (if I understand you) data could diverge during the time the transition takes?

It's not so much changes during the transition (though that's a separate issue: it'd be lovely to be able to 'abort' and reverse intro transitions if the surrounding `{{#if ...}}` condition becomes falsy, and – though much harder – vice versa). It's more that you potentially end up in a situation where you transition a few nodes in and out, just like the good old `$el.slideUp()` days, and suddenly the state of the DOM is no longer a function of the state of the viewmodel. Of course you *can* do that already, if you're determined, but Ractive guides you towards expressing these things as a function of the viewmodel (in fact that's its main job), which I believe makes apps a lot more predictable. That's my only real concern – that we're lowering the guiderails a bit.

heavyk:
> That would neatly solve the ongoing 'how to use multiple decorators' problem too.

+1 dude. I like that idea too.

martypdx:
@Rich-Harris I think adding `ractive.transition` would be useful as a ractive method, but not (initially) replace the current dedicated transition attributes. Along those lines, I was thinking to use the shorter suffix `in` and `out` along with transition name to have nice consistency:

```html
<div on-click='set("foo", bar)' as-tooltip='{ title }' fade-in-out>lorem isum</div>
```

JonDum:
@martypdx How do you specify transition parameters with that syntax?

martypdx:
@JonDum same as decorators. Attribute value is interpreted as javascript expression representing arguments.

JonDum:
@martypdx Oh ok nice. That's a bit cleaner then.

PR #2169, opened 2015-09-26
implement getKeypath on keypath models - fixes #2164

### LINKED ISSUE #2164: Using @keypath as a parameter to a function inside a partial with a given context fails
Check out:  http://jsfiddle.net/jondum/q3p4yk5v/

```
<main></main>

<script id='template' type='text/ractive'>
   
    {{> test foo.test }}

    {{#partial test}}
       {{ funcTest(@keypath) }}
    {{/partial}}
</script>
```

```
var ractive = new Ractive({
    el: 'main',
    template: '#template',

    data: {
       
       funcTest: function(value) {
           return value + "111";
       },
        
       foo: {
           test: {
              value: 1
           }
       }
    }
});
```

This setup throws a `Uncaught TypeError: model.getKeypath is not a function`

I'm guessing that it's not correctly looking at the parent context for the `funcTest`?

evs-chris:
This looks a little weird, but it kinda makes sense that keypath models should hand back their value as their keypath. The other route I looked at, using `get()` from `ExpressionProxy`, causes other tangential breakage. Does this look right?

### DIFF
--- a/src/model/specials/KeypathModel.js
+++ b/src/model/specials/KeypathModel.js
@@ -12,6 +12,10 @@ export default class KeypathModel {
 		return this.value;
 	}
 
+	getKeypath () {
+		return this.value;
+	}
+
 	handleChange () {
 		this.value = this.parent.getKeypath();
 		this.dependants.forEach( handleChange );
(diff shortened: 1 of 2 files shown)


Rich-Harris:
Yeah, I think this is correct. Good catch :+1:

Issue #1981, opened 2015-06-01
Unshift is broken on data rendered with components


pangolingo:
Wen using a component, using `unshift` (and `splice`) on the data array behaves incorrectly. It removes the element from the end of the array and adds two new elements to the beginning of the array. `push` still works as expected

This occurs on both edge and latest.

Here's a demo: http://jsfiddle.net/zeuk7j17/

Madgvox:
Looks like there's a bug with implicit bindings across component boundaries. As a workaround, if you bind the value explicitly, the bug does not occur: http://jsfiddle.net/zeuk7j17/1/

pangolingo:
Thanks for figuring out a workaround. If I were to try to figure out a bugfix, do you have any suggestions for where to start looking?

lazylester:
seems to be related... not sure... even the suggested workaround (thanks for that @Madgvox ) doesn't work when using computed attributes in the component. Since I'm very new to ractive, it's possible I'm doing something stupid too. See it here (adapted from the workaround by @Madgvox ) http://jsfiddle.net/ac00dnmb/ . 
I didn't even get to the point of having the computed attributes depend on attributes from the parent. In my example I'm just returning a fixed value.

Madgvox:
You put the computed property on the component, then attempted to pass it _into_ the component from the parent. Moving the computed value definition from the component to the main ractive fixes your issue: http://jsfiddle.net/ac00dnmb/1/

lazylester:
Yes, your fix works @Madgvox, thank you. Putting the computed method in the component is suggested in the documentation http://docs.ractivejs.org/latest/computed-properties, last example on the page. Also, the component seems to be the appropriate location (similar to Backbone collections and models), since it applies to the member and not the list.
Nevertheless, it works, can't beat that! Thanks again @Madgvox. Confess I don't understand your workaround.

Madgvox:
It appears I mislead you with my example; it seems I solved a different problem than the one you were trying to fix. Can you give me a summary of what you're attempting to do, so I can point you in the correct direction?

lazylester:
The last example on the docs page for computed properties (http://docs.ractivejs.org/latest/computed-properties) is what I'm trying to do. I'm not seeing how to apply your workaround to that example.
Here's a jsfiddle for clarification: http://jsfiddle.net/Nightingill/brpbu630/2/  it exhibits the bug which is the subject of this thread. How would one apply the workaround?
Many many thanks for your help @Madgvox

[...]

PR #2170, opened 2015-09-26
clear mappings on component rebind to avoid shuffling issues - fixes #2147

### LINKED ISSUE #2147: Splice behaves unexpectedly
Seems splice is over-writing values which shouldn't be touched.

http://jsbin.com/talequ/8/edit?html,js,output
http://jsfiddle.net/va6jU/106/

You see that 'push' behaves as expected, but when you start splicing, values are changed, re-ordered and overwritten.

Try pushing once, the splicing once.

You will see the push adds a value to the end of the list,
then the splice changes the last item (a push) to a value of splice,
then moves the last item (formerly a push, now a splice) to the beginning,
adds a splice at the beginning.

(I know that is a bit convoluted).

If you click splice multiple times, the original 'test' values are over-written.

If you click splice until all 'test's are gone, then click push a few times, then splice again, the 'test's return.

evs-chris:
The issue here is that the component gets an implicit mapping to a particular model nested on its parent and, on rebind, it finds the old implicit mapping to the now wrong model and rebinds to it. This just empties the mappings on rebind to let the re-resolution happen naturally. This is another one for which I'm not positive that this is the best approach. Is there a better way to handle it?

Thanks to @ISNIT0 and @heavyk for test cases and taking a swing at this. It was deep and interesting spelunk through the internals of shuffling.

### DIFF
--- a/src/view/items/Component.js
+++ b/src/view/items/Component.js
@@ -216,6 +216,7 @@ export default class Component extends Item {
 
 		// update relevant mappings
 		const viewmodel = this.instance.viewmodel;
+		viewmodel.mappings = {};
 
 		if ( this.template.a ) {
 			Object.keys( this.template.a ).forEach( localKey => {
--- a/test/browser-tests/components/data-and-mappings.js
+++ b/test/browser-tests/components/data-and-mappings.js
@@ -1061,3 +1061,39 @@ test( 'Bindings, mappings, and upstream computations should not cause infinite m
 
 	t.htmlEqual( fixture.innerHTML, '{"bar":""}<input />' );
 });
+
+test( 'components should update their mappings on rebind to prevent weirdness with shuffling (#2147)', t => {
+	const Item = Ractive.extend({
+		template: '{{value}}'
+	});
+
+	const ractive = new Ractive({
+		el:fixture,
+		template: `
+			<div>--23--</div>
+			<div id="s1">{{#s1}}<Item />{{/}}</div>
+			<div>--13--</div>
+			<div id="s2">{{#s2}}<Item />{{/}}</div>
+			<div>--12--</div>
+			<div id="s3">{{#s3}}<Item />{{/}}</div>
+		`,
+		components: { Item },
+	});
+
+	const items = [ { value: 1 }, { value: 2 }, { value: 3 } ];
+
+	ractive.set('s1', items.slice() );
+	ractive.splice( 's1', 0, 1 );
+	t.deepEqual( ractive.get( 's1' ), [ { value: 2 }, { value: 3 } ] );
+	t.htmlEqual( ractive.find( '#s1' ).innerHTML, '23' );
+
+	ractive.set('s2', items.slice() );
+	ractive.splice( 's2', 1, 1 );
+	t.deepEqual( ractive.get( 's2' ), [ { value: 1 }, { value: 3 } ] );
+	t.htmlEqual( ractive.find( '#s2' ).innerHTML, '13' );
+
+	ractive.set('s3', items.slice() );
+	ractive.splice( 's3', 2, 1 );
+	t.deepEqual( ractive.get( 's3' ), [ { value: 1 }, { value: 2 } ] );
+	t.htmlEqual( ractive.find( '#s3' ).innerHTML, '12' );
+});


evs-chris:
It looks like this would also fix #2065 and #1981.

heavyk:
no way, that was it?!? wow... gonna test this tonight

heavyk:
dude, it totally works!! big virtual hug. :+1: 

there are a few more things not working in edge still. investigating.

heavyk:
quick question. how would I implement this in 0.7.3 ??

mappings are very different now and so is the `unbind` method on component.

evs-chris:
I'm not sure what the odds of an 0.7.4 release are, but I think this would cover this particular bug when applied to master:

```diff
diff --git a/src/virtualdom/items/Component/prototype/rebind.js b/src/virtualdom/items/Component/prototype/rebind.js
index cb9b868..949c1e9 100644
--- a/src/virtualdom/items/Component/prototype/rebind.js
+++ b/src/virtualdom/items/Component/prototype/rebind.js
@@ -1,6 +1,17 @@
 export default function Component$rebind ( oldKeypath, newKeypath ) {
        var query;

+       const mappings = this.instance.viewmodel.mappings;
+       let m;
+       Object.keys( mappings ).forEach( k => {
+               m = mappings[k];
+               if ( m.keypath.startsWith( oldKeypath ) ) {
+                       m.unbind( true );
+                       m.keypath = m.keypath.replace( oldKeypath, newKeypath );
+                       m.setup();
+               }
+       });
+
        this.resolvers.forEach( rebind );

        for ( let k in this.yielders ) {
diff --git a/test/__tests/components.js b/test/__tests/components.js
index 516c5f6..93dcfac 100644
--- a/test/__tests/components.js
+++ b/test/__tests/components.js
@@ -895,4 +895,40 @@ test( 'Unresolved keypath can be safely torn down', t => {

        ractive.set('show', false);

-});
\ No newline at end of file
+});
+
+test( 'components should update their mappings on rebind to prevent weirdness with shuffling (#2147)', t => {
+       const Item = Ractive.extend({
+               template: '{{value}}'
+       });
+
+       const ractive = new Ractive({
+               el:fixture,
+               template: `
+                       <div>--23--</div>
+                       <div id="s1">{{#s1}}<Item />{{/}}</div>
+                       <div>--13--</div>
+                       <div id="s2">{{#s2}}<Item />{{/}}</div>
+                       <div>--12--</div>
+                       <div id="s3">{{#s3}}<Item />{{/}}</div>
+               `,
+               components: { Item },
+       });
+
+       const items = [ { value: 1 }, { value: 2 }, { value: 3 } ];
+
+       ractive.set('s1', items.slice() );
+       ractive.splice( 's1', 0, 1 );
+       t.deepEqual( ractive.get( 's1' ), [ { value: 2 }, { value: 3 } ] );
+       t.htmlEqual( ractive.find( '#s1' ).innerHTML, '23' );
+
+       ractive.set('s2', items.slice() );
+       ractive.splice( 's2', 1, 1 );
+       t.deepEqual( ractive.get( 's2' ), [ { value: 1 }, { value: 3 } ] );
+       t.htmlEqual( ractive.find( '#s2' ).innerHTML, '13' );
+
+       ractive.set('s3', items.slice() );
+       ractive.splice( 's3', 2, 1 );
+       t.deepEqual( ractive.get( 's3' ), [ { value: 1 }, { value: 2 } ] );
+       t.htmlEqual( ractive.find( '#s3' ).innerHTML, '12' );
+});
```

heavyk:
you're the best!

martypdx:
This looks like right thing to do given the circumstances, but part me winces at a push at the end of an array bound to a list of components triggering a resolution of all component arguments.

heavyk:
@martypdx - I'm not sure I understand. the PR or the patch?

evs-chris:
Unless I misunderstand (not unlikely) push and pop would not be affected. Only things that change the length of the array before the end would actually trigger the remapping in rebind, and that was already happening for explicit mappings. This basically just makes sure implicit mappings rebind too.

heavyk:
ah, I see.
well, if it works - great. later, optimisze it... right?

PR #2172, opened 2015-09-26
Symlinks for keypaths - RFC

evs-chris:
This is a port of #1722 to the lovely new Model replacement for viewmodel and friends. It basically allows you to create a link between two keypaths that makes them equivalent, which is particularly handy for master-detail type scenarios. For instance:

```html
<ul>
  {{#each items}}
    <li on-click="link(@keypath, 'current')">{{.name}} {{.number}}</li>
  {{/each}}
</ul>
<button on-click="push('items', {})">Add</button>
<div>
  Name: <input value="{{ current.name }}" /><br/>
  Number: <input value="{{ current.number }}" />
</div>
```

This passes the tests that were added with #1722, but there probably need to be a few more. I would greatly appreciate any feedback on concept, construction, or further tests.

### DIFF
--- a/src/Ractive/construct.js
+++ b/src/Ractive/construct.js
@@ -142,6 +142,9 @@ function initialiseProperties ( ractive ) {
 	// observers
 	ractive._observers = [];
 
+	// links
+	ractive._links = {};
+
 	if(!ractive.component){
 		ractive.root = ractive;
 		ractive.parent = ractive.container = null; // TODO container still applicable?
--- a/src/Ractive/prototype.js
+++ b/src/Ractive/prototype.js
@@ -10,6 +10,7 @@ import findParent from './prototype/findParent';
 import fire from './prototype/fire';
 import get from './prototype/get';
 import insert from './prototype/insert';
+import link from './prototype/link';
 import merge from './prototype/merge';
 import observe from './prototype/observe';
 import observeList from './prototype/observeList';
@@ -32,6 +33,7 @@ import subtract from './prototype/subtract';
 import teardown from './prototype/teardown';
 import toggle from './prototype/toggle';
 import toHTML from './prototype/toHTML';
+import unlink from './prototype/unlink';
 import unrender from './prototype/unrender';
 import unshift from './prototype/unshift';
 import update from './prototype/update';
@@ -50,6 +52,7 @@ export default {
 	fire,
 	get,
 	insert,
+	link,
 	merge,
 	observe,
 	observeList,
@@ -75,6 +78,7 @@ export default {
 	toggle,
 	toHTML,
 	toHtml: toHTML,
+	unlink,
 	unrender,
 	unshift,
 	update,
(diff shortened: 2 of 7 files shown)


evs-chris:
This now allows links in components to establish an implicit mapping to the source. The destination is always local to the component though, so it shouldn't make for any wackiness upstream unless the destination is mapped explicitly. Even then, it shouldn't cause any problems unless the link ends up as a cyclical reference with a mapping and multiple links.

As it stands, a link restores whatever it replaced when it is unlinked, so the link more or less shadows the old value. Would that be preferable to setting `undefined` in the linked model, as the old PR did? I think that, especially in the context of components with a destination mapping (for whatever reason), shadowing makes the most sense. But my sense has been known to not make sense, if that makes sense.

heavyk:
that'd be especially useful for timelines ... what a great idea

martypdx:
Nice :)

If I understand correctly, `link` is like `set` with two-way binding semantics.  Conceptually mostly likely used for state vs data properties?

evs-chris:
That's the goal. Basically, writing one side of the link is the same as writing the other, so yeah two way binding for arbitrary keypaths. It got a lot easier with models.

And yes, it's mostly a state thing to me. Primarily to make it easier to have a swappable piece of model in focus without resorting to getting weird with passing keypath strings to data functions and manually `update`ing.

heavyk:
I have a usecase where this would be especially effective. (I think)

https://github.com/heavyk/affinaty/blob/develop/src/lib/decorators/markdown.js#L41-L45

what that decorator does, is it takes a keypath, and then transforms that content (it has markdown) into html. I then stick the innerHTML into the div (I know there's no sanity check... it's on the TODO)

but, here's the problem, when I use that in the comment list, it's inside of an each statement, so when I slice comments into the front, the keypath changes (which is why I can't use the observe in the decorator because the keypath changes out from underneath it).

https://github.com/heavyk/affinaty/blob/develop/src/partials/comment-list.html#L58

how could I do something where I can observe an item in an array, when that item can change its position in the array?

evs-chris:
@heavyk I'm not sure I follow... are you editing a particular comment or just rendering them into a lit (that may get updated over the life of the view) from markdown? If it's the former, this would definitely help. You could link the current comment to a special path to edit it and everything would stay in sync. If it's the latter, I've had good results from just adding marked to my data and calling it with the data in a triple e.g. `{{{ ~/libs.marked(.someMarkdownString) }}}`. This also has the benefit of playing nicely with splice, though a decorator should too. If you pass the actual text ref to the decorator instead of the keypath, it will automatically get updates (well, re-rendered) as the model changes.

```html
<div class="comment-body" decorator="markdown:{{.text}}">
```

heavyk:
hi @evs-chris, it's both :)

the list does get updated and and I need to add editing to a particular comment (the ones that are yours).

three braces is an option I hadn't really thought of either. I elected to do it with a decorator for embedded videos. since there's no easy way to do it with css, they need their widths set to the parent element's width... (I know, it sounds like `width: 100%` but it's not that easy.)

but, your other idea of passing the ref directly, I think is my exact solution... gonna try it now. thanks!!

and I think I just got an idea on how to workaround the edits without linking too with the same concept... !!! :)

Rich-Harris:
:metal:

PR #2173, opened 2015-09-27
Let partial expressions resolve to an actual template - RFC

evs-chris:
This is something that has seemed to me to be something of an arbitrary restriction for a while now. If you can change which partial is rendered with an expression that resolves to a (string) name, why not let it also resolve directly to a template object as well? This adds support for just that in two slightly different ways.

First, and most useful to me, if a partial expressions resolves to an object with a template property that is a string, the string is parsed and used as the template. This is particularly useful for self-mutating UI apps. A step further for ractiveception, if you will.

The second is basically the same, but the value of the expression should be the result of a call to `Ractive.parse`.

So, for reference, my use-case here is to replace the slightly hacky method of adding a specially named partial to the ractive instance at runtime and then setting the expression reference to resolve to the new special name.

Here's the contrived example:

```html
<textarea lazy value="{{foo.template}}"></textarea>
<div>{{>foo}}</div>
```
```js
new Ractive({
  el: 'body',
  template: '#up-there',
  data: {
    foo: { template: 'oh look, is this also my template?<br/>{{ foo.template }}<br/>{{bar}}' },
    bar: 'yep' 
  }
});
```


heavyk:
this would make modals easier.

I do some funky stuff to emulate that now. +1 (even though I already figured out a workaround)

martypdx:
So order of resolution is:

1. named partial
2. expression that resolves to a reference that then matches a named partial
3. parse as string template

What is behavior if runtime only (no parser)?

evs-chris:
No, if the resolution is a string, it will never try to parse. It only parses if the resolution is an object like `{ template: 'this will be parsed' }`. I was trying to make it very unlikely to accidentally parse a partial.

I hadn't considered the case of runtime only. I suppose it would throw... What happens with string partials in runtime only environments?

martypdx:
> It only parses if the resolution is an object like { template: 'this will be parsed' }

I see, but it's not the template spec, it's just using a object literal to clarify the parameter?

Meaning:

`{{ > { template: '<h1>All about Dynamism in Ractive {{plugin}}<h1>'} { plugin: 'Partials'} }}`

( As opposed to:

```js
Ractive.components.component = Ractive.extend({
    template: '{{>content}}'
});
```

```html
<component plugin='Components'>
    <h1>All about Dynamism in Ractive {{plugin}}<h1>
</component>
```

-- haha )

> I hadn't considered the case of runtime only. I suppose it would throw... What happens with string partials in runtime only environments?

IDK :blush:

evs-chris:
It will also take a pre-parsed template e.g. `{ t: [ ... ] }` and use the template array directly as the partial template.

heavyk:
oh dude I didn't really grasp the greatness of this idea at first.

imagine custom interfaces and injectable templates with this.
the entire site could be "modded" by the user.
people could create their own visualizations and stuff too.

simple, but seriously what a great idea!

JonDum:
> I hadn't considered the case of runtime only. I suppose it would throw... What happens with string partials in runtime only environments?
> IDK :blush:

It throws a `Uncaught TypeError: parse is not a function`. I think that's fine though. If you are pre-parsing your templates for performance and using ractive.runtime.js you're probably aware of why this error would happen.

Issue #2166, opened 2015-09-25
Can we discuss a better way to specify data context for component à la partials :)


JonDum:
I've yet to have a ractive project where I don't end up creating a loop of components like this:

```
 {{#each stuff}}
    <thing ../>
 {{/each}}
```

Where the  the `<thing/>` component needs to have several to dozens of properties set. It's asinine to map all of them as html `attribute="value"` pairs, _especially_ when they can change quite frequently when rapidly prototyping. 

Currently, I get around this by doing

```
 {{#each stuff}}
    <thing model="{{this}}"/>
 {{/each}}
```

and

```
thing = Ractive.extend({ 
  oninit: function() {

      /* Copy the model to the data  :|  :|  :|   */
      var model = this.get('model');
      this.set(model);

  }
})
```

But this workaround has been a super pain in the ass and it smells terrible. You get all sorts of oddities with two way binding and you have to be super careful not to overwrite properties, or even return new computed objects via slice or something—as any of that will break two way binding. It gets even more hairy when you realize that some primitives are always passed by value whereas objects aren't, so you have to wrap things in `{value: "foo"}` or instead revert to `<div>{{model.foo}}</div>` in your template.

What makes me most sad is that partials sort of already have this figured out and it makes components look second-class:

```
{{#stuff}}
    {{> thing someData}}
{{/}}
```

Would it be hard to do something like?

```
{{#stuff}}
    <thing someData/>
{{/}}
```

Madgvox:
Perhaps a reserved `default` attribute that acts like `Ractive.defaults.data` for the component instance.

```html
{{#stuff}}
  <thing default="{{ someData }}" />
{{/}}

martypdx:
Why not just reference the data implicitly and not pass anything? You can always pin down the parameters later...

martypdx:
Also, see [this conversation](https://groups.google.com/forum/#!searchin/ractive-js/data$20components/ractive-js/5z1b-t1CsN0/pLRYRDLvNJkJ)

Madgvox:
What @martypdx said. I forgot to mention the implicit data part

JonDum:
@martypdx @Madgvox By 'implicit' do you mean accessing the parent's/roots properties directly as if they were on the component's viewmodel?

Madgvox:
@JonDum yep, if you have `isolated:false` on the component, you can access the parent context.

JonDum:
[...]

PR #2174, opened 2015-09-27
Allow components to get a special alternate context via 'this' mapping - RFC

evs-chris:
Per @Rich-Harris' plan in #2166, I was wondering what this would entail, and basic support appears to be super simple.

Should components with an alternate context have `this` resolution applied to `get`, `set`, and other non-template contexts? My gut says yes, but I could see where you may want to use `'this.whatever'` so that you don't have an overly broad external context interfere with internal properties. I'm on the fence. I guess it depends on how we expect component authors to use this, so please, component authors, throw in your two `LC_LITTLEST_CURRENCY`.

@martypdx suggested `@context` as the attribute name. Any opinions? I personally prefer `this` from an aesthetic standpoint, because all bike sheds should be green with blue trim.

### DIFF
--- a/src/Ractive/prototype/get.js
+++ b/src/Ractive/prototype/get.js
@@ -13,10 +13,14 @@ export default function Ractive$get ( keypath ) {
 		// if this is an inline component, we may need to create
 		// an implicit mapping
 		if ( this.component ) {
-			model = resolveReference( this.component.parentFragment, key );
+			if ( this.viewmodel.has( 'this' ) && this.viewmodel.joinKey( 'this' ).has( key ) ) {
+				keys.unshift( 'this' );
+			} else {
+				model = resolveReference( this.component.parentFragment, key );
 
-			if ( model ) {
-				this.viewmodel.map( key, model );
+				if ( model ) {
+					this.viewmodel.map( key, model );
+				}
 			}
 		}
 	}
--- a/src/Ractive/prototype/get.js
+++ b/src/Ractive/prototype/get.js
@@ -2,7 +2,21 @@ import { splitKeypath } from '../../shared/keypaths';
 import resolveReference from '../../view/resolvers/resolveReference';
 
 export default function Ractive$get ( keypath ) {
-	if ( !keypath ) return this.viewmodel.get( true );
+	if ( !keypath ) {
+		const result = this.viewmodel.get( true );
+
+		// merge in alt context if available
+		if ( this.component && result.hasOwnProperty( 'this' ) ) {
+			for ( let k in result.this ) {
+				if ( !result.hasOwnProperty( k ) ) {
+					result[k] = result.this[k];
+				}
+			}
+			delete result.this;
+		}
+
+		return result;
+	}
 
 	const keys = splitKeypath( keypath );
 	const key = keys[0];
--- a/src/Ractive/prototype/set.js
+++ b/src/Ractive/prototype/set.js
@@ -28,14 +28,22 @@ export default function Ractive$set ( keypath, value ) {
 
 
 function set ( ractive, keypath, value ) {
+	const keys = splitKeypath( keypath );
+	const model = ractive.viewmodel;
+
 	if ( typeof value === 'function' ) value = bind( value, ractive );
 
 	if ( /\*/.test( keypath ) ) {
-		ractive.viewmodel.findMatches( splitKeypath( keypath ) ).forEach( model => {
+		model.findMatches( keys ).forEach( model => {
 			model.set( value );
 		});
 	} else {
-		const model = ractive.viewmodel.joinAll( splitKeypath( keypath ) );
-		model.set( value );
+		// prefer component alternate context to data
+		if ( ractive.component && !model.has( keys[0] ) && model.has( 'this' ) ) {
+			keys.unshift( 'this' );
+			ractive.viewmodel.joinAll( keys ).set( value );
+		} else {
+			ractive.viewmodel.joinAll( keys ).set( value );
+		}
 	}
 }
--- a/src/view/items/shared/EventDirective.js
+++ b/src/view/items/shared/EventDirective.js
@@ -135,6 +135,17 @@ export default class EventDirective {
 			event.index = this.parentFragment.indexRefs;
 
 			if ( passedArgs ) passedArgs.unshift( event );
+
+			// check to see if we need to adjust the keypath
+			if ( this.ractive.component ) {
+				const mappings = this.ractive.viewmodel.mappings;
+				for ( let k in mappings ) {
+					if ( event.keypath.indexOf( mappings[k].getKeypath() ) === 0 ) {
+						event.keypath = event.keypath.replace( mappings[k].getKeypath(), k );
+						break;
+					}
+				}
+			}
 		}
 
 		if ( this.method ) {
--- a/src/view/items/shared/EventDirective.js
+++ b/src/view/items/shared/EventDirective.js
@@ -142,6 +142,9 @@ export default class EventDirective {
 				for ( let k in mappings ) {
 					if ( event.keypath.indexOf( mappings[k].getKeypath() ) === 0 ) {
 						event.keypath = event.keypath.replace( mappings[k].getKeypath(), k );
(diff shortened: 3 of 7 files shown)


heavyk:
you're on fire dude! +1

totally should have the alt context for `get` `set` .. it makes no sense otherwise... I wouldn't want to get the template above me's data.

martypdx:
> @martypdx suggested @context as the attribute name. Any opinions? I personally prefer this from an aesthetic standpoint, because all bike sheds should be green with blue trim.

Stick with `this` :+1: 

> Should components with an alternate context have this resolution applied to get, set, and other non-template contexts? My gut says yes, but I could see where you may want to use 'this.whatever' so that you don't have an overly broad external context interfere with internal properties. I'm on the fence.

IMO should be consistent across template resolution and API, so yes. Force resolution should happen against supplied `this` context`, not root. In other words anything not explicitly defined on data happens against `this` context.

evs-chris:
The last commit adjusts the API methods to use the alternate context too. Two questions have popped up:

1. On `this.get()` in a component with an alternate context, should the `this` mapping be returned? Or is the current behavior of returning the component root model correct?
2. I was expecting keypaths in events fired from components to have the keypath relative to the component, but as it stands currently, they are relative to the root instance. Is that

evs-chris:
Is it just me or does this look like it's not far off from a lightweight component? It still has its own viewmodel, but barely. Would it make more sense to do this as a lightweight component?

heavyk:
> Would it make more sense to do this as a lightweight component?

yes. because I think these will essentially be used only in loops where the `this` is the data
 they could also be thought of as specific components.

heavyk:
> should the this mapping be returned? 

I'm not sure I understand right. do you mean returning the `this` and also the value?

evs-chris:
When `get()` (no args or empty string) is called on a component, the component will more or less glue together its data object and mappings and return that. I was wondering if that should continue to be the case. The alternative would be to basically convert `get()` into `get('this')` if an alternative context is provided. As it stands now, the object returned is the data along with the mappings, including the `this` mapping.

heavyk:
right. I think it should be in the context of `this`. always, if I need the parent's data, I put `this.parent.get()`

martypdx:
> Is it just me or does this look like it's not far off from a lightweight component? It still has its own viewmodel, but barely. Would it make more sense to do this as a lightweight component?

Components needs their own state, so it needs a viewmodel. 

> When get() (no args or empty string) is called on a component, the component will more or less glue together its data object and mappings and return that. I was wondering if that should continue to be the case. The alternative would be to basically convert get() into get('this') if an alternative context is provided. 

I would leave it as-is given that we don't delineate state versus props. If we did, this and previous answer might be different...

> As it stands now, the object returned is the data along with the mappings, including the this mapping.

So if I had this:

```html
<timer this='{{time}}' showSeconds='true' mode='analog'/>
```

And assuming time was `{ hours: 10, minutes: 1, seconds: 3 }`, I would expect `console.log` in:

```js
Timer = Ractive.extend({
     data: {
        showHours: true,
        showMinutes: true,
        showSeconds: false,
        mode: 'digital'
     },
    oncomplete () {
        console.log( this.get() );
    }
}
```

would log:

```js
{ hours: 10, minutes: 1, seconds: 3, showHours: true, showMinutes: true, showSeconds: true, mode: 'analog' }
```

JonDum:
Couple more tests, not sure if useful:


```js
test( 'Alt context should work in a loop', t => {

	const Foo = Ractive.extend({
		template: '',
        data() {
            dogGoes: 'woof'
        }
	});

	const r = new Ractive({
		el: fixture,
		template: `
            {{#things}}
                <Foo this='{{this}}'/>
            {{/}}
        `,
		components: { Foo },
		data: {
			things: [ 
                { bar: true, zip: true},
                { qux: true, zap: true },
            ]
		}
	});

    const comps = r.findAllComponents();

    // ensure model is being passed
    t.equal( comps[0].get('bar'), true);
    t.equal( comps[0].get('zip'), true);

    t.equal( comps[1].get('qux'), true);
    t.equal( comps[1].get('zap'), true);

    // and ensure other object's properties in `things` aren't being shared
    t.strictEqual( comps[0].get('qux'), undefined);
    t.strictEqual( comps[0].get('zap'), undefined);
    t.equal( comps[0].get('dogGoes'), 'woof');

    t.strictEqual( comps[1].get('bar'), undefined);
    t.strictEqual( comps[1].get('zip'), undefined);
    t.equal( comps[1].get('dogGoes'), 'woof');



});

test( 'Components with alt context should not leak properties', t => {

	const Foo = Ractive.extend({
		template: `{{bar}}`,
		data() { return { bar: true }; },
        init() {
            this.set( 'qux' , true );
        }
	});

	const r = new Ractive({
		el: fixture,
		template: `
            {{#things}}
                <Foo this='{{this}}'/>
            {{/}}
        `,
		components: { Foo },
		data: {
			things: [ 
                { bar: false, zip: true },
                { bar: true, bip: false },
            ]
		}
	});

    t.strictEqual( r.get('things.0.qux'), undefined);
    t.strictEqual( r.get('things.1.qux'), undefined);

});
```

evs-chris:
@JonDum the leaky context test won't pass because if there is an alternate context set, any properties that aren't explicitly mapped or in the component data at init will go into the alt context.

@martypdx I updated `get` to merge in alternate context if there is one. As it was, you would have had to use `result.this.whatever` to get your `whatever` property off of it. Do you have any insight into event keypaths not being component relative for mappings? I think they should probably be remapped by the EventDirective before it adds them to the event. Does that sound right?

martypdx:
> Do you have any insight into event keypaths not being component relative for mappings? I think they should probably be remapped by the EventDirective before it adds them to the event. Does that sound right?

Yes, I think that's right - I just don't think the EventDirective is currently doing any of the needed work to set the event context, keypath, and index based on the component (versus event origin)

Rich-Harris:
Regretfully I think we should park this for the moment in light of some of the nuanced discussion on #2166 – it's probably better to restrict ourselves to fixing bugs between now and 0.8 rather than changing behaviour in ways that aren't universally agreed upon, and revisit it once the issues list is cleaner (ha!). How does that sit with everyone? Just want to make sure that all the valid points from @thomsbg and @JonDum don't get glossed over.

evs-chris:
Absolutely! I just wanted to see what it would take to implement and how it would play out in practice.

FWIW, after a bit of thought and playing, I actually like this more now. The ability to bind all references to a target object works really well for widgety components that focus on one thing ('edit this contact', 'check out these details', and the like). I do appreciate that adding one more resolution path is not necessarily ideal for cognitive load, but this particular path seems like it would make isolation a bit easier to deal with. I think a 'no-ambiguity' mode would go further for helping large projects.

martypdx:
I agree with @evs-chris about it's usefulness.

I think the broader discussion is about state vs props, not really this feature. 

IMO  I think that this feature is brilliant because it just works like you expect it to _without_ really having to think about it.

JonDum:
I'm now getting a `undefined` exception in `ArrayWrapper` [here](https://github.com/ractivejs/ractive/blob/dev/src/Ractive/static/adaptors/array/index.js#L57). It's something to do with `ractive-select` setting its `_items` internally from a regular non-ractified array [here](https://github.com/JonDum/ractive-select/blob/master/src/ractive-select.js#L177).

I added a simple `if(!storage) return` in `ArrayWraper.teardown()` and that made the error go away, but I'm not sure if that has any other repercussions. 

Looks like it's coming from here in `Model.applyValue()`

```js
var shouldTeardown = !this.wrapper.reset || this.wrapper.reset(value) === false;

if (shouldTeardown) {
    this.wrapper.teardown(); // should this check if the wrapper is actually ractified?
    this.wrapper = null;
    this.parent.value[this.key] = this.value = value;
    this.adapt();
} else {
    this.value = this.wrapper.get();
}
```

evs-chris:
I'm going to close this for now because it's not likely to hit anytime real soon, dev has wandered a good ways away, there are some bigger things I'd like to land first, and I have an idea about how to address this that would also allow iterative yields. I'll leave the branch for future reference.

PR #2177, opened 2015-09-29
Plain text parsing mode - RFC

evs-chris:
I have a need to process a template as plain text i.e. without elements and entity escaping. I have a nasty regex pre-processor that mangles things that regular parsing normally consumes and a post-processor that puts them back, but it's not exactly a foolproof solution.

So, I was wondering if anyone would be opposed to adding a parser flag that causes it to skip processing for elements and entity refs. This would let Ractive library consumers effectively use Ractive as a complete replacement for non-virtualdom-based mustache implementations.


heavyk:
+1 +1 +1 - I ran into that the other day. I was actually trying to use ractive as a template for some js code and things got complicated because it wasn't xml.

totally awesome!!

JonDum:
> I was actually trying to use ractive as a template for some js code and things got complicated

My god, you guys are crazy sons of bitches :smiling_imp: :smirk_cat:

evs-chris:
@ractivejs/team-ractive Is anyone opposed to this getting merged?

fskreuz:
I think more tests, especially the following cases where the mustache interferes with the tag:

```
<element {{#if bool }}attr="value"{{/if}}>
<element attr="{{#if bool }}value{{/if}}>
<element {{# bool }}attr="value"{{/}}>
<element attr="{{# bool }}value{{/}}>
<element attr="{{# arr }}{{.}}{{/}}>
<element attr="{{#each arr }}{{.}}{{/}}>
{{#if bool }}<element>{{/if}}
{{# bool }}<element>{{/if}}
{{#each arr }}<element>{{/}}
// and more
```

evs-chris:
I wasn't really clear about this in the initial posting, but the only way to trigger this behavior is to call `Ractive.parse("some string", { textOnlyMode: true })` directly. You can then use the resulting template directly with the Ractive constructor or `Ractive.extend` or by assigning to a partial name. Elements and components (and entities) are completely ignored in this mode, so sections cannot interfere with elements, as there are none - only text, and there is no way get a component in which to `{{yield}}` or use `{{>content}}`.

I completely understand if no one wants to add this because it serves a pretty esoteric use-case. But it is super useful if you ever need to use Ractive's flavor of mustache for pre-processing html or js. In my case, it's a CMS where the pages are compiled from "components" that are themselves bits of Ractive template with various settings attached. Letting the designers use the same template language for both components and the resulting pages has been great for learning curve except for the issues like expressions with `<` and disappearing directives.

heavyk:
I totally want it! I have not had the time yet to test it though. however, I will test it tomorrow since you're wanting to merge. (unless you think it's already good enough to merge ;)

fskreuz:
Appears to be useful in server-side rendering. Parsing just the mustaches, it looks like Ractive could skip a ton of work by not minding the tags, and just pull off a Mustache on the entire operation. Looks to me like a potential performance workaround.

Rich-Harris:
Can't say it's something I've ever had a need for but I believe you all when you say it's useful!

evs-chris:
Sweet! This will save me tons of headache.

PR #2179, opened 2015-09-29
Section type can change between updates. Always update the correct value.

### LINKED ISSUE #2178: [BUG] A section can't change it's sectionType if it's initialised with null
Link: http://jsbin.com/kifaqehoho/2/edit?html,js,output

I have a simple array that loops and renders into a div, however if I initialise it with a null then it never updates. Infact it updates with a wrong value.

NikhilVerma:
Fixes #2178


unwitting:
+1, this can cause some pretty invisible problems if you've constructed a template without specifying a prop you'll later use for display.

heavyk:
+1 also.. I have noted some interesting behaviour with nulls also at times. dunno if this will fix them but sounds good to me.

NikhilVerma:
Yeah I don't mind either of the two. #2180 might help with some performance testing though. This one just solves a specific issue I had.

Rich-Harris:
Thanks for this @NikhilVerma – I merged #2180, does that mean this PR is no longer needed?

NikhilVerma:
@Rich-Harris I think having the extra null check is still going to be a bit faster in cases where it's going to be initialised with a null. But I don't see that as an issue. So I'll close this PR.

PR #2180, opened 2015-09-29
Allow untyped sections to change types on the fly - alt for #2178

evs-chris:
If the template doesn't have a specific section type, this will let it flip on the fly. This should only affect plain mustache sections.

I also cheated a bit on `bind`, since there was a DRY TODO on `update`. With this `bind` is slightly more expensive if there's a model as it now just delegates to `update`, but its code got way shorter.


evs-chris:
Forgot to mention #2178...

PR #2182, opened 2015-09-29
ES2015 object literal property shorthand

evs-chris:
`{ foo, bar, baz: 'bat' }` because I'm used to doing it everywhere else now. Opinions?

### DIFF
--- a/src/parse/converters/expressions/primary/literal/objectLiteral/keyValuePair.js
+++ b/src/parse/converters/expressions/primary/literal/objectLiteral/keyValuePair.js
@@ -1,4 +1,4 @@
-import { KEY_VALUE_PAIR } from '../../../../../../config/types';
+import { KEY_VALUE_PAIR, REFERENCE } from '../../../../../../config/types';
 import readKey from '../../../shared/readKey';
 import readExpression from '../../../../readExpression';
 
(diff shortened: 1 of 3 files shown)


src/parse/converters/expressions/primary/literal/objectLiteral/keyValuePair.js
```diff
@@ -10,6 +10,8 @@ export default function readKeyValuePair ( parser ) {
 	// allow whitespace between '{' and key
 	parser.allowWhitespace();
 
+	const refKey = parser.nextChar() !== '\'' && parser.nextChar() !== '"';
```
MartinKolarik: Should also fail with `{ 4 }`, but otherwise :+1: :+1:

evs-chris: `readKey` won't read `4` as a key, so it bails while trying to parse the key.

MartinKolarik: Why wouldn't it? `4` is a valid key. Also see https://github.com/ractivejs/ractive/blob/59686405a594eb239c7aec7954bfccaebd200f34/src/parse/converters/expressions/shared/readKey.js#L16

evs-chris: It won't if you forget to switch back to the right branch before checking :)

Thanks and fixed..

heavyk:
nice!! :+1:

martypdx:
This will be a nice addition to template writing

Rich-Harris:
YES

PR #2186, opened 2015-10-02
Name bound checkboxes with duplicate values and crossing component boundaries - #2163

### LINKED ISSUE #2163: Two-way binding only works within a single instance of component
Two-way data binding correctly only works within a single instance, not across component boundaries ([live demo](http://jsbin.com/qepupayego/1)).

Question from [stackoverflow](http://stackoverflow.com/questions/32755097/two-way-binding-doesnt-work-inside-ractive-components).

evs-chris:
Named checkbox bindings where more than one checkbox shared a value were not handled. This adds support for them. It also moves the group registry from a private object in binding group onto a property of the model so that components are guaranteed to get the same group if they share a model even if mappings or some other factor make the hash not equal. This should fix #2163.

I'm 99% sure this is correct, but I would like a sanity check on both functionality and construction.

I removed the RadioBinding TODO to move to binding group because they have no model in which to store the group.

### DIFF
--- a/src/view/items/element/binding/CheckboxNameBinding.js
+++ b/src/view/items/element/binding/CheckboxNameBinding.js
@@ -1,22 +1,17 @@
 import Binding from './Binding';
 import { isArray } from '../../../../utils/is';
-import { arrayContains } from '../../../../utils/array';
+import { arrayContains, removeFromArray } from '../../../../utils/array';
 import getBindingGroup from './getBindingGroup';
 import handleDomEvent from './handleDomEvent';
 
-function isChecked ( binding ) {
-	return binding.node.checked;
-}
-
-function getValue ( binding ) {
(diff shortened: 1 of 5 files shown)


heavyk:
awesome dude! this fixed another tiny issue for me on edge. great work

JonDum:
@evs-chris is killing it so hard lately. Do you have a git tip or something man?

martypdx:
> I removed the RadioBinding TODO to move to binding group because they have no model in which to store the group.

But the browser handles the "linkage" so it works in components, right? :D http://jsfiddle.net/x3w18j13/

evs-chris:
@martypdx It works in this branch... or did I misunderstand your question?

martypdx:
@evs-chris just did fiddle to prove it works and sort things out in my mind, thx!

Issue #2191, opened 2015-10-06
Improper detection of context chain in #resolveAmbiguousReference


NikhilVerma:
Example: http://jsbin.com/xafadaxolu/1/edit?html,js,output

So if a fragment is not a root it's not marked that it has a context chain. And then if that is followed by a context which has a prototype property. It doesn't work.

![image](https://cloud.githubusercontent.com/assets/87832/10309559/dd4349f6-6c5a-11e5-89ae-968bb0141c56.png)

My solution is to patch Model.prototype.has() but perhaps there is a better way?

```javascript
	has ( key ) {
		const value = this.get();
		if ( !value ) {
			return false;
		}

		var hasProperty = hasProp.call( value, key );
		if ( hasProperty ) {
			return true;
		}

		// We climb up the constructor chain to find if one of them contains the key
		var constructor = value.constructor;
		while ( constructor !== Function && constructor !== Array && constructor !== Object) {
			if ( hasProp.call( constructor.prototype, key ) ) {
				return true;
			}

			constructor = constructor.constructor;
		}

		return false;
	}
```

See PR: #2192

fskreuz:
Best if you open this one as a PR. Then everyone can review it, and see your progress in the form of successive commits.

NikhilVerma:
@fskreuz done!

PR #2192, opened 2015-10-06
Better detection of existing keys in models

NikhilVerma:
See issue #2191

### DIFF
--- a/src/model/Model.js
+++ b/src/model/Model.js
@@ -280,7 +280,26 @@ export default class Model {
 
 	has ( key ) {
 		const value = this.get();
-		return value && hasProp.call( value, key );
+		if ( !value ) {
+			return false;
+		}
+
+		var hasProperty = hasProp.call( value, key );
+		if ( hasProperty ) {
+			return true;
+		}
+
+		// We climb up the constructor chain to find if one of them contains the key
+		var constructor = value.constructor;
+		while ( constructor !== Function && constructor !== Array && constructor !== Object) {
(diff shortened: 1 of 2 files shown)


evs-chris:
Out of curiosity, and I have no idea the performance implications here, why not replace the `hasProp` const with
```js
function hasProp(prop) {
  return prop in this;
}
```
?

Rich-Harris:
Thanks @NikhilVerma! @evs-chris `prop in this` checks the whole prototype chain rather than own properties. In fact my initial thought was 'why don't we just use `prop in value` instead of ascending the constructor chain?', but of course that fails when you have property names that clash with e.g. `Array.prototype`, as in this scenario...

```html
{{#each items}}
  {{#each some}}...{{/each}}
{{/each}}
```

```js
items: [{ some: 'a' }, { other: 'b' }, { some: 'c' }]
```

...whereby the second iteration picks up the `items` array's `some` method, which we definitely don't want. Took me some headscratching to figure out why that test was failing when I tried `prop in value`...

evs-chris:
Hadn't considered that at all... @NikhilVerma Nice!

NikhilVerma:
Awesome :D

martypdx:
btw - We might eventually want to use `getPrototypeOf` instead of walking the constructor chain. We need it for babel classes anyway on ie8.

I tested out [this polyfill](https://github.com/es-shims/es5-shim/blob/master/es5-sham.js#L53-L81) on ie8 and seemed to work fine.

PR #2194, opened 2015-10-06
First swing at escaping . in keypaths - RFC

evs-chris:
__Probably post 0.8__

This is a fairly requested feature that I believe was present at one point and removed due to complexity. After the massive awesome `Model` changes from @martypdx and @Rich-Harris I think a lot of the complexity has been reduced. I needed a mental break from the day job, so I figured I'd take a swing at this.

All of the tests pass, and the following now works as you'd expect:
```html
{{ foo\.bar }} {{ foo.bar\.baz }} {{ foo.bar.baz }}
{{ .['foo.bar'] }} {{ foo['bar.baz'] }} {{ foo['bar']['baz'] }} {{ foo[key] }}
```
Data:
```js
{
  'foo.bar': 'works1',
  foo: { 'bar.baz': 'works2', bar: { baz: 'still works' } },
  key: 'bar.baz'
}
```
Output:
```
works1 works2 still works
works1 works2 still works works2
```

### So...

If you see anything I missed or screwed up horribly or have some more tests to apply, please feel free to mention them or push them to this branch.

### DIFF
--- a/src/Ractive/prototype/shared/makeArrayMethod.js
+++ b/src/Ractive/prototype/shared/makeArrayMethod.js
@@ -1,5 +1,5 @@
 import { isArray } from '../../../utils/is';
-import { normalise } from '../../../shared/keypaths';
+import { splitKeypath } from '../../../shared/keypaths';
 import runloop from '../../../global/runloop';
 import getNewIndices from '../../../shared/getNewIndices';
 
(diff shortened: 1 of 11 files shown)


evs-chris:
Also, I'm not particularly wild about the `foo\.bar` syntax and would normally say that we should just stick with `.['foo.bar']`. However, keypaths get returned in several different places, so there has to be an internal representation that can leak out and be fed back in. I don't think it's worth trying to twiddle out `foo['bar.baz']` from the internal representation of `['foo', 'bar.baz']` over `foo.bar\.baz`.

Rich-Harris:
:metal:

Issue #395, opened 2014-01-17
Progressive enhancement


fdecampredon:
Actually when we prerender Ractive template on server, then try to rerender them on client the dom is replaced.
Do you think it would be possible to have an options that would allow Ractive to not replace the dom but use the already built in one ?
To ensuring that the dom already present would be the same that the dom produced prehaps you could use some kind of checksum (a little alike what React guys do)

Thanks for your work anyway

PaquitoSoft:
Just wanted to add another developer who is quite interested in this feature.

kurdin:
Is there any progress on this?

Rich-Harris:
@kurdin was just about to direct you here from #957 :-) No, there hasn't been any progress, but I'm anxious for us to make a start on it soon. I envision it as being the big new feature for 0.8.x.

kurdin:
@Rich-Harris Would love to help with testing :) 

btw: I did some experiments with DOM changes after ``ractive.render()`` event and it looks like if you modify something inside ractive init element and then change the data to trigger render update, those new changes to DOM that reflects new data can be corrupted.  Is this normal? 

Here is an example: http://jsfiddle.net/msx4r0s2/1/ 
if I open dev tools (inspect element) and modify #app.h1 to remove it's content then click on radio button to trigger language change, that modified H1 will always be empty. Is it a bug or this is how ractive deal with DOM changes?

Rich-Harris:
@kurdin watch this space :)

That behaviour you're experiencing is expected - Ractive works by storing references to DOM nodes (including text nodes within an element) and updating them when necessary. By making the assumption that the DOM *hasn't* been modified by something other than Ractive, it can make those updates in a very granular fashion that saves throwing out big chunks of perfectly good DOM.

So it's best to avoid imperative DOM manipulation (unless it's done in a way that doesn't conflict with Ractive's data binding). Though in theory, the same logic that would handle progressive enhancement could theoretically allow us to force the DOM to resync with Ractive's view of the world...

NikhilVerma:
Sorry to bump but this is the very first thing the folks in my team asked when I proposed RactiveJs for server side rendering. Right now it's way easier for us to migrate to Ractive compared to React so we have compromised for a server render -> client render approach. But this would frankly be amazing.

msokk:
Still planned? Currently pondering about ways using Ractive on the backend, so this capability would be much appreciated.

hkaju:
+1 for more info on the current status of this.

[...]

PR #2196, opened 2015-10-07
First swing at progressive enhancement - WIP RFC

evs-chris:
__Probably post 0.8__

I looked at this before all of the major changes that @Rich-Harris and @martypdx made over the last few months, and it was a bit more difficult than I wanted to dive into. Well, now basic support seems to be relatively painless.

This adds an instance option `enhance` that defaults to `false`, and when it is enabled, doesn't clear out the target element before rendering a Ractive instance. Then it passes the occupants of the target element along through the render process for each item to check to see if it can use an existing node before creating its own.

* Elements shift nodes out of the passed occupants array until they find a matching tag. If a node doesn't match, it is removed from the DOM. If one does match, it is acquired.
* Text nodes check to see if the next available node is a text node, and if it is will use it and remove it from the list. If not, it will insert a new text node in front of the next node. This is necessary, as mentioned elsewhere (#395), because Ractive creates multiple adjacent text nodes for various bits of template and HTML merges adjacent text nodes (or more precisely has no way to have adjacent text nodes).
* Everything else basically passes the list of occupants through.
* After an element has rendered its fragment, any nodes remaining in its occupant list are removed from the DOM.

This does not have special handling for anything else at this point. Notably, extra attributes on elements will hang around. If the target DOM does not quite match, depending on the differences, a fair amount may be reused. However, if there is an additional tag in the target DOM that isn't in the fragment all of the nodes from that point on will be recreated because there is no allowance for lookahead. It wouldn't be too hard to add, but I'm not sure that we would necessarily want to.

### So
Comments, critiques, commits, and suggestions for concept, construction, and testing are more than welcome!


evs-chris:
Well, it seems phantom is not a fan of some of the new tests, but they pass just fine in the browser.

Also, I have not looked into browser support at all (lookin' at you IE) and just assumed that what little this touches is pretty universal as far as DOM support goes.

src/Ractive.js
```diff
@@ -59,6 +59,7 @@ defineProperties( Ractive, {
 	Promise:        { value: Promise },
 
 	// support
+	enhance:        { writable: true, value: false },
```
fskreuz: Not sure about the name. `prerendered`? Anyone?

src/Ractive/prototype/render.js
```diff
@@ -12,7 +12,9 @@ export default function Ractive$render ( target, anchor ) {
 		if ( others ) others.forEach( teardown );
 
 		// make sure we are the only occupants
-		target.innerHTML = ''; // TODO is this quicker than removeChild? Initial research inconclusive
+		if ( !this.enhance ) {
+			target.innerHTML = ''; // TODO is this quicker than removeChild? Initial research inconclusive
```
fskreuz: Doesn't this leave behind attached handlers, references etc.?

Rich-Harris:
YESSSSSS

I'm impressed at how little extra code is needed to make this work. True, we're not doing any deep crazy diffing, we're just checking that things are as they should be and discarding them if they're not, but it's still great to see.

I've taken the liberty of moving these tests into their own file, adding a bunch of new tests, and simplifying the ones we had. (Some of the tests – attributes and two-way binding, for example) don't pass yet, of course. I'd love to get those passing – going to take a quick look now – and if it works, I actually don't see any good reason not to merge this. It'll be a 'here be dragons' feature, but it can't do any harm to people not using it, so we may as well let 'er rip.

I suspect – though I haven't checked – that it's the `node.remove()` calls that are causing issues in PhantomJS. I think that's a relatively new method. We have to use `node.parentNode.removeChild(node)`. There's already a helper, I'll wire it up

src/Ractive.js
Rich-Harris: I quite like it... makes me think of Blade Runner

6lre7ol

The *real* reason I like it is because it ties it to its stated purpose. It's also nice and easy to type! But it's certainly up for debate

src/Ractive/prototype/render.js
Rich-Harris: `remove()` and `empty()` only remove jQuery handlers – if someone was storing their own reference, or had added an event listener of their own, I'm not sure there's anything we can do? (We *do* tidy up stuff added with Ractive in the line just above.)

Rich-Harris:
Too excited not to merge this – Travis is reporting a mysterious failure, but it builds fine locally. Maybe if we ignore it it'll go away.

@evs-chris I figured out the Geico reference... took me a while (and a Google search)

heavyk:
NICE!!

@Rich-Harris if you remove line 20 in `src/view/Element.js` the travis error will be fixed. `detachNode` is imported twice.

evs-chris:
@Rich-Harris cool. Thanks for fixing up the tests. I've been meaning to get round to it, but i thought we probably wouldn't get to this until after 0.8.

I was riding the sleep version of the Ballmer peak when I named the repo. Almost named the PR "you're in good hands" and was also chuckling to myself every time I read `occupants`.

Rich-Harris:
@heavyk so weird – I saw that error but there was no duplicate import in my local copy. And yet on this machine, there was... must be some kind of git curse. Anyway, fixed now, thanks!

I put a small demo up: http://jsfiddle.net/rich_harris/kxpdgo89/ Will have to try and update http://examples.ractivejs.org/ to use this (it's doing the server-side rendering part already, just not the enhancement), to put this stuff through its paces. Very excited.

Issue #1880, opened 2015-03-27
Partial doesn't accept `.` keypath


Crisfole:
See reproduction here: https://jsfiddle.net/UG7Eq/12/

I get the following (well written, and self-explanatory) error:

    Expected a property name at line 3 character 17:
        <p>{{>test .}}</p>
                   ^----

martypdx:
`{{>test this}}` works. Unless there's an expression parsing reason, I would think `.` should work as well.

Crisfole:
Oh really? That's weird. Thanks for the workaround.

Crisfole:
Here's an additional few tests that demonstrate the behavior that's working and that's not working.

https://github.com/Crisfole/ractive/tree/bug/1880

It also adds a tests for 'can a partial take a context?' which was missing from the tests I saw.

Crisfole:
It looks like `readRefinements` is rejecting expressions that look like `.`.

Crisfole:
I can't quite figure out why, but it looks like `readExpression` is trying to consume the `.` that's separated by a space from the initial expression. Is that because it's assuming that the `.` is an indexer?

Crisfole:
So, yes, `readExpression` is assuming `.` is part of the name...apparently spaces are allowed in an expression. Is there any good reason for allowing spaces in expressions? Can you add a flag to disallow spaces in an expression (much like `relaxedNames`) that you could use in `readPartial`?

Madgvox:
An `expression` in ractive is a very general term that includes things like `1 + 1`, `foo * bar( "baz" )`, and `foo.bar.baz[ qux ].do( "thing" )[ i ]`, so from a flexibility standpoint there are very good reasons to allow spaces in expressions.

Crisfole:
Ah, I'd forgotten mathematical expressions. My guess, then, is that if this bug were to be fixed it'd need to be a breaking change to the "renderPartial" syntax. If the partial name can be an arbitrary expression that's not delimited by mustaches or parentheses then you'd have to write a specialized version of "readExpression" that allowed (and didn't consume) a trailing period. "readRefinement" would then work correctly.

PR #2202, opened 2015-10-08
Relative partial context - #1880

evs-chris:
I ran into this in a late night coding session. The fix, more-or-less as suggested by @Crisfole, is to simply not allow whitespace in refinements if the relaxedNames flag is set. Any objections?

For reference in #1880, whitespace is normally allowed in refinements because all of the following are equivalent in js:
```js
const foo = { bar: 'baz' };
foo.bar;
// and
foo  .  bar;
// and
foo.
   bar;
// and
foo
  .bar;
// and
foo    ['bar'];
// and
foo
    [
       'bar'
];
```

### DIFF
--- a/src/parse/converters/expressions/shared/readRefinement.js
+++ b/src/parse/converters/expressions/shared/readRefinement.js
@@ -4,8 +4,8 @@ import { name as namePattern } from './patterns';
 import readExpression from '../../readExpression';
 
 export default function readRefinement ( parser ) {
-	// if names are relaxed, refinement is strict, meaning no space between reference and refinement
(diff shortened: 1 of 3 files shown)


JonDum:
LGTM :+1:

martypdx:
So `relaxedNames` is more strict? Maybe should be `strictNames` or `strictRefinements`?

evs-chris:
@martypdx yeah, I started to do that initially, but went the slightly lazier and less flaggy way instead. I have added an additional flag to avoid confusion.

Rich-Harris:
:+1:

PR #2205, opened 2015-10-09
fix #2204

### LINKED ISSUE #2204: Value not rendered in expected position after set()
See the attached fiddle. After two seconds setTimeout will `set()` different data and you can see the position of the key/value which should be first change from the first div to second.

http://jsfiddle.net/osobb0mv/

(Sorry... the example template is a little more convoluted than I'd prefer.. I believe it could be pared down but i was hitting some other *possible* bug when attempting to do just that.)

Rich-Harris:
**not ready for merge** just a failing test for now

### DIFF
--- a/src/view/Fragment.js
+++ b/src/view/Fragment.js
@@ -190,7 +190,15 @@ export default class Fragment {
 	}
 
 	firstNode () {
-		return this.items[0] ? this.items[0].firstNode() : this.parent.findNextNode( this.owner );
+		let node;
+		for ( let i = 0; i < this.items.length; i++ ) {
+			node = this.items[i].firstNode();
+
+			if ( node ) {
+				return node;
+			}
+		}
+		return this.parent.findNextNode( this.owner );
 	}
 
 	// TODO ideally, this would be deprecated in favour of an
--- a/test/__support/js/samples/render.js
+++ b/test/__support/js/samples/render.js
@@ -1219,6 +1219,30 @@ const renderTests = [
 		template: `{{foo[key]}}`,
 		data: { foo: { 'bar.baz': 'yep' }, key: 'bar.baz' },
 		result: 'yep'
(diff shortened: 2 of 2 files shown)


evs-chris:
I think I found a variation of this the other day when testing my most complex project with edge. I didn't have time to distill it to a testable case, so this is great! Anyways... looks like it boils down to `Fragment` assuming that if the first piece of it doesn't have a node, then none of it will. Here's a patch that checks all of the children before punting to the parent:

```diff
diff --git a/src/view/Fragment.js b/src/view/Fragment.js
index 2927a38..6bb9f38 100644
--- a/src/view/Fragment.js
+++ b/src/view/Fragment.js
@@ -190,7 +190,16 @@ export default class Fragment {
        }

        firstNode () {
-               return this.items[0] ? this.items[0].firstNode() : this.parent.findNextNode( this.owner );
+               let node;
+               for ( let i = 0; i < this.items.length; i++ ) {
+                       node = this.items[i].firstNode();
+
+                       if ( node ) {
+                               return node;
+                       }
+               }
+               return this.parent.findNextNode( this.owner );
+               //return this.items[0] ? this.items[0].firstNode() : this.parent.findNextNode( this.owner );
        }

        // TODO ideally, this would be deprecated in favour of an
```

With that, all of the tests on this branch turn green. Is there a reason to only check the first fragment bit before punting up-template?

Rich-Harris:
And @evs-chris continues an epic streak... :runner: 

Of course! I think the original intention might have been that each item was supposed to defer to its next sibling if it didn't have a node, until you reach the end of a fragment, at which point the search moves up a level. But this is much more elegant and sensible.

evs-chris:
@Rich-Harris Sorry, I snitched your PR so I could be lazy and let the usual build automation make a build  for me to try out :)

PR #2207, opened 2015-10-10
More robust . escaping

MartinKolarik:
This PR addresses the following issues:

1. Multiple escapes in a single keypath - `ractive.get('foo\\.bar\\.baz')` now works correctly (was interpreted as `ractive.get('foo\\.bar.baz')` before).

2. `{{foo['bar.baz']}}` updates on `ractive.set('foo.bar\\.baz')`

3. `$`'s in keypaths aren't incorrectly converted to `.`

4. Backslashes can also be escaped, so calling `ractive.set('foo\\\\.bar', 1)` is the same as `ractive.set({ 'foo\\': { bar: 1 } })`

### DIFF
--- a/src/Ractive/static/adaptors/magic.js
+++ b/src/Ractive/static/adaptors/magic.js
@@ -1,3 +1,5 @@
+import { escapeKey } from '../../../shared/keypaths';
+
 let magicAdaptor;
 
 try {
@@ -81,7 +83,7 @@ class MagicWrapper {
 			const originalDescriptor = Object.getOwnPropertyDescriptor( this.value, key );
 			this.originalDescriptors[ key ] = originalDescriptor;
 
-			const childKeypath = keypath ? `${keypath}.${key}` : key;
+			const childKeypath = keypath ? `${keypath}.${escapeKey( key )}` : escapeKey( key );
 
 			const descriptor = createOrWrapDescriptor( originalDescriptor, ractive, childKeypath );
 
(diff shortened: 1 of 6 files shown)


evs-chris:
1 & 2. Well dang.
3. Sorry, I completely forgot that `$` is legal in an identifier in JS.

Thanks for cleaning up the stuff I missed :)

Rich-Harris:
I've merge dev back into this branch to make it mergeable – unfortunately one of the tests is failing on `npm run build`. Not able to look into it right now, @MartinKolarik @evs-chris any ideas?

MartinKolarik:
>unfortunately one of the tests is failing on npm run build

Hmmm. All tests are passing when I run them locally.

evs-chris:
It dies on me from the console, but works fine in the browser. I'm going to try to test it directly in node and see what happens.

MartinKolarik:
>All tests are passing when I run them locally.

Actually, `npm test` works fine, but `npm run build` fails for me as well..

evs-chris:
It also works fine when run directly in node. I'm at a loss... commenting out the last part of offending test passes the build.

MartinKolarik:
Looks like it has something to do with magic mode.

evs-chris:
Sheesh... finally found it, thanks to @MartinKolarik pointing in the right magical direction.

MartinKolarik:
@evs-chris :+1:

heavyk:
you guys are on fire!

Issue #2206, opened 2015-10-10
Method call events in partial passed through a component via {{> content}} do not work


JonDum:
http://plnkr.co/edit/3k04BCLDUSnkihFs8uDs?p=preview

main template:
```
        <Foo>
            <button on-click="doStuff()">doStuff()</button>
        </Foo>
```
inside Foo's template
```
      <div class='Foo' things='{{things}}'>
         {{> content}}
      </div>
```
js:
```
        Foo = Ractive.extend({
          template: '#comp'
        })

        app = new Ractive({
            el: '#container',
            template: '#template',
            
            components: {
              Foo: Foo
            },
            
            doStuff: function() {
              
              // where's ma alert, ma?
              alert('doStuff()');
                
            }
        });
```

Results in `Uncaught Error: Attempted to call a non-existent method ("doStuff")` when clicking the button.
Looks like the method call is looking on the component instead of the parent?

evs-chris:
Yes, the `content` partial lives in the context of the component. You should be able to use `{{yield}}` to get what you're looking for.

JonDum:
`{{yield}}` isn't what I want though :(

I made a pagination component that I'm going to open source. It works like this:

```
<paginator items='{{things}}'>
    <div class='thing'>
        {{.}}
    </div>
</paginator>
```

Where whatever is inside the `<paginator>` tag is what gets looped over inside the component. So I want the user to be able to do whatever they would normally in an `{{#each}}` but instantly get pagination. If I do `{{yield}}` instead, it doesn't even work. It creates an infinite loop and throws a maximum call stack exceeded error.

evs-chris:
Ah, that's an unfortunate situation. I've had the same problem in the past, but I side-stepped the issue by mangling the content partial that was passed in to do what I wanted. _I_ don't know of any way to work around it without getting hacky, but someone else may. I think there's an outstanding issue on something like this, but it didn't jump out at me when I looked.

fskreuz:
If you go with `{{yield}}`, you could [pull out the looping outside](http://jsfiddle.net/gvhec2se/), reducing your component into a wrapper for styling. This way, the inner HTML's context is the parent and you have direct access to `doStuff`.

```
<paginator>
  {{# things }}
    <button on-click="doStuff()">{{.}}</button>
  {{/}}
</paginator>
```

[...]

Issue #2176, opened 2015-09-28
The many places for functions


matthew-dean:
Okay, this may be obvious to everyone else who uses Ractive, and not an issue, but I find remembering all the places where to put functions when making a new Ractive maddening.

If I want to have a type of evaluation function like `{{ somefunc(val) }}`, I have to do this:

``` js
ractive = new Ractive({
    el: $target[0]
    , template: $template.html()
    , data: ractiveData
});
```

...and then I think, "Wait, never mind, I can't do anything with this; I need to put it inside my data." So then I have to do this:

``` js
ractiveData.somefunc = function() {  // stuff };
ractive = new Ractive({
    el: $target[0]
    , template: $template.html()
    , data: ractiveData
});
```

(If you're asking, "Why not just expand data{} and put the function there" it's because I often have generic functions that pass in data to make a Ractive, and while the data changes, the functions do not, because the functions are a requirement of the _template_ not the data, so the functions have to be attached after the data object is received.)

Later, I decide to call a custom function within an element like:

```
<button on-click="preset({ render: [{ keypath: 'data.LAST_INVOICE',  id: 'invoice_body' }]  })" />
```

Inevitably, that's where I go wrong. I usually make the mistake of putting the function attached to the data, or if I defined the custom functions, and decide to add an evaluation function later, I usually make the mistake of putting those functions together. Instead I need to do:

``` js
ractiveData.somefunc = function() {  // stuff };
// ractiveData.preset = function() // stuff };  NOPE
ractive = new Ractive({
    el: $target[0]
    , template: $template.html()
    , data: ractiveData
    , preset: function() { // stuff }
});
```

Then, later I want to fire an event delegate, so it looks like:

``` js
ractiveData.somefunc = function() {  // stuff };
// ractiveData.preset = function() // stuff };  NOPE
ractive = new Ractive({
    el: $target[0]
    , template: $template.html()
    , data: ractiveData
    , preset: function() { // stuff }
//  , nextStep: function(e, args) { }  NOPE, HAHA SUCKER
//  , prevStep: function(e, args) { }   
});
ractive.on({ 
    , nextStep: function(e, args) { }  //and I always forget the event var being passed in here
    , prevStep: function(e, args) { }
});
```

So, in the end, I end up with this "spread out" effect.

[...]

Issue #1353, opened 2014-10-14
#with aliasing isn't bidirectional


evs-chris:
I'm not sure if this is something that can be addressed, but creating a new context with `{{#with { key: .some.path }}}<input value="{{.key.value}}" />{{/with}}` only updates from data to synthetic context and not from synthetic context to data. See http://jsfiddle.net/br7hy8hg/1/

If the value referenced in the `#with` happens to be an array or object, the data is updated in both places, but no change events are fired on the original keypaths, which means the view gets out of sync on the original keypaths.

Madgvox:
Closing as this topic has been / is being explored in many other threads (as you can see above).

PR #2209, opened 2015-10-11
Template aliasing - RFC

evs-chris:
__Probably post 0.8~~, not ready for merge, and no tests yet~~__

This is one of those things that pops up now and then as a little bit surprising: setting context with `with` isn't really bidirectional (#1353). There has been a proposed syntax to address that `{{#with foo.bar.baz as v1, lah.dee.dah as v2}}`, and this is my attempt to see what it would take and how it would work. What this does is introduce a new step in parsing sections, a new node type, and another step in ambiguous reference resolution to achieve aliasing within a branch of a template.

### Construction

The parsing step sits between block name and expression parsing. It looks for aliases in the form `[expression] as [non-path ref]` with optional additional `,` separated aliases if the section type is a `with`. If aliases are found, they are stuck on the template node in a new `z` property and the type is changed to `ALIAS` (19). Children are still parsed, but most of special stuff, like looking for `else`, is skipped.

The new node, `Alias`, is pretty simple. It only exists to resolve its own references and supply an `aliases` map to children. This is my first new node type, so I'm not positive that all of the methods are right. It seems to be well behaved in the stuff I've thrown through it though.

The resolution step is pretty simple. During the walk up the fragment tree, it just checks for fragment owners with an `aliases` property with a key matching the root of the current ref. If there is one, the model stored there is joined if possible. If not, resolution continues as usual.

Any valid reference is usable, so you can alias `@index` and other special refs or a regular keypath or an expression.

### Segue - components

Something that also pops up every now and then with components is how to let events fire in the context of the owner while the content guts stay in the context of the component (recently #2206). Obviously bubbling is one good solution, but that requires a little bit tighter integration between component-er and component-ee. Table that for a second while I segue the segue.

I was thinking that perhaps it would be useful to introduce a new special reference (yay!) `@ractive` that returned `this.ractive` when resolved. This could be useful in some scenarios, like the question about where to put functions that popped up recently (#2176). Then things like `@ractive.someFun(...)` could be possible. It might also be useful for combining method and proxy event syntax at some point in the future.

Now, what if template aliasing allowed you to override the `@ractive` special ref? The event directive could look for a `@ractive` override before using its own `ractive` to fire an event. This could allow something like:
```html
Name: <input value="{{current.name}}" /><br/>
Phone: <input value="{{current.phone}}" />
<List values="{{people}}">
  <div class="person" on-click="link(@keypath, 'current')">{{.name}} {{.phone}}</div>
</List>
```
and the List component:
```html
{{#each page(values)}}
  {{#with @ractive.parent as @ractive}}{{>content}}
{{/each}}
{{#each range(1, values.length / pageSize)}}
  <a href="#" on-click="set('page', .)">{{.}}</a>
{{/each}}
```
Here, the `link` call in the `content` partial would happen in the context of the component's parent, since it has an alias defined. The pagination link `set` calls would happen in the context of the component because there is no alias overriding the `@ractive` reference.

### So...

If you made it this far, thanks! I would love to get some feedback, as always, on concept and construction ( and testing when this grows some tests ).

### DIFF
--- a/src/config/types.js
+++ b/src/config/types.js
@@ -14,6 +14,7 @@ export const COMPONENT         = 15;
 export const YIELDER           = 16;
 export const INLINE_PARTIAL    = 17;
 export const DOCTYPE           = 18;
+export const ALIAS             = 19;
 
 export const NUMBER_LITERAL    = 20;
 export const STRING_LITERAL    = 21;
--- a/src/parse/converters/mustache/readAliases.js
+++ b/src/parse/converters/mustache/readAliases.js
@@ -1,0 +1,65 @@
+import readExpression from '../readExpression';
+import refineExpression from '../../utils/refineExpression';
+
+const legalAlias = /^(?:[a-zA-Z$_0-9]|\\\.)+(?:(?:(?:[a-zA-Z$_0-9]|\\\.)+)|(?:\[[0-9]+\]))*/;
+
+export default function readAliases( parser ) {
+	let aliases = [], alias, start = parser.pos;
+
+	parser.allowWhitespace();
+
+	alias = readAlias( parser );
+
+	if ( alias ) {
+		aliases.push( alias );
+
+		parser.allowWhitespace();
+
+		while ( parser.matchString(',') ) {
+			alias = readAlias( parser );
+
+			if ( !alias ) {
+				parser.error( 'Expected another alias.' );
+			}
+
+			aliases.push( alias );
+
+			parser.allowWhitespace();
+		}
+
+		return aliases;
+	}
+
+	parser.pos = start;
+	return null;
+}
+
+function readAlias( parser ) {
+	let expr, alias, start = parser.pos;
+
+	parser.allowWhitespace();
+
+	expr = readExpression( parser, [] );
+
+	if ( !expr ) {
+		parser.pos = start;
+		return null;
+	}
+
+	parser.allowWhitespace();
+
+	if ( !parser.matchPattern( /as/i ) ) {
+		parser.pos = start;
+		return null;
+	}
+
+	parser.allowWhitespace();
+
+	alias = parser.matchPattern( legalAlias );
+
+	if ( !alias ) {
+		parser.error( 'Expected a legal alias name.' );
+	}
+
+	return { n: alias, x: refineExpression( expr, {} ) };
+}
--- a/src/parse/converters/mustache/readAliases.js
+++ b/src/parse/converters/mustache/readAliases.js
@@ -2,8 +2,9 @@ import readExpression from '../readExpression';
 import refineExpression from '../../utils/refineExpression';
 
 const legalAlias = /^(?:[a-zA-Z$_0-9]|\\\.)+(?:(?:(?:[a-zA-Z$_0-9]|\\\.)+)|(?:\[[0-9]+\]))*/;
+const asRE = /^as/i;
 
-export default function readAliases( parser ) {
+export function readAliases( parser ) {
 	let aliases = [], alias, start = parser.pos;
 
 	parser.allowWhitespace();
@@ -11,6 +12,7 @@ export default function readAliases( parser ) {
 	alias = readAlias( parser );
 
 	if ( alias ) {
+		alias.x = refineExpression( alias.x, {} );
 		aliases.push( alias );
 
 		parser.allowWhitespace();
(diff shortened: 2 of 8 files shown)


src/config/types.js
```diff
@@ -14,6 +14,7 @@ export const COMPONENT         = 15;
 export const YIELDER           = 16;
 export const INLINE_PARTIAL    = 17;
 export const DOCTYPE           = 18;
+export const ALIAS             = 19;
```
Rich-Harris: WE'RE RUNNING OUT OF NUMBERS! HALP!

Rich-Harris:
Strongly in favour of this. The two-way binding stuff is really cool (whatever those pesky Flux-heads think), but honestly I'm probably more excited about stuff like `{{#each people as person}}` which is going to be such a big win for template readability.

Implementation-wise (and forgive the armchair quarterbacking! I haven't delved into this nearly as far as I'd like), is there a reason to have an `Alias` node instead of the `Section` managing its aliases the way it does with index/key references (i.e. passing them down to child fragments)?

I don't totally understand the `@ractive` stuff, but I think that's just because it's been a long day. :zzz:

src/config/types.js
evs-chris: I heard CostCo has a sale on 100s on black Friday.

evs-chris:
> I'm probably more excited about stuff like {{#each people as person}}

Totally forgot about that one! I'll have to add it...

> forgive the armchair quarterbacking

By all means! Direction is the main reason I opened this as an incomplete PR :)

>  is there a reason to have an Alias node instead of the Section managing its aliases the way it does with index/key references (i.e. passing them down to child fragments)

Well, that's the direction I started in, but it seemed easier to have a SSOT for the aliases, resolving them once in the parent and letting further resolution look up the nearest alias. Since the aliases don't really bind like indices and keys, it also didn't seem like registering and tracking them would make much sense. Though, to be fair, I haven't done any serious testing yet. I decided to go with a separate node when I started adding a bunch of conditional logic in `Section` and `Mustache`, which `Section` extends, to keep alias blocks from trying to bind to a model and act like a conditional. Once the model is out, there are bunch of places that start throwing because they assume that if they're rendered there's a model.

Also, I haven't checked yet, but on my list of things to look at is whether expressions get the same resolver when they appear more than once in a template, especially nested within another expression. So would this actually be an optimisation (proper spelling, eh)?

From
```
{{#if calculateSomethingExpensive(.) }}{{ makeStringy(calculateSomethingExpensive(.)) }}{{/if}}
````
To
```
{{#with calculateSomethingExpensive(.) as thing }}{{#if thing }}{{ makeStringy(thing) }}{{/if}}
```

> I don't totally understand the @ractive stuff

Not that it will overcome sleep deprivation, but the tl;dr is allow access to the instance in the template and allow aliasing to change which instance is returned inside alias block to allow more granular control for components.

**Comment by evs-chris** [comment]
I've added iterative aliasing along with a few basic tests. I think I may have accidentally offended phantom at some point, because it pretty regularly refuses to succeed on tests that happily pass in a normal browser.

In an attempt to clarify my construction rambling one post up:

I introduced `Alias` to avoid extensive conditionals around model existence in `Sections` (and `Mustache`s) because an alias block has no expression to bind and it provides no context. It's just a lookup. There is now alias handling built directly onto `{{#each}}` sections, since while conceptually the same as a `{{#with}}` alias, they are entirely different beasts under the hood (pelt?).

I elected not to copy aliases around like index refs because index refs are bound to a single model that is special-case updated when stuff gets shuffled and rebound. Copying the aliases around seemed like it would add a bunch of unnecessary bookkeeping, and the only other alternative I can think of (another special model that proxies its upstream) was a bit more complexity than seemed necessary. Though, I may have confused myself about how other refs work now.

**Comment by evs-chris** [comment]
I got a few minutes to test out computation aliases being used to share chunks of a computation, and it does behave as I expected. In edge right now, the computation is called twice, and in this branch, it is only called once.

**Comment by Rich-Harris** [comment]
this is very cool. no reason not to include it in 0.8

lebowski

**Comment by heavyk** [comment]
SWEEEEET!!! congrats @evs-chris !!

**Comment by matthew-dean** [comment]
Now, combine that with declarative components FTW.

```
{{#component List { values } }}
  {{#each page(values)}}
    {{#with @ractive.parent as @ractive}}{{>content}}
  {{/each}}
  {{#each range(1, values.length / pageSize)}}
    <a href="#" on-click="set('page', .)">{{.}}</a>
  {{/each}}
{{/component}}

<List values="{{people}}">
  <div class="person" on-click="link(@keypath, 'current')">{{.name}} {{.phone}}</div>
</List>
```
Something like that would be amazing super-sauce.

PR #2210, opened 2015-10-11
Fix #2201

### LINKED ISSUE #2201: "itemscope" boolean attribute is not added to the DOM element
While trying to add an `itemscope` attribute to an element, I noticed that no matter what I do, It is never added to the DOM element in the rendered template...

I've tried :
`<div itemscope></div>`
`<div itemscope=""></div>`
`<div itemscope="itemscope"></div>`
`<div itemscope="true"></div>`

...and It never worked. Am I missing something ?

[Code exemple here](http://jsfiddle.net/jpm4w8e2/1/)

MartinKolarik:
@martypdx explained the problem very well in [this comment](https://github.com/ractivejs/ractive/issues/2201#issuecomment-146438758).

The solution is to use a combination of `setAttribute()` and `removeAttribute()`, because if `node.propertyName` isn't defined, setting it won't have any effect.

### DIFF
--- a/src/view/items/element/Attribute.js
+++ b/src/view/items/element/Attribute.js
@@ -78,19 +78,19 @@ export default class Attribute extends Item {
 
 		// should we use direct property access, or setAttribute?
 		if ( !node.namespaceURI || node.namespaceURI === html ) {
-			const propertyName = propertyNames[ this.name ] || this.name;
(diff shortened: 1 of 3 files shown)


martypdx:
@MartinKolarik :+1:

wmcmurray:
Awesome, thanx @MartinKolarik !

Issue #2214, opened 2015-10-12
Template section not rendered when new key is added to data


woodlandhunter:
See fiddle. There's no output when the `baz` key is added with `set()`, but it works as expected if `data` is initialized with the `baz` key.

http://jsfiddle.net/ozghcocL/

I also discovered that it works if you change each `baz` in the `{{#if ...}}` to `~/baz`

dagnelies:
what a big bad bug! definitely a show-stopper :/

I really do hope ractive 0.8 will be more stable

martypdx:
Actually there is a `0.8` BREAKING CHANGE along with bug happening here. 

Default resolution for ambiguous reference is now context, not root of the component. So `baz` resolves to each item of `fields`, which is not the same as `ractive.set( 'baz', [1,2,3])`.

Still an issue with second half of if/else not showing up...

Rich-Harris:
> Default resolution for ambiguous reference is now context

That only applies if there's a forced resolution though (i.e. two-way binding), right? Shouldn't be a breaking change in this scenario

martypdx:
> That only applies if there's a forced resolution though (i.e. two-way binding), right? Shouldn't be a breaking change in this scenario

Ah, true. My mistake.

PR #2215, opened 2015-10-12
Make expression proxies notify their deps when their bits resolve - #2214

evs-chris:
This is unfamiliar territory for me, so is this a proper fix?


Rich-Harris:
Pretty sure this is correct :+1:

PR #2221, opened 2015-10-14
Don't transform CSS at-rules and keyframe declarations (fixes #781)

### LINKED ISSUE #781: Treatment of at-rules in encapsulated CSS
See #656. Media queries are supported, but `@keyframes` and `@import` and so on are not.

MartinKolarik:


Rich-Harris:
:+1:

Issue #2222, opened 2015-10-15
Pushing to array with modified arrays throws "Error: Section was already shuffled!"


Madgvox:
New repro: http://jsfiddle.net/np5ch3ot/1/

---

~~From @blackgate's https://github.com/ractivejs/ractive/issues/1978#issuecomment-148325465:~~

> ~~Same issue with pattern observers on arrays:
http://jsfiddle.net/np5ch3ot/~~

> ~~On 0.8 edge pushing to an array doesn't seem to work.~~

Madgvox:
Oops, I didn't check to see what version the fiddle was on before posting it! Luckily for me (and unluckily for those fixing the bugs :wink:) the fiddle is still broken in edge, seemingly for a completely different reason.

JonDum:
I'm getting this too in a medium-complexity app, but I'm having trouble reducing it down for another test case. The causes might be the same, but the set up is different than @Madgvox's

evs-chris:
It looks like the issue is that the observer is being called multiple times in a single runloop and each call is doing a shuffle. That can be solved by using strict or defer with the observer. It's a breaking change though (I think) so it should probably be fixed.

I think one of the issues is that the observer is firing for all array keypaths on shuffle operations, when they should really only fire on the changed keypaths. I also think that `RepeatedFragment` should tolerate multiple shuffles. I'm not positive about any of that or how particularly harebrained this is, but it's what I've come up with:

```diff
diff --git a/src/Ractive/prototype/observe.js b/src/Ractive/prototype/observe.js
index c640b80..5550c5d 100644
--- a/src/Ractive/prototype/observe.js
+++ b/src/Ractive/prototype/observe.js
@@ -1,5 +1,5 @@
 import runloop from '../../global/runloop';
-import { isEqual, isObject } from '../../utils/is';
+import { isArray, isEqual, isObject } from '../../utils/is';
 import { splitKeypath } from '../../shared/keypaths';
 import { cancel } from '../../shared/methodCallers';
 import resolveReference from '../../view/resolvers/resolveReference';
@@ -136,7 +136,7 @@ class PatternObserver {

                const pattern = keys.join( '\\.' ).replace( /\*/g, '(.+)' );
                const baseKeypath = baseModel.getKeypath();
-               this.pattern = new RegExp( `^${baseKeypath ? baseKeypath + '.' : ''}${pattern}$` );
[...]

PR #2236, opened 2015-10-21
Shuffling pattern observers and repeated sections - #2222

evs-chris:
This is my patch from #2222 with tests added.

First, if an array with object children is shuffled, then the pattern observer will fire for all of the keys. It seems like it should only fire on the observed keys, so this adds shuffle support that does that.

Second, if an observer makes a change that causes a shuffle, and the observer fires more than once, the repeated fragment refuses to shuffle more than once. This also adds support for the multiple shuffles per runloop to repeated sections by mapping the original new indices through the accumulated shuffles.


Rich-Harris:
:clap: think my brain would have melted if I'd tried to fix that

Issue #2234, opened 2015-10-21
Filtered Array printed using {{#each}} gets displayed in unpredictable order


giovannipiller:
I'm using `{{#each}}` to cycle an array, filtering out (a simple `{{#if}}`) elements that don't match a certain condition (ex. text in a search input).
Once I remove this condition, by clearing text in the input, list order gets compromised.

A fiddle is better than a thousand words: http://jsfiddle.net/pkgeuLyL/

--

**Steps to reproduce**
- upon page load, list is displayed in order (1, 2, 3, 4,.....)
- write some data in the filter input (ex. 1)
- list should now look something like this (1, 10, 11, 12,...), which is fine
- now empty the previously typed text from the filter input

**Expected result**
List should be displayed in the starting order (1, 2, 3, 4,.....)

**Actual result**
List is displayed in a somewhat unpredictable order (1, 9, 10, 11,......)

--
Example is broken on Edge, works fine on 0.7.

fskreuz:
Ahh, bug.

However, you can use a computed property for this to avoid the `if-else` in the template. It's also because `list + query -> [computation] -> filtered list`, if we're systems engineers and mathematicians (Disclaimer: I'm not :trollface:)

http://jsfiddle.net/pkgeuLyL/1/
```

<div id="printmeout"></div>

<script type="text/ractive" id="filterme">
    <input type="text" value="{{search}}"/>
    <hr>
    <ul>
        {{#each filteredList}}
            <li>{{.}}</li>
        {{/each}}
    </ul>
</script>

var printmeout = new Ractive({
    template: '#filterme',
    el: '#printmeout',
    data: {
        search: '',
        entries: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15],
    },
    computed: {
    	filteredList: function(){
            var search = this.get('search');
        	return this.get('entries').filter(function(entry){
                return !search || entry.toString().indexOf(search) !== -1
            });
        }
    }
})
```

giovannipiller:
> However, you can use a computed property for this to avoid the if-else in the template.
Agreed, it's generally much better to use computed properties. I had reasons though (will post them as soon as I figure out if they're valid :smile:).

PR #2237, opened 2015-10-21
Skip iterations without nodes when looking for next node in repeated sections - #2234

evs-chris:
This looks for the next iteration after the target index that actually has a node. The behavior in edge now makes fragments looking for the next node append instead of inserting if the next iteration doesn't actually have a node.


heavyk:
holy shit dude, you're a machine

PR #2239, opened 2015-10-22
Register yield fragments during bubble - fixes #2235

### LINKED ISSUE #2235: Isolated stops {{> content}} / {{yield}} from updating, but still renders initially.
http://plnkr.co/edit/BgRbvFwwiG4KTPpt3Wqg?p=preview

Is this intended? Ran into this scenario today where I need `isolated` set to true to avoid an infinite loop, but it stops it's `{{> content}}` from updating. I was under the impression that `isolated` stops the component from accessing its parent's properties, not including its own `{{> content}}` it gets from its parent. 

I thought that maybe `{{> content}}` was dependent on `foo` from its parent in this scenario, but the same thing happens with `{{yield}}` and, annoyingly, I can't use `{{yield}}` more than once. Any thoughts?

evs-chris:
It looks like the special component handling during bubble keeps the yield fragment from getting handled in the runloop. This overrides `bubble` on `Yielder` to register the fragment with the runloop so that the fragment updates and isn't left in a dirty state after the update. Does that sound right?

### DIFF
--- a/src/view/items/Yielder.js
+++ b/src/view/items/Yielder.js
@@ -1,6 +1,7 @@
 import Item from './shared/Item';
 import Fragment from '../Fragment';
 import parse from '../../parse/_parse';
+import runloop from '../../global/runloop';
 import { warnIfDebug } from '../../utils/log';
 import { removeFromArray } from '../../utils/array';
 
@@ -40,6 +41,13 @@ export default class Yielder extends Item {
 			ractive: this.container.parent,
 			template
 		}).bind();
+	}
+
+	bubble () {
+		if ( !this.dirty ) {
+			runloop.addFragment( this.fragment );
+			super.bubble();
+		}
 	}
 
 	detach () {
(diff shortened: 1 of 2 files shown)


src/view/items/Yielder.js
```diff
@@ -42,6 +43,14 @@ export default class Yielder extends Item {
 		}).bind();
 	}
 
+	bubble () {
+		if ( !this.dirty ) {
+			runloop.addFragment( this.fragment );
```
Rich-Harris: For me, the tests still pass if this is commented out – which I *think* makes sense, because we're now bubbling up the correct route (through its `containerFragment`, which is responsible for calling its `update()` method). So I think we can lose this line (and the associated `import` statement)?

evs-chris: I'm pretty sure you're right. I'll merge this and verify that the original fiddle still passes.

PR #2249, opened 2015-10-29
Remove non-transition nodes from DOM before calling deferred observers (#1869)

### LINKED ISSUE #1869: Deferred observers firing before node removal
Interesting case from @wpf500: http://jsfiddle.net/rpyw1c0y/

martypdx:
Fixes #1869.  (Re)moving the `TransitionManager.init` into `.start` seems to do the trick.

Also pruned some uncalled code and now Transition Manager doesn't need to know about decorators.


Rich-Harris:
nice

Issue #1997, opened 2015-06-15
Feature request: @global


dagnelies:
Hi,

Simple request: a new `@global` token to reference the global scope. Can also be seen as an alternative to #1996.

Sometimes, you don't want to add more and more stuff to ractive, or just use a lib out of the box.

For example, you could have lines like:
```
<p>{{@global.moment(isoDate).format('MMMM Do YYYY, h:mm:ss a')}}</p>
```
To see the given date using momentjs. This would be kind of handy.
This could also be usable for both expressions and method calls.

It would also be backward compatible with the existing stuff.

Rich-Harris:
Finally merged this! http://jsfiddle.net/rich_harris/zpf48Lnu/

dagnelies:
great great great! I'm looking forward to test this out! thanks again!

PR #2250, opened 2015-10-31
@global and @ractive special refs - RFC

evs-chris:
__Post 0.8__

This adds two new special refs, `@ractive` and `@global`, that give access to the current Ractive instance and the global object (`window` in a browser environment). Unlike other special refs, these support accessing child paths.

### Why?

Well, for `@global`, it's an outstanding feature request from #1997. Anything in global scope can now be used in the template, which is particularly handy for using external libraries. It's also nice for integration into pages that aren't controlled entirely by Ractive. Two-way binding for child paths also works nicely. There is a single shared model that controls access to the global from within Ractive.

For `@ractive`, as mentioned in several issues recently, it's nice to be able to use functions that are added directly to the Ractive instance in a template. It's also convenient, at times, to track other things as properties on a particular Ractive instance, like, for instance, a shared Ace Editor instance for an edit decorator. For fiddle and development fun, you can also do things like `<input type="checkbox" value="{{@ractive.constructor.DEBUG}}" />` and `{{@ractive.constructor.VERSION}}`. Two-way binding is also supported for child paths on `@ractive` refs. Each `RootModel` creates a single `RactiveModel` instance when needed, and the resolution process pulls it from the fragment Ractive instance.

In addition to being accessible from the template, both `@ractive` and `@global` are accessible in keypaths from the API e.g. `ractive.set('@global.some.value', 42)` so that they can be used in method events and more easily keep template refs in sync with outside state.

### DIFF
--- a/src/model/RootModel.js
+++ b/src/model/RootModel.js
@@ -3,6 +3,8 @@ import { extend } from '../utils/object';
 import Computation from './Computation';
 import Model from './Model';
 import { handleChange, mark } from '../shared/methodCallers';
+import RactiveModel from './specials/RactiveModel';
+import GlobalModel from './specials/GlobalModel';
 
 export default class RootModel extends Model {
 	constructor ( options ) {
@@ -60,11 +62,18 @@ export default class RootModel extends Model {
 		return '';
 	}
 
+	getRactiveModel() {
+		return this.ractiveModel || ( this.ractiveModel = new RactiveModel( this.ractive ) );
+	}
+
 	has ( key ) {
 		return ( key in this.mappings ) || ( key in this.computations ) || super.has( key );
 	}
 
 	joinKey ( key ) {
+		if ( key === '@global' ) return GlobalModel;
+		if ( key === '@ractive' ) return this.getRactiveModel();
+
 		return this.mappings.hasOwnProperty( key ) ? this.mappings[ key ] :
 		       this.computations.hasOwnProperty( key ) ? this.computations[ key ] :
 		       super.joinKey( key );
--- a/src/model/specials/GlobalModel.js
+++ b/src/model/specials/GlobalModel.js
@@ -10,6 +10,10 @@ class GlobalModel extends Model {
 		this.adaptors = [];
 		this.changes = {};
 	}
+
+	getKeypath() {
+		return '@global';
+	}
 }
 
 export default new GlobalModel();
--- a/src/model/specials/GlobalModel.js
+++ b/src/model/specials/GlobalModel.js
@@ -1,0 +1,15 @@
+/* global global */
+import Model from '../Model';
+
+class GlobalModel extends Model {
+	constructor ( ) {
+		super( null, '@global' );
+		this.value = typeof global !== 'undefined' ? global : window;
+		this.isRoot = true;
+		this.root = this;
+		this.adaptors = [];
+		this.changes = {};
+	}
+}
+
+export default new GlobalModel();
--- a/src/model/specials/RactiveModel.js
+++ b/src/model/specials/RactiveModel.js
@@ -1,0 +1,13 @@
+import Model from '../Model';
+
+export default class RactiveModel extends Model {
+	constructor ( ractive ) {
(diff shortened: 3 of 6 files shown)


heavyk:
I have a lot of decorators that use `node._ractive` to get the instance. I guess that wouldn't change, right..

other than that, that would be really useful. historically, I've always exported the global I need access to in the template. it's definitely useful.

the only problem with the `@global` approach is that templates can now use things outside of it's instance. meaning, if I have for example a time component which uses `@global.moment.toString` I would now have to ensure that I did have the global variable when I go to use the same component in another project. whereas now, each component `require`s the thing it uses. it could potentially allow people to not be so careful with their globals. I'm -0.2 on `@global` for that reason, even though I'd never use it myself. but +2 on `@ractive` so, I guess I'm +0.9 overall :)

nice!

fskreuz:
>  if I have for example a time component which uses @global.moment.toString I would now have to ensure that I did have the global moment variable when I go to use the same component in another project. whereas now, each component requires the thing it uses.

A responsible module/component always requires its dependencies (unless it's something implicit in the project, say "Assume ES5 shim is always in effect"). As long as the component using `{{ @global.moment.toString }}` is actually `require`ing MomentJS, then that's fine. The global will be there. 

It's also really handy for "traditional libraries" (those that don't advertise themselves as any kind of module, just a global). Normally, in RequireJS, that would require me to map *all* the library globals to their respective module names in a config file.

evs-chris:
@heavyk, there's actually a supported API for getting the Ractive instance from a DOM node. Instead of `node._ractive`, use `Ractive.getNodeInfo(node).ractive`. It also gives you a few other useful tidbits from the template state.

heavyk:
right on dude! thanks for that info...

I updated my really shitty chrome extension to use that API.

https://github.com/heavyk/ractive-dataz

martypdx:
Apart from the `@`, would `@global` be any different than:

```
Ractive.prototype.data = function(){ 
    return { global: window }; 
};
```
?

evs-chris:
I suppose nothing really apart from maybe binding support and slightly faster resolution.

evs-chris:
I misread that... phone... sorry.

Yes, not much. The root ref would be read-only, all instances share the model (shared updates in Ractive), automatic handling of node vs browser, and slightly faster resolution.

JonDum:
What about having a `@root` ref that supports data binding to the root component?

I've actually found a lot of times that there's no easy way to get around needing to access state on the root component from a deeply nested component. It would be super amazing to be able to do something like 

```
{{#if @root.view == 'foo' }}
{{/if}}
```
And have it actually get updated when root's `view` changes.

fskreuz:
I think `root` already has a special meaning in Ractive.

JonDum:
@fskreuz You can use `root` sans `@` in a template, but binding doesn't work. 

*Technically*, you can use the bare property (if not isolated) and ractive will walk up to the root looking for it and binding will work. There's just no explicit way to say, don't use a parent component's `foo`, but use the root `foo` or nothing at all. aka, avoiding naming conflicts.

evs-chris:
Interestingly, when I fix a small bug in this PR, you can access stuff on the root instance using `@ractive.root.get('some.key.path')`, and due to the way capture works, it actually updates with the root keypath. So event if you have an isolated component, if you use `@ractive.parent.get(...)` or `@ractive.root.get(...)` or `@ractive.get(...)` in a mustache, it binds like you'd expect it to from within any regular expression. Of course, as with all expressions, the binding is read-only.

evs-chris:
Another use for `@ractive` would be to have access to a (to be proposed) `joinPath` Ractive method that allows things like `{{#each some.deeply.nested.list}}{{.someProperty}}<button on-click="splice(@ractive.joinPath(@keypath, '../'), @index, 1)">Remove</button>{{/each}}`.

JonDum:
Woah, that's seriously cool, Chris!

heavyk:
oh yeah seriously! that's cool!! :+1:

PR #2254, opened 2015-11-02
support Object, String, Number and Boolean in templates (fixes #2247)

### LINKED ISSUE #2247: Object.keys not available in template
I had something like this in my template:

```
{{#if Object.keys(panels).indexOf(active) !== -1 }} ... {{/if}}
```

and it didn't work. turns out I can't use `Object.keys` in the template. no big deal cause I made it a computed value, but was that intentional?

http://jsfiddle.net/fs6fxwt9/

Rich-Harris:
#2247


heavyk:
:+1: :)

PR #2257, opened 2015-11-02
Append arguments automatically on re-fired events, more robust arguments spread in method calls

### LINKED ISSUE #2253: Fired events lose arguments at parent handler
When firing from child component, handler on parent fires but doesn't receive the arguments. Same applies to `fire` when done via method call.

Further testing shows that it only loses arguments outside the component. Listening inside the same component works fine. Narrowed it down to reproxying arguments logic for component method calls. Not sure how to proceed from here.

Works on 0.7 (parent handler receives arguments), breaks on 0.8 (parent handler receives blank arguments): http://jsfiddle.net/r9zmsz89/2/

```js
var ChildComponent = Ractive.extend({
    template: '<button type="button" on-click="foo()">Click me!!!</button>',
    oninit: function () {
        this.on('far', function () {
            console.log('on child', arguments); // ['foo', 'bar']
        });
    },
    foo: function () {
        this.fire('far', 'foo', 'bar');
    }
});

var ParentComponent = Ractive.extend({
    template: '<child-component on-far="boo" />',
    components: {
    	'child-component': ChildComponent
    },
    oninit: function () {
        this.on('boo', function () {
            console.log('on parent', arguments); // []
        });
    }
});

new ParentComponent({
    el: 'body',
    append: true
});
```

martypdx:
Fixes #2253. Component event directives now append existing arguments by default:

`<widget on-foo='bar' on-qux:'bizz:{{buzz}}'/>`

will now fire `bar` and `bizz` with original arguments from `foo` and `qux` appended to any supplied arguments (added to value of `{{buzz}}` in `on-qux` firing of `bizz`).

Also makes `...argument` in method calls more robust by allowing inclusion of spread at end of method allowing other arguments to be included:

`<widget on-foo='bar( "foo", ...arguments )'/>`

### DIFF
--- a/test/browser-tests/events/method-calls.js
+++ b/test/browser-tests/events/method-calls.js
@@ -183,7 +183,7 @@ test( 'component "on-" with additive ...arguments', t => {
 
 	const ractive = new Ractive({
 		el: fixture,
-		template: '<Component on-foo="foo(\'fooarg\', ...arguments)" on-bar="bar(\'bararg\', ...arguments)"/>',
+		template: `<Component on-foo="foo('fooarg', ...arguments)" on-bar="bar('bararg', ...arguments)"/>`,
 		components: { Component },
 		foo ( arg1, e, arg2, arg3 ) {
 			t.equal( arg1, 'fooarg' );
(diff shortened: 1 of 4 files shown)


test/browser-tests/events/method-calls.js
```diff
@@ -174,6 +174,35 @@ test( 'component "on-" with ...arguments', t => {
...
+test( 'component "on-" with additive ...arguments', t => {
+	t.expect( 7 );
+
+	const Component = Ractive.extend({
+		template: `<span id="test" on-click="foo:'foo', 42">click me</span>`
+	});
+
+	const ractive = new Ractive({
+		el: fixture,
+		template: '<Component on-foo="foo(\'fooarg\', ...arguments)" on-bar="bar(\'bararg\', ...arguments)"/>',
```
fskreuz: I used template strings when I tried fixing this. No need to escape the quotes when written this way.

```js
template: `
  <component ... />
`,
```

fskreuz:
:+1: I wasn't even close

PR #2264, opened 2015-11-08
Component keypaths - RFC/Review

evs-chris:
This was a TODO in `Model.getKeypath()`, and I bumped into it in one of my components. This adds accounting for the component instance and its mappings when getting the keypath for a model so that `@keypath` inside a component is relative to the component if the reference happens to be in the context of a mapping. Any events that fire inside a component will also get an adjusted keypath that is relative to the component.

This also adds a `@rootpath` and `event.rootpath` that are just the current `@keypath` and `event.keypath` i.e. not adjusted.

I started going a bit cross-eyed when trying to puzzle out how to handle keypaths that get updated and under which circumstances they do. I think I've sufficiently accounted for that by storing the adjusted keypath models on their component instance and registering them as children of the actual keypath model. Then, when one updates, it moves its cached instance on its component instance and continues the change propagation downstream.


fskreuz:
Getting carried away with the `@` thingies :trollface: So essentially, they're like "absolute paths" and "relative paths" for keypaths right? As long as they're consistent (and documented), this is a :+1: 

Quick check on docs, there isn't a `getKeypath` or `getRootPath` method. Would it make sense to have those for the JS side of things?

evs-chris:
Can't have too many special refs! :grinning:

The current `@keypath` basically has the "absolute" behavior now, but I figured it'd be nice to keep around. So I named it `@rootpath`, though I'm not opposed to axing it altogether or naming it something else. The "relative" version that this adds stems from a `TODO` comment leftover from #2022 and #1740. `@keypath` and `@rootpath` always resolve to the same thing if they're not in a component template.

I don't think there's really a place in the API for these since they deal strictly with template-related state. Aside from events, which were already included, the only other place that I can think of with keypath access is `Ractive.getNodeInfo`, and I just added that to this PR.

evs-chris:
I just noticed that the shuffle test is failing in phantom, but it passes in Chrome and Firefox. Also Firefox has a failing decorator teardown transition test and a few enhancement tests. I'll have to look into those tomorrow.

evs-chris:
Found a spot that I missed on the first run through that raised a question: should the `_ractive` added to element nodes store the keypath? It's readily available from the context and `Ractive.getNodeInfo` can easily supply it from there when requested.

martypdx:
@evs-chris I'm thinking let's merge this because it can fix the keypath problem.

As a follow up, rather than `@rootkeypath` and needing to add join paths to ractive (can't find that PR right now), why don't we implement a class for `@keypath` value that returns the string for `toString` and `valueOf`, but then we could hang props and methods off of it?

evs-chris:
@martypdx I'm good with merging this as-is and then addressing any further issues (observers) or api adjustments in another PR. The only thing I was questioning was the way I set up keypath model children of other keypath models.

martypdx:
> The only thing I was questioning was the way I set up keypath model children of other keypath models.

Can you provide more detail? I'm not following...

src/model/Model.js
```diff
@@ -270,13 +270,36 @@ export default class Model {
...
 	}
 
-	getKeypathModel () {
-		return this.keypathModel || ( this.keypathModel = new KeypathModel( this ) );
+	getKeypathModel ( ractive ) {
+		let keypath = this.getKeypath(), model = this.keypathModel || ( this.keypathModel = new KeypathModel( this ) );
+
+		if ( ractive && ractive.component ) {
+			let mapped = this.getKeypath( ractive );
+			if ( mapped !== keypath ) {
```
evs-chris: Here, the rootpath (top-level keypath) is the same, but the keypath (mapped) is not. So a child `KeypathModel` is created and attached to the main `KeypathModel`. That lets children update appropriately when the main instance is shuffled, etc. Is this the correct way to handle the "keypath is the same but not really because it's also mapped" issue?

martypdx: Got it. Took a look at the tests though, we need to check mapping though two levels of components.

evs-chris:
I added a nested component test and it comes out mostly as I expected. The only odd thing is that since there's not way to tell if a path was mapped or not at the reference point, this will always return the mapped path. `<cmp foo="{{bar.baz}}"/>` - cmp: `{{#with foo.bar}}{{@keypath}}{{/with}} {{#with bar.baz.bar}}{{@keypath}}{{/with}}` - yields: `foo.bar foo.bar` because they reference they same model and have a mapping to reverse.

martypdx:
> because they reference the same model and have a mapping to reverse

I think that's acceptable for first go. 

I played around with a `MappedModel` that wraps the underlying model, but if your solutions is working, let's move forward with that.

I'll go ahead and merge, then I can merge the observer `*` fix

PR #2265, opened 2015-11-09
Root refs in components should use component viewmodel not context

evs-chris:
In a component, in `{{#with some.mapped.path}}{{~/local}}{{/with}}`, `local` is trying to resolve using the mapped context root instead of the component root. I believe the correct behavior would be to use the component root instead, right?


martypdx:
LGTM

Issue #2269, opened 2015-11-10
{{#each expression}} doesn't update correctly


MartinKolarik:
See http://jsfiddle.net/va6jU/110/ and click splice.

The problem is that `newIndices` array received by `RepeatedFragment#shuffle()` doesn't account for the fact there's an expression in `{{#each}}` section. In this case, the actual number of rendered fragments is 1 less than `items.length`, so when we get [here](https://github.com/ractivejs/ractive/blob/0155d0eb3b6797c8b74b4b58b1b648aa891fa9b8/src/view/RepeatedFragment.js#L360), we try to unrender a non-existing fragment.

gaelicmichael:
I believe I'm encountering this problem too. The error I'm getting is:
TypeError: context is undefined
if (isArray(context.get())) {
ractive.js (line 6634, col 1)

evs-chris:
@gaelicmichael the original fiddle works in edge now... are you still getting an error?

gaelicmichael:
Yes -- I just updated to v0.8.0-edge (Sun Nov 15 2015 21:44:30) and I get the same error:
TypeError: context is undefined
if (isArray(context.get())) {
ractive.js (line 6634, col 1)

Note that I was not getting this error in my code when I was using 0.7.3.

evs-chris:
I've looked through all of the ways I can think of to arrive at that line with an undefined context and I got nothing. How are you emptying and reloading the array? Also is there an observer that is also modifying the array that is not also deferred?

gaelicmichael:
File is: https://github.com/Digital-Innovation-Lab/prospect/blob/master/js/edit-record.js

Array is cleared on line 395
The fillRecordFromTemplate() function fills it.

What do you mean by "deferred"?  Do I need to wait for a promise to refill it?  Example and/or documentation?

evs-chris:
That helps a lot. Thanks!

Ractive can keep observers from firing until the DOM modifications from the current runloop (single set, update, toggle, etc) are complete, which tends to step around issues where a value is modified more than once in a runloop by observers. `ractive.observe('somepath', v => ..., { defer: true })`

gaelicmichael:
Your comment is not clear to me. Are you telling me that my code has a problem with the way that it is using Ractive? If so, I need clearer directions to documentation or example that will fix the problem.

gaelicmichael:
Let me also point out that the variable I am observing is NOT the array that is being changed.

evs-chris:
[...]

PR #2274, opened 2015-11-12
Computations are models that don't shuffle

### LINKED ISSUE #2267: Concatenating two or more arrays results in duplicate DOM nodes rendered.
Indexes are getting rendered multiple times when they shouldn't be. 

http://plnkr.co/edit/hrJyefxS964OxUkF21pz?p=preview

Click the first button, then the second. Each subsequent click adds duplicate keypaths getting rendered, but `app.get(vars)` returns a normal array as expected. (`app.get('vars').forEach(v => console.log(v))`)

I'm iterating over a computed var that returns two other arrays concatenated together. Doesn't seem to matter whether one is concat'd to the other, or if they're concated to a new, clean array.

```
            computed: {
                vars: function() {
                    
                    var things = this.get('things');
                    var stuff = this.get('stuff');
                    
                    
                    // return things.concat(stuff); 
                    
                     return [].concat(things).concat(stuff); //still broken

                },
            },
```

evs-chris:
Computations are one of those weird dep/child hybrids that automatically inherit the model shuffle, but since they aren't really shuffle-able, they should not have a shuffle method. This overrides the prototype shuffle property with undefined during construction and fixes #2267 and #2269. Thoughts?

### DIFF
--- a/src/model/Computation.js
+++ b/src/model/Computation.js
@@ -62,6 +62,9 @@ export default class Computation extends Model {
 
 		this.boundsSensitive = true;
 		this.dirty = true;
+
+		// computations don't shuffle
+		this.shuffle = undefined;
 	}
 
 	get ( shouldCapture ) {
--- a/src/model/Computation.js
+++ b/src/model/Computation.js
@@ -63,7 +63,7 @@ export default class Computation extends Model {
 		this.boundsSensitive = true;
 		this.dirty = true;
 
-		// computations don't shuffle
+		// TODO: computations don't shuffle, but this is a bit hackish
 		this.shuffle = undefined;
 	}
 
--- a/test/browser-tests/shuffling.js
+++ b/test/browser-tests/shuffling.js
@@ -184,6 +184,36 @@ test( 'method event directives with no args should shuffle without throwing', t 
 	r.unshift( 'items', 2 );
 });
 
+test( 'shuffling around a computation with an index ref', t => {
+	const r = new Ractive({
+		el: fixture,
+		template: `{{#each items.slice(1)}}{{2 * @index}} {{@index * .}}|{{/each}}`,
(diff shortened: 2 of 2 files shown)


MartinKolarik:
>but since they aren't really shuffle-able, they should not have a shuffle method

I was wondering how to fix it... now that you've said it, it is seems kinda obvious :+1:

src/model/Computation.js
```diff
@@ -62,6 +62,9 @@ export default class Computation extends Model {
 
 		this.boundsSensitive = true;
 		this.dirty = true;
+
+		// computations don't shuffle
+		this.shuffle = undefined;
```
martypdx: Clearing on the instance is probably a sign the api needs to evolve. If we put on the prototype, then it gets called as noop:
```
constructor (...) {
    ...
}

shuffle () {}
```

But that means it gets called for each change.

So should we add a prototype property for `shouldShuffle` (or the negative `noShuffle`) and test that instead of existence of a `shuffle` method?

evs-chris: I was kinda thinking a simpler `Model` base class that could be used instead of `Model` for some of the more esoteric `Model` subclasses.

evs-chris: After having slept on it, nothing really jumped out as a best solution, so I punted with a TODO for now.

JonDum:
Hey I think there's another bug with this. When splicing a computed array, it still calls `model.shuffle()` in [makeArrayMethod](https://github.com/ractivejs/ractive/blob/3de5b6e347b3e267498824e63e948558f93fe1e0/src/Ractive/prototype/shared/makeArrayMethod.js#L23). I think there needs to be a `model instanceof Computation` to avoid calling the non-existent shuffle function there too.

PR #2280, opened 2015-11-16
Update text input types with safe string value

evs-chris:
This adds a check for text inputs so that they don't get set to `undefined` but get a "safe" string instead. Thoughts?


fskreuz:
Just encountered this today. A blank string instead of `undefined` should be fine. :D

PR #2284, opened 2015-11-20
Regression from #2269

evs-chris:
This is the follow-on from the second issue in #2269. The issue is a regression in edge with a number of different components, including a shuffle from and immediate observer, a two-way binding select, and setting a populated array to a different array instance while also shuffling.

In some scenarios, a repeated fragment can be rebound with an undefined context. I wasn't able to pin down exactly how it happened because the template was particularly complex. My attempts at reproducing with a simpler template failed, but adding a check for a truthy context kept from throwing.

After a shuffle and before the runloop completes, observers that cause updates to a repeated fragment may cause some interaction with now-undefined iterations on that fragment. To avoid that, this adds a check that filters out the undefined portions.

Finally, select bindings forcing an update before the end of a runloop can cause some pretty nasty things to happen when the select should be disappearing anyways, so this bumps that to the end of the outermost runloop.

## Alternatively...

I think it may be better to call this a breaking change for 0.8 and recommend deferred observers for anything that will write to a model or cause another dom update. Are we shooting for absolute minimal breaking changes for 0.8?

### DIFF
--- a/src/view/RepeatedFragment.js
+++ b/src/view/RepeatedFragment.js
@@ -19,6 +19,10 @@ function getRefs ( ref, value, parent ) {
 	return refs;
 }
 
+function activeIterations( array ) {
+	return array.filter( i => !!i );
+}
+
 export default class RepeatedFragment {
 	constructor ( options ) {
 		this.parent = options.owner.parentFragment;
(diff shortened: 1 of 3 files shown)


JonDum:
Sounded like this was related to #2310 so I checked it out and tested it. Didn't fix it :cry:, but it didn't break anything else and tests passed haha.

evs-chris:
I was hoping it was related too, but I haven't quite had the spare cycles to dig in.

evs-chris:
I'm hoping #2406 and a few other improvements have made this unnecessary, so I'm going to close this run and if the issue turns out not to be resolved, I'll try, try again.

Issue #2286, opened 2015-11-22
Error: call to Function() blocked by CSP


3demax:
Hi everyone! I'm trying to use Ractive in my applicaton for FirefoxOS.
The problem is that FxOS have strict CSP for privileged apps:
```
Firefox OS and the Web runtimes for Android and desktops enforces the following CSP for privileged apps:

"default-src *; script-src 'self'; object-src 'none'; style-src 'self' 'unsafe-inline'"
```
That's why I get "Error: call to Function() blocked by CSP" while trying to create Ractive object.
Therefore, question emerges:
*Are there any possibilities to avoid using calls to Function in Ractive?*

JonDum:
> Error: call to Function() blocked by CSP

Oh god seriously? That's like the core of Ractive's expressions.

fskreuz:
Yup. Been there before. Also, FirefoxOS CSP cannot be "relaxed" so you can't really allow privileged apps to use `Function` or `eval`. There's no way getting around it aside from Ractive using something other than `Function`.

heavyk:
you can use unsafe-eval for chome apps. it worked for me

fskreuz:
Mozilla has a really spread out documentation regarding CSP. Finally found one that summarizes the ones regarding the apps: https://developer.mozilla.org/en-US/Apps/Build/Building_apps_for_Firefox_OS/CSP#Applicable_CSP_Restrictions

> **eval is disallowed**
You may not use the eval() function or the eval operator. Both will throw security errors.

> **The function constructor is banned**
You may not use the Function() constructor. Using it will throw a security error.

> **Dynamic code execution with setTimeout and setInterval is banned**
You MUST pass callable objects (i.e.: functions) to the setTimeout and setInterval functions. Passing strings will not create the timer and the function will return 0.

Further down:

> **My templates aren't being rendered in my privileged packaged app.**
If you're using a client-side templating library (such as Mustache or Nunjucks), you must ensure that you're precompiling your templates. This means that you must run a script to compile your templates from HTML to JavaScript before packaging your app. Privileged apps cannot use remote loaders that download and compile templates on the fly because the compiler uses eval() or new Function().

> **I set the CSP in my manifest to allow the use of [feature], but it still doesn't work.**
*You cannot loosen the default restrictions on privileged or certified packaged apps*. The CSP field in the app manifest can only be used to tighten the CSP, not loosen it.

JonDum:
[...]

PR #2287, opened 2015-11-23
Add CSP compatible parsing mode

MartinKolarik:
This adds a new parser option `insertExpressions` (name suggestions welcome). When set to true, generated functions are stored under `e` property of the template and no `eval` is needed on run time.

Refs: #1897, #2286.


src/parse/utils/insertExpressions.js
```diff
@@ -0,0 +1,27 @@
...
+	while ( i-- ) {
+		args[i] = `_${i}`;
+	}
+
+	// Functions created directly with new Function() look like this:
+	//     function anonymous (_0 /**/) { return _0*2 }
+	//
+	// With this workaround, we get a little more compact:
+	//     function (_0){return _0*2}
+	return new Function( [], `return function (${args.join(',')}){return(${x.s})}` )();
```
fskreuz: Isn't the whole idea of CSP avoiding the use of `Function`? Or am I missing something here?

martypdx: This module is called at parse time (see previous change), not during render.

src/shared/getFunction.js
```diff
@@ -0,0 +1,10 @@
+export let functions = {};
+
+export default function getFunction ( str, i ) {
+	if ( functions[ str ] ) return functions[ str ];
+
+	let args = new Array( i );
+	while ( i-- ) args[i] = `_${i}`;
+
+	return ( functions[ str ] = new Function( args.join( ',' ), `return (${str})` ) );
```
martypdx: But this isn't going to work because we're back to creating Functions.

martypdx:
I think we need to store the function itself, not the string representation as `template.e`.

Then we should also be able to move the function creation logic solely into the parser. So everything still works for client side parsing, but if just using ractive runtime than there are no Function constructor calls.

At that point, wouldn't we be able to just make that the default behavior?

src/shared/getFunction.js
fskreuz: Exactly my point. The functions are still strings.

fskreuz:
So if I understand correctly now, to avoid `Function`, one must precompile the templates beforehand and the pre-compiled expressions are already functions, not just the `_0+_1` that need to be revived by `Function`. If so, sounds like a plan.

evs-chris:
One of the things I was going to try out was a standalone precompiler that took a regular old template and added `fn: function(...) { ... }` to all of the expression nodes. Then a simple check for an `fn` key would let Ractive know whether or not to create a new function.

src/shared/getFunction.js
martypdx: I think @MartinKolarik has it covered, we just need to (re)move this last line, or move it into the parser. Trying it out now...

martypdx:
> One of the things I was going to try out was a standalone precompiler that took a regular old template and added fn: function(...) { ... } to all of the expression nodes.

I like @MartinKolarik's approach of coalescing all the functions to the template root and looking them up by key. But maybe `fn` would be better name for node?

I moved the `new Function` out of `getFunction` and instead added into to parser. Once I turned the functionality on globally, ran into some issues like `[[mustache]]` causing a function to be created (I beleive `r` uses `s` for static `true`).

src/shared/getFunction.js
MartinKolarik: @martypdx no need to move anything, if functions are in the template, we don't get past 4th line. New function is created only if there isn't one (parsing without `insertExpressions`).

MartinKolarik:
>I think we need to store the function itself, not the string representation as template.e.

That's what actually happens, though I admit the test may suggest otherwise - I was just trying to find a way to compare the functions for testing, since references will obviously differ.

>I beleive r uses s for static true

I added an `isArray()` check for `r`, and I believe this covers all cases. At first, I tried to use a more sophisticated function for finding expressions, but that got too complex too quickly, so I thought looking for the right keys anywhere in the templates might be a better idea. What do you think?

JonDum:
> That's what actually happens, though I admit the test may suggest otherwise - I was just trying to find a way to compare the functions for testing, since references will obviously differ.

@MartinKolarik maybe this for the test?
```js
var a = function (_0,_1) { return _0 + _1 }
Object.toString.call(a)
< "function (_0,_1) { return _0 + _1 }"
```

MartinKolarik:
@JonDum that's what I did initially, but it didn't work when running tests in phantom, so I changed that to just checking keys on `parsed.e`

martypdx:
I moved the createFunction functionality completely into the parser and made pre-parsing functions the default behavior. All the tests still pass _except_ the expression parse tests which are testing the parsed result.

Before I fix those, I wanted to see if there was any reason _not_ to make this the default behavior. Also, I think we can write the functions as we go rather than post-processing them.

There's also some additional work to pre-write the computed property string.

martypdx:
> I added an isArray() check for r, and I believe this covers all cases. At first, I tried to use a more sophisticated function for finding expressions, but that got too complex too quickly, so I thought looking for the right keys anywhere in the templates might be a better idea. What do you think?

@MartinKolarik working through some of the edge cases I changed the function to look explicitly for `.x` keys.

MartinKolarik:
>Before I fix those, I wanted to see if there was any reason not to make this the default behavior. 

Increasing size of parsed templates for no real benefit (if you don't need this on).

>@MartinKolarik working through some of the edge cases I changed the function to look explicitly for .x keys.

It currently fails for `{{foo.bar[x + 6].baz}}` and `<a on-focus="method(x + 5)"></a>`

>There's also some additional work to pre-write the computed property string.

Computed properties are defined during run time, so there isn't much we can do about them, is there?

JonDum:
> Before I fix those, I wanted to see if there was any reason not to make this the default behavior. Also, I think we can write the functions as we go rather than post-processing them.

@martypdx So the only downside is extra Kb in the template(s), but you save cpu cycles and battery and the JIT compiler *probably* has an easier time with these functions than with `new Function()`?

I'd have to do some testing to be sure, but I'd imagine it's such a small difference that it doesn't matter too much on cold load times. Especially if they're already bundled up in a request for the `js` instead of an additional request. 

Also, we'd avoid any future CSP issues that may arise in new version of browsers/iOS.

So :+1: for default from me.

MartinKolarik:
I think the performance gains are just as negligible as the additional KBs, so I'm fine with whatever options you guys think is better. If we come to a conclusion that those extra KBs are insignificant compared to the rest of template, then we can even not make this an option, and simply always output all functions.

src/shared/getFunction.js
```diff
@@ -0,0 +1,10 @@
+import parser from '../Ractive/config/custom/template/parser';
+
+export let functions = {};
+
+export default function getFunction ( str, i ) {
+	if ( functions[ str ] ) return functions[ str ];
+
+	// Could we actually end up here???
+	return parser.createFunction( str, i );
```
MartinKolarik: If we leave this in, we'll need to cache those new functions.

evs-chris:
I don't know if it's required anywhere, but I think there is something to be said for being able to have strict-JSON compatible templates.

martypdx:
> I think the performance gains are just as negligible as the additional KBs

I think @MartinKolarik is probably right on the negligible nature of this. I went back to @Rich-Harris' original [perf investigation](https://gist.github.com/Rich-Harris/f3d63e4ef4d6680d3052) and while the mystery of the slighty faster _running_ generated function remains (at least on chrome), I measured the actual time to _create_ the generate function and it seems to run around .075 - .1 ms. So if you had 100 expressions, you might save 10ms on render. :D

I imagine anyone concerned about file size would be zipping, so I suspect the file sizes would be very similar to non-functions.

> I don't know if it's required anywhere, but I think there is something to be said for being able to have strict-JSON compatible templates.

Just to recap, without the precompiled functions, as long as the parser is available the expressions can be parsed on the client. So maybe we add a parser option of `csp: [true|false]` to control generating the functions. If set to false, it would require the parser version of Ractive on the client.

> It currently fails for {{foo.bar[x + 6].baz}} and <a on-focus="method(x + 5)"></a>

Hey, I borked that :)

> > There's also some additional work to pre-write the computed property string.
> Computed properties are defined during run time, so there isn't much we can do about them, is there?

Yeah, not entirely sure what best course is. One course is to require parser if you're going to use them. Though here is that parser === non-csp. You could add posted-processing of component build process that translates them. Though we'd have to expose that through parser somehow.

MartinKolarik:
>Yeah, not entirely sure what best course is. One course is to require parser if you're going to use them. Though here is that parser === non-csp. You could add posted-processing of component build process that translates them. Though we'd have to expose that through parser somehow.

If you use CSP, you can just use real functions instead of strings. If you don't use CSP, nothing changes, and those functions will be created on run time. They are created by different function, so no parser is necessary.

Only possible problem I see is that someone might get confused, because setting `csp: true` won't (can't) do anything about those functions.

Handling this during build process *is* a way, but... is it worth it for saving those few characters? I personally declare all computed properties as real functions.

MartinKolarik:
@martypdx are we good to merge this?

martypdx:
@MartinKolarik I've got a few changes to push that allow the computation signatures to work without the parser, but after thinking about it I think that deploying the ractive runtime should _always_ be csp compliant. Given that security is part technical and part perception, I think it would be better to know that using the non-parser runtime means having no `new Function` or `eval`.

I've added a method `Ractive.parse.computedStrings` that would make it a one-liner to update tooling like `rcu` to be able to convert computation strings:

```js
parse.computedStrings = function( computed ) {
	if ( !computed ) return [];

	Object.keys( computed ).forEach( key => {
		const value = computed[ key ];
		if ( typeof value === 'string' ) {
			computed[ key ] = getFunctionFromString( value );
		}
	});
};
```

So upside is no `csp` option to manage and including only the runtime is "guaranteed" no eval. (Both these things make the code simpler too).
 
The downside of this being that using the runtime-only with computation strings flat out won't work, but I bet 99.9% percent of people in the situation are using tooling to package components - and the burden will be to add the conversion line to the tooling.

Thoughts?

(@evs-chris: `JSON.stringify` omits function properties so AFAICT they would just be removed at that point. They templates would be functional assuming the parser was included with the runtime.)

Issue #2291, opened 2015-11-28
Can't place some utf symbols in the template


tovarishStepanov:
For example &#128527; (&#128527); looks like ?, however &#9824; (&#9824) displays properly;

fskreuz:
[Definitely something funny going on here](http://jsfiddle.net/h163v7gu/1/). Went to check if this was a bug or wasn't really supported. The funny thing is [it works on v0.4](http://jsfiddle.net/h163v7gu/2/). It broke v0.5 onwards, including edge.

```js
var instance = new Ractive({
    el: document.body,
    append: true,
    template: '<div>Hello &#128527; World &#9824; People</div>',
});

console.log(Ractive.parse('Hello &#128527; World &#9824; People'));
// {"v":3,"t":["Hello � World ♠ People"]}
console.log(instance.toHTML());
// <div>Hello � World ♠ People</div>
```

PR #2293, opened 2015-11-30
Add support for unicode characters in the SMP and SIP

evs-chris:
As discovered in #2291, Ractive doesn't currently allow any characters outside of the BMP because `String.fromCharCode` doesn't support anything outside the BMP. Things are getting better though, so this checks to see if `String.fromCodePoint` is available, and if it is, allows characters from the SMP and SIP as well.

I don't know if it would be better to wholesale allow the other planes, as they are mostly undefined/private(?) beyond the SIP.

Also, I didn't add a test here because it seems straightforward enough and I don't want to tempt the wrath of the elder browsers and engines by including a unicode character from beyond the reasonable scripts. If this PR is acceptable but needs a test, I can take a swing at it. Suggestions or pushes are also more than welcome.

For reference: https://en.wikipedia.org/wiki/Plane_(Unicode)


Rich-Harris:
:+1:

Issue #1990, opened 2015-06-11
`onchange` triggered for root instance but not for components


dagnelies:
Hi,

dunno if it's a bug or a feat, but `onchange` does not trigger for values mapped to a parent instance:

http://jsfiddle.net/o843futo/

it'd be nice to mention this in the docs or alter the behavior so that onchange is always called.

this is probably related to #1978

Cheers

fskreuz:
Definitely a bug. Here's a 0.6 fiddle http://jsfiddle.net/o843futo/1/ which works.

martypdx:
IMO it's a nofix by design. With the strong data ownership introduced in `0.7` `onchange` only fires for data owned by the instance. 

Observe works across all data accessible by the component. Is there a use for `change` beyond what observers can do?

dagnelies:
@martypdx Let's say it like this: it's very conter-intuitive.

If it's meant to be that way, all right, but it should at least be documented with a big warning. Otherwise, people will go nuts trying to understand why their onchange isn't triggered, like me. 

...but ideally, I'd call both, just for the sake of being what's natural/expected/intuitive.

Madgvox:
I have to agree that it is a bit counter-intuitive. It makes sense for a component to "own" any data that passes through it. It doesn't have to actually own the data, as per `0.7`, but it should at least know about it. Basically, if you're able to touch the data, it should be able to touch back.

fskreuz:
I agree with the rest. :+1:

dagnelies:
Really a pity this is closed.

From a design perspective, it's really nonsense. If a value of a component changes, it should trigger `onchange`. That's it. Not depending on who maybe has the ownership.

@martypdx mentioned:

> Observe works across all data accessible by the component.

What I also don't get is: if observers could catch all data, why couldn't `onchange` as well? Why do we differentiate a `observe('*',...)` and on `onchange`.

Btw, the `observe('*',...)` never worked for components anyway ( http://jsfiddle.net/mku42tuy/ )

Madgvox:
I closed this due to the lack of continued discussion and the `wontfix` tag. If there's still discussion to be had (and it appears there is!) then I will gladly reopen the thread.

dagnelies:
This is getting worse. Now an isolated component scope change gets triggered in the root scope instead!

martypdx:
@dagnelies take a look at #2319 which fixes all the issues with using observe in lieu of change events and let me know what you think.

[...]

Issue #1978, opened 2015-05-28
0.7.3 Pattern observers not working in components for values mapped to parent data model


frenchroasted:
After upgrading from 0.6.x to 0.7.x, I noticed some differences in my app. I've boiled it down to a difference between observing someArray.* in 0.6.x and 0.7.x. I've made a couple of jsFiddles to illustrate the difference. The fiddles are the same except for the included ractive library, but they exhibit different behavior.

[Ractive 0.6.1 example](http://jsfiddle.net/frenchroasted/x9tunmd6/)
[Ractive 0.7.3 example](http://jsfiddle.net/frenchroasted/vn984r2L/)

In 0.7.3, the only observer that triggers is the statically-defined one (e.g. myArray.0.propName). The wildcard observer does not execute (e.g. myArray.*.propName). I didn't get very far in my debugging due to the complexity of the keypath/dep/mapping code, but I know the observers in my app are being defined and I can see the mappings back to the original data model. However, the observer execution just never happens.

For what it's worth, I also evaluated with "Edge" using jsFiddle's "Frameworks and Extensions" built-in, and Edge shows the same behavior as 0.7.3.

frenchroasted:
I played with the jsFiddles a bit more and found that the pattern observers are equally blind to changes in objects, so it's not limited to array observation. I found "obj.*.a" was a working observer in 0.6.1, but it did not recognize changes to the object in 0.7.3.

martypdx:
Thanks for the fiddles! Looks like it's related to data over the component boundary.

frenchroasted:
You're right, my original title could have been more clear and accurate. I just updated the issue label. My temporary workaround is to iterate through indices/properties and set static observers on each, but that solution is... how to put it? Less observant than the pattern syntax. It may not catch all the different kinds of modifications on the object that I might care about. Given all the other fixes and improvements since 0.6.x, it seems a shame to revert to that older version because of this.

Madgvox:
Here is another repro found by @dagnelies from #1988:

>  Hi,

> Here is the fiddle with `observe("*", ...)`:

> http://jsfiddle.net/mku42tuy/

> And here is the fiddle with `observe("value", ...)`:

> http://jsfiddle.net/mku42tuy/2/

> it results in the same for both latest and edge.

dagnelies:
Just as a comment, I tried to debug/investigate it a little but failed miserably. Understanding even the basics of how it's working is quite tough.

[...]

PR #2296, opened 2015-12-02
fix root * pattern observers on mapped data (#1978)

martypdx:
Mapped data now correctly fires the pattern observer (see #1978, #1990):

```js
const widget = Ractive.extend({
	oninit () {
		this.observe( '*', v => t.equal( v, expect ) );
	}
});

const r = new Ractive({
	el: fixture,
	template: `<widget value='{{foo}}'/>`,
	data: {
		foo: 'foo'
	},
	components: { widget }
});
```

~~Still an outstanding broader issue that keypaths are wrong, so mark this __NOT READY FOR MERGE__~~


heavyk:
:+1:

Issue #2298, opened 2015-12-02
Two-way binding doesn't work straightforwardly inside partials


wmcmurray:
I've seen a lot of issues about two-way bindings inside partials but none exactly like what I'm trying to achieve *(I believe)* :

In this [simplified fiddle](http://jsfiddle.net/1mtz3ey3/1/), I've made a partial used to display a `<select>` with options. When I change the value of this generated `<select>`, the value in the context above **doesn't get updated** while the other way around works (changing the value in the context above **changes** the value in the generated `<select>`).

At this point, I would think that the partials are like an other context without bindings outside of it **BUT** the variable printed in the partial **doesn't** seems to follow the value of the `<select>`. Isn't that stange ?

I understand that I could do this to solve the issue:

```
{{#with food as value}}
    {{> select}}
{{/with}}
```

...but it feels strange to me.

I expect partials to be **self-contained** (devoid of external aliases to make them work) and able to bind **straightforwardly** with their parent context.

Dunno if it's easy to implement but...

![](http://www.quickmeme.com/img/46/46b35216e5fc5bc36b9b718384ee8f0b80311fc74feda5fb1b850a1b977bbf49.jpg)

Thank you :)

hstonge:
+1

martypdx:
Yes, I agree this should work because `{{>partial [context] }}` gets normalized to `{{#with [context[}}{{>partial}}{{/with}}`, so we just overlooked the new alias case of `context as alias`.

BTW - you can also pass a context that you use as `this` in the partial, see http://jsfiddle.net/fx5ayh1k/:

```handlebars
   {{#partial select}}
    <select value="{{this}}">
      <option value="cookie">Cookie</option>
      <option value="cake">Cake</option>
      <option value="pasta">Pasta</option>
      <option value="pizza">Pizza</option>
    </select>
    <p>The value inside partial : <span class="right">{{this}}</span></p>
   {{/partial}}


   {{>select food }}

wmcmurray:
Ohhh well done ! :+1: 

I haven't tested this case because I wanted to add more functionalities to my partial.
ex: adding an empty option `<option value="">-</option>` at the begining like so:

`{{>select {value: food, empty: true}}}`

...resulting in a new bug discovery :smile:

evs-chris:
[...]

PR #2299, opened 2015-12-02
Support for aliasing shortcut for partials - part of #2298

evs-chris:
This adds support for `{{>foo bar as bat, bip as bop}}` -> `{{#with bar as bat, bip as bop}}{{>foo}}{{/with}}` like the existing context shortcut for partials. Any objections?


martypdx:
:grin:

PR #2304, opened 2015-12-06
use locally-defined namespaces where applicable

Rich-Harris:
See #2038. This allows the use of funky namespaces in tags like `<v:documentProperties>`.

I'm not *certain* that it's the correct solution. If you have a tag like that in normal HTML it seems to just inherit the parent `namespaceURI`, and the element name has the prefix contained within it: http://jsfiddle.net/vmsegykr/ Does that mean we should also just ignore the namespace prefix?


Rich-Harris:
Closing in favour of #2305

PR #2307, opened 2015-12-06
Remove test warnings

Rich-Harris:
When running the test suite you get a console full of warnings. It's annoying. They should either be suppressed, or considered part of the test. This PR does that

### DIFF
--- a/src/Ractive/prototype/resetTemplate.js
+++ b/src/Ractive/prototype/resetTemplate.js
@@ -1,4 +1,5 @@
 import { default as templateConfigurator } from '../config/custom/template';
+import { createDocumentFragment } from '../../utils/dom';
 import Fragment from '../../view/Fragment';
 
 // TODO should resetTemplate be asynchronous? i.e. should it be a case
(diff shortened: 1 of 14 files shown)


martypdx:
Awesome!

I do wonder if we could do something like:

```js

test.noWarn = function( descr, t ) {
    onWarn( () => {} ); // suppress
    this( descr, t );
}

// then use it like
test.noWarn( 'Double teardown is handled gracefully (#1218)', t => {
 	...
}
```

Rich-Harris:
Ah, nice idea. Or maybe

```js
t.suppressWarnings();
t.onWarn( msg => ... );
```

That way we could still import `test` directly from `qunit` (otherwise we'd have to be careful about where the augmentation took place), and we're already adding to the `t` object (`htmlEqual`)

martypdx:
> Ah, nice idea. Or maybe
> 
> > t.suppressWarnings();
> > t.onWarn( msg => ... );

Yep, even better. :D

PR #2308, opened 2015-12-07
Allow keywords as references

### LINKED ISSUE #1934: Ractive 0.7 doesn't allow subsequent output?
I just upgraded from 0.6 to 0.7. I got this error:
```
"Uncaught ParseError: Expected closing delimiter '}}' after reference at line 22 character 40:
<div class="message-item {{ type }} {{ new }}">
```
I changed nothing with the template. But apparently having two mustaches in a row fails in 0.7.2?

I reset to 0.6.1 and no longer have the error.

Rich-Harris:
This makes it possible to have things like `{{new}}`, since sometimes you don't control the shape of your data. Good idea? Bad idea? Would fix #1934

### DIFF
--- a/src/parse/converters/readExpressionOrReference.js
+++ b/src/parse/converters/readExpressionOrReference.js
@@ -1,3 +1,4 @@
+import { REFERENCE } from '../../config/types';
 import readExpression from './readExpression';
 import readReference from './expressions/primary/readReference';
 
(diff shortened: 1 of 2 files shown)


Madgvox:
:+1: from me. No reason not to.

martypdx:
js allows it as a property name:

```js
foo = {
	while: 'while',
	do: 'do'
};
foo.new = 'new';
foo.if = 'if';

console.log(foo);
```

So as long as there's no ambiguity I guess I'm ok.

Issue #1919, opened 2015-04-14
Can't find Error Event


Crisfole:
Is there an error event I can hook [rollbar](https://rollbar.com/) or any other global error handler up to?  Right now I'm not getting any feedback when an error reaches the console in production.

martypdx:
@Crisfole a specific example would be helpful so I can get my head around which class of errors you'd want to propagate.

Crisfole:
I want to propagate anything that gets triggered on a top level Ractive (basically: if it has no parent).

When I run:

    window.ractive = new Ractive({ template: "#template", el: "body" });

or whatever, I'd like to be able to say:

    window.ractive.on('error', function(errorInfo) {
        rollbar.reporterrror(errorInfo); // made up syntax, but it makes the point.
    })

martypdx:
Oh, add it as a event hook. That would actually be doable. :)

martypdx:
With current event bubbling, what you wrote would work on top-level, but you could also do:

```js
ractive.on('error *.error', function() {
    // all instance errors including components
})

Crisfole:
So it already exists?

martypdx:
@Crisfole no, I marked as enhancement. The eventing and bubbling infrastructure already exists. 

Would just need to add `this.fire( 'error', errorInfo )` at the right places. In the `log` module [here](https://github.com/ractivejs/ractive/blob/dev/src/utils/log.js#L83-L86) would be one place, probably some others.

Crisfole:
Would this include things like 'printWarning' and 'printWarningOnce'?

martypdx:
> Would this include things like 'printWarning' and 'printWarningOnce'?

I wouldn't think so, what's your thinking?

Crisfole:
A lot of loggers, like Rollbar in particular, but also nLog, log4j, etc. etc. include the ability to log at different levels, including warnings.

It might make sense to either a) fire error and add a level parameter, or b) fire 'warn' for those.

MartinKolarik:
I think it makes sense to have to different events for errors and warnings.

pateketrueke:
Is there any progress on this?

PR #2313, opened 2015-12-08
Adding error catch to component Hooks

ISNIT0:
See:https://github.com/ractivejs/ractive/issues/1919
This will fire an error event to a component that fails to execute oninit, oncomplete, etc.

It also allows for the rest of the component to continue functioning, instead of breaking completely.

This may not be the best place to put the change, but here at least it will stop components dying.

Thoughts?

### DIFF
--- a/src/Ractive/prototype/shared/hooks/Hook.js
+++ b/src/Ractive/prototype/shared/hooks/Hook.js
@@ -32,7 +32,12 @@ export default class Hook {
 	fire ( ractive, arg ) {
 		function call ( method ) {
 			if ( ractive[ method ] ) {
-				arg ? ractive[ method ]( arg ) : ractive[ method ]();
+				try {
+          arg ? ractive[ method ]( arg ) : ractive[ method ]();
+        } catch (e) {
+          ractive.fire('error', e);
+          console.error(e);
+        };
 				return true;
 			}
 		}


src/Ractive/prototype/shared/hooks/Hook.js
```diff
@@ -32,7 +32,12 @@ export default class Hook {
 	fire ( ractive, arg ) {
 		function call ( method ) {
 			if ( ractive[ method ] ) {
-				arg ? ractive[ method ]( arg ) : ractive[ method ]();
+				try {
+          arg ? ractive[ method ]( arg ) : ractive[ method ]();
+        } catch (e) {
+          ractive.fire('error', e);
+          console.error(e);
```
evs-chris: I would use logging from src/util/log. Perhaps an error method is in order. I think it may be safer to let it die here than possibly get in an undefined state further on, though I'm often wrong.

martypdx:
> It also allows for the rest of the component to continue functioning, instead of breaking completely.
> 
> This may not be the best place to put the change, but here at least it will stop components dying.
> 
> Thoughts?

I don't know that I would want things to continue if I had an error if there were a lifecycle event. 

#1919 is just about propagating ractive thrown errors and warnings through an `error` event.

Can you explain in more detail what you're looking for?

PR #2319, opened 2015-12-13
Better, 'onchange' worthy observers at root/wildcard *

martypdx:
Follow-on to #2296, I started testing cases around #1990 being able to use observers faithfully for component root and `*` wildcard I found a number of issues around root observers and `'*'` observers:

* Template expressions with internal `@(..` keypaths are no longer included 
* Both `*` and root observers use correct component-relative keypaths
* Adding new properties works correctly


Rich-Harris:
Nice!

PR #2324, opened 2015-12-18
Update ConditionalAttribute.js for Performance

benawhite:
if str is empty, let's not waist time manipulating dom elements


Rich-Harris:
:+1: makes sense! Thanks

Issue #2123, opened 2015-09-05
Setting children of computed values


Rich-Harris:
Related to #2122: if `obj` is a computed value with a `foo` property, setting `obj.foo` should cause an error if `obj` does not have a setter function.

If `obj` _does_ have a setter, there are two possibilities:
1. an error – you should set `obj` directly
2. the setter should be called with the current value, except with the `foo` property overwritten

Of those I think I favour 1 – there's no way of knowing whether it's safe to mutate the value of `obj` from outside the computation.

Rich-Harris:
Implemented option 1 as 732a07c1e7b926d272e31379ba01238459a6b187 on #2122

martypdx:
Seems artificially restrictive to me, what problem is this trying to solve?

If I understand correctly, it means if I have a filtered list, I can no longer edit the items of that list:

```html
{{#each filtered(items)}}
    <li><input value='{{this.name}}'>
{{/each}}
```

Rich-Harris:
That already fails – it just does so without warning! http://jsfiddle.net/rich_harris/by1kmhxt/. To do that you'd need to iterate over `items` then wrap the input in an `{{#if filter(this)}}` block.

Rich-Harris:
Here's the warning you'd get with #2122:

<img width="775" alt="screen shot 2015-09-05 at 10 52 58 am" src="https://cloud.githubusercontent.com/assets/1162160/9699776/577f7878-53bc-11e5-9f0a-5caa0801aeb0.png">

martypdx:
I understand we have the limit of not being able to map items of `items` to corresponding `filtered(items)` (someday :) ), but there are times when that limitation is fine and you just want to operate on the data tree under the computation.

I think a warning would be ok, but not an error (looks like this is the case from screenshot).  But I think the guidance is wrong, it should just warn that the computation will not notify equivalent properties of it's dependencies.

Rich-Harris:
It *does* error if you attempt `ractive.set('computed.child', 'foo')`, the warning is only for attempted two-way binding. But it turns out the error causes a problem with a particular component edge case, so I've disabled that in #2124 anyway, at least temporarily :)

> there are times when that limitation is fine and you just want to operate on the data tree under the computation

[...]

PR #2331, opened 2016-01-06
Attempt to allow opt-in computation child binding - RFC

evs-chris:
__Pending feedback__

There have been a number of related issues popping up around #2123 as people start trying out 0.8 because expression binding from < 0.8 does not work in 0.8. @martypdx suggested that perhaps we could add an _opt-in_ flag to re-enable said behaviour at about the same time I was looking into the same, so it must be a good idea right? :smile: This is my attempt to do so and perhaps improve upon what was available in earlier versions of Ractive.

I think this is a pretty straightforward patch, but I'm not sure if I've thought about it from all of the various cutting angles. It checks a new flag, `computationChildBinding` for truthiness in order to drop the `isReadonly` constraint from `ComputationChild` instances. It also adds a constructor to make sure that `get` is called on the parent model to supply the `ComputationChild` value. These two things are enough to get dependents of a single expression to update together. For instance `{{#each fn()}}<input value="{{.}}" />{{.}}{{/each}}` will keep both of the `.` refs in sync, but if there is anything that uses `fn()` anywhere else, those are left out of sync.

To resolve that last little issue - I figured that any time you're trying to bind to a child of an expression, you're probably going to be updating one of the models on which the expression depends. So, this looks up the computation when `applyValue` is called and marks each of its dependencies. The result, while probably a bit inefficient, is that now any computation child that modifies its captured deps will cause other computations and fragments with the same deps to update.

Thoughts?


martypdx:
> To resolve that last little issue - I figured that any time you're trying to bind to a child of an expression, you're probably going to be updating one of the models on which the expression depends. So, this looks up the computation when applyValue is called and marks each of its dependencies. The result, while probably a bit inefficient, is that now any computation child that modifies its captured deps will cause other computations and fragments with the same deps to update.
> 
> Thoughts?

Unsure about this the applyValue computation dependency part. If it actually works, and performs well, then we don't really need to restrict binding on computations, right?

evs-chris:
I don't know about performance, but it does work in cases such as:

```js
let template = '{{#each expr()}}{{.val}}: <input value="{{.val}}" /><br/>{{/each}}  {{JSON.stringify(expr())}}';
var r = new Ractive({
  template,
  el: '#some-el',
  data: {
    foo: [ { val: 'a' }, { val: 'b' }, { val: 'c' } ]
  },
  computationChildBinding: true
});

r.set('expr', function() { return r.get('foo'); });
```

Leaving out the `applyValue` part still leaves everything but the `JSON.stringify(expr())` interpolator working as expected.

My gut says that in most cases performance would be acceptable, especially if we leave it behind the flag as a possible sharp edge. I'll see if I can come up with a decent performance test.

JonDum:
I think the applyValue thing is exactly what I would expect to happen :+1:.
So this is super cool. I'll try out your branch and let you know how it
works.

martypdx:
Oh seesh, I got the "dependencies" mixed up, what you're doing makes total sense. 

What I thought you were doing, and might be worth investigating, is to call update on the computations _derivation_ dependencies. In your example `expr` captures `foo` as a derived dependency, so updating `0.val` would cause equivalent of `ractive.update('foo')` to be called.

In other words, when you change an expression child value, have all the inputs to the computation check their values.

Also, I assume you don't need to guard calling `applyValues` when `computationChildBinding` is false because it throws anyway and `applyValues` doesn't get called?

martypdx:
As to the option name, it's tough because under the covers both expressions and computed properties are mostly same thing. Choosing one or the other may be misleading because it applies to both.

What about `allowedDerivedBindings` or just `derivedBindings`? It doesn't use either term :), but maybe that's a good thing.

evs-chris:
> so updating `0.val` would cause equivalent of `ractive.update('foo')` to be called

That's what this is doing, unless I've misunderstood you. I tried just marking the computation (`deps` as opposed to `dependencies`... slightly confusing :grinning:) to begin with, but since computations aren't shared (and even if they are, different signature would have different computations) it wasn't affecting related computations.

> Also, I assume you don't need to guard calling applyValues when computationChildBinding is false because it throws anyway and applyValues doesn't get called?

Yes. At that point, the model is read only, so nothing (from userland, anyways) makes it to `applyValue`.

> As to the option name

Yeah, the more I look at `computationChildBinding` the less I like it. `derivedBindings` is feeling pretty good :smile:. I'll push that through with the tests.

@JonDum I haven't had a chance to try this on my biggest, nastiest project yet. I'd love to know how it worked out for you.

JonDum:
Tested this out a bit. Definitely... wonkey. I tried it out on app that has a bunch of divs you can drag around with the mouse. They all share the same component, but for whatever reason sometimes it just wouldn't work on one of them, randomly. It's going to be hard to isolate since it's on such a large app, but I'll see if I can dig in more.

JonDum:
Oh oh oh actually I just noticed something that may help. The one it doesn't work on happens to be the first item in the array.

```
{{#each blocks}}
  <block ../>  // all sets fail on the 0 index of this 
{{/each}}
```

evs-chris:
@JonDum thanks for checking. I played around with components and expressions some and haven't been able to get it to misbehave. Is the expression inside or outside the component?

JonDum:
@evs-chris 

http://plnkr.co/edit/h6a0RMhjP8qHWtJk5ZLE?p=preview

Couldn't repro it ugh :( :( :(

However, I did notice that changes in the "computed" list don't get reflected in the "direct", but they do vice versa. That may be a bug with the `applyValue`?

evs-chris:
@JonDum yep that was a bug with applyValue not accounting for computed properties. I had only been using expressions up to this point. It should now handle computed data

guilhermeaiolfi:
About the flag name. It seems that something like ```twoWayDerived``` would be aligned with what we already have. Or ```syncDerived```. Although ```derived``` doesn't seem like a good name (too technical). ```syncIndirectBindings```?

martypdx:
>  The one it doesn't work on happens to be the first item in the array.

Hmm, sometimes that means js falsely bug like `!key` where key is part of `items.0`.

JonDum:
Tried this out again. The previous issue has been fixed, but that only revealed another bug :)

`Uncaught TypeError: Cannot read property 'radioname-bindingGroup' of undefined`

Looks like it is not liking two way binding from a derived computation.

Here's an isolated plnkr: http://plnkr.co/edit/d8bLoCTbqg88osD43eOl

JonDum:
Also, getting that ol' `model.shuffle is not a function` error again when using an `makeArrayMethod` method on a computed array.

http://plnkr.co/edit/oP002jtEOAP71dilms4N?p=preview

evs-chris:
@JonDum I think I've got the name binding thing sorted.

The `model.shuffle` error in your example is in fact illegal because it would be modifying the computation (blocks) directly. I don't see a clean way handle that at the moment, so I added a check to throw if the model shuffling is readonly. If the `blocks` computed returned an object with an array member, the array member could be spliced e.g. `ractive.splice('blocks.list', 0, 1, 'yep')`.

evs-chris:
@JonDum I tried to get the reference expression PR to die with the error you mentioned in #2409 and #2174 (? :laughing:). I couldn't get it to die with a double teardown, so I suspect that it probably is something caused over here. Can you give me an example to trace through? I have ractive-select running locally, so something based on, or that I could throw in next to, the demo would be great.

JonDum:
I haven't been able to isolate it... _yet_. The big app it's happening on is currently experiencing a torrent of infinite loops with this branch and I've spent most of today trying to figure _that_ out so I could then figure out that double teardown. It's proving very difficult to determine why the infinite loop is occurring — can't make any timeline/js method profiles and pausing the debugger doesn't seem to work. I did manage to get that tab up to 4GB of ram though before it died, which is pretty neat I guess haha.

JonDum:
On other thing I've noticed, is that I think this comp-bind stuff is _**slow**_. Like really slow. Noticeably laggy and jerky compared to 0.7.3. It's likely something I'm doing wrong, yet only a few things have changed in upgrading to 0.8—0.7.3 was much more performant.

evs-chris:
Hmmm... have you tried a build within the last week? You probably have but I thought I'd ask since there were a few exponential recursion scenarios that were resolved. If not, I think you're hitting the speculated performance issues with notifying all of the computation's dependencies to update along with the computation. If you comment out line 31 of ComputationChild.js does it speed up?

JonDum:
Commenting that out didn't help. I think I know why it's so laggy though.

I got this with a single 'set' from the hover decorator:
screen shot 2016-03-01 at 1 33 11 pm

It should be ending in the first fat column (set  > set > applyValue > mark...), but it goes on to dispatch a bunch of observers which was strange for this keypath. After looking at it further I realized that this one `set` is also dispatching every single observer that is also on a computed keypath, even if they are totally different keypaths. Gonna keep investigating why/how.

evs-chris:
Superceded! #2722

PR #2344, opened 2016-01-13
Make sure nodes queued to detach after transition has started actually get detached - #2310

### LINKED ISSUE #2310: Ractive 0.8 not unrendering fragments.
http://jondum.github.io/ractive-select/demo/edge.html

Really odd one here. If you select an item from one of the dropdowns, the `<span>` for the checkmark of the previously selected item stays behind. If you keep selecting items you end up with several checkmarks left over. The data model is correct (`main.findAllComponents('select')[0].get('_items').map(i => i.selected)`), so it's got to be an issue with the fragment not getting cleaned up?

Relevant template section:
```html
    <ul class='dropdown{{#if open}} open{{/if}} {{class}}' on-click='select(event)'>
        {{#each _items}}
        <li value='{{.value}}' {{#if selected}}selected{{/if}}>
                {{#if selected}}<span class='checkmark'></span>{{/if}}{{label}}
            </li>
        {{/each}}
    </ul>
```

and how `_items` is set:

```js
    updateItems: function() {

        var select = this.find('select');
        var options = select.querySelectorAll('option'); // the select's options are {{yield}}ed
        var value = this.get('value');
        var attr, label;

        var items = this.get('items');

        var newItems = [];

        if (options && options.length > 0) {

            for (var i = 0, len = options.length; i < len; i++) {
                var opt = options[i];
                attr = opt.getAttribute('value');
                if (attr == value) {
                    label = opt.textContent;
                }
                newItems.push({
                    label: opt.textContent,
                    value: attr,
                    selected: opt.selected
                });
            }

        }

        this.set('label', label);
        this.set('_items', newItems);

        this.updateSize();

    },
```

evs-chris:
This fixes #2310. The issue is that the transition manager starts and completes its outros before the deferred observers fire, and the deferred observers cause nodes to be removed and their transition manager is a child of the already completed original transition manager. So when the second round goes to detach, they never do because the parent transition manager is already done.

This skips the parent queue if the parent is already done and just goes ahead with the detachment. Would it be more appropriate to move the transition manager start to after the deferred observers fire?

@JonDum, can you try this out and let me know if it definitely fixes your problem?

@gaelicmichael, if you have a few spare cycles, could you try this as an alternative for the fixes in #2284?

I can provide test builds if anyone needs them.

### DIFF
--- a/src/global/TransitionManager.js
+++ b/src/global/TransitionManager.js
@@ -74,7 +74,7 @@ function check ( tm ) {
 	// we notify the parent if there is one, otherwise
 	// start detaching nodes
 	if ( !tm.outrosComplete ) {
-		if ( tm.parent ) {
+		if ( tm.parent && !tm.parent.outrosComplete ) {
 			tm.parent.decrementOutros( tm );
 		} else {
 			tm.detachNodes();
--- a/test/browser-tests/plugins/transitions.js
+++ b/test/browser-tests/plugins/transitions.js
@@ -356,3 +356,19 @@ test( 'Conditional sections that become truthy are not rendered if a parent simu
 
 	t.ok( !transitionRan );
 });
+
+test( 'Nodes that are affected by deferred observers should actually get dettached (#2310)', t => {
+	const r = new Ractive({
+		el: fixture,
+		template: `{{#if bar}}<span>baz</span>{{/if}}`,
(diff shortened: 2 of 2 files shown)


JonDum:
Yuppp fixes it! You da man.

Issue #1445, opened 2014-11-05
inline partials doesn't work with partials from script tags


sigod:
```
<script id="template" type="text/ractive">
    {{>partial}}
</script>

<script id="partial" type="text/ractive">
    {{>partial2}}
    
    <!-- {{>partial2}} -->
        Test
    <!-- {{/partial2}} -->
</script>
```
In console: `Ractive.js: Could not find template for partial "partial2"`.

jsfiddle: http://jsfiddle.net/mhd1r1ng/

--
More complex example: http://jsfiddle.net/otew34uj/1/

In console it says: `Uncaught TypeError: Cannot read property 'detach' of null `. But if we remove context expression (e.g. leave just `{{>partial}}`) we'll get no errors at all.

PR #2351, opened 2016-01-21
Let runtime parsed partials contain other top-level partials - #1445

evs-chris:
Per #1445, if you have a partial script with another partial inlined, the inlined partial is lost. This allows the inlined partials to be added to the ractive partials store if there aren't already partials with the same names.

### DIFF
--- a/src/view/items/partial/getPartialTemplate.js
+++ b/src/view/items/partial/getPartialTemplate.js
@@ -14,6 +14,11 @@ export default function getPartialTemplate ( ractive, name, parentFragment ) {
 		// parse and register to this ractive instance
 		let parsed = parser.parseFor( partial, ractive );
 
+		// register extra partials on the ractive instance if they don't already exist
+		for ( let k in parsed.p ) {
+			if ( !( k in ractive.partials ) ) ractive.partials[ k ] = parsed.p[ k ];
+		}
+
 		// register (and return main partial if there are others in the template)
 		return ractive.partials[ name ] = parsed.t;
 	}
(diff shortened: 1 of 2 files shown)


fskreuz:
:+1:

src/view/items/partial/getPartialTemplate.js
```diff
@@ -14,6 +14,11 @@ export default function getPartialTemplate ( ractive, name, parentFragment ) {
 		// parse and register to this ractive instance
 		let parsed = parser.parseFor( partial, ractive );
 
+		// register extra partials on the ractive instance if they don't already exist
+		for ( let k in parsed.p ) {
+			if ( !( k in ractive.partials ) ) ractive.partials[ k ] = parsed.p[ k ];
+		}
+
```
martypdx: Sorry late to the party on feedback, extend/fill-gap functionality already exists [here](https://github.com/ractivejs/ractive/blob/dev/src/utils/object.js#L78-L102).

Though not sure we should be using `for` `in` unless we actually need to go deeper into the prototype. My understanding is that it (for/in) prevents optimization of the function.

evs-chris: I did not know we had a `fillGaps` in util! I've pushed a change that cleans it up a little an (re-)introduces a use of `fillGaps`.

It looks like `Object.getOwnPropertyNames` would be the ideal replacement for the for-in, but it would need to be polyfilled in legacy alongside `Object.keys` for IE8.

Issue #2224, opened 2015-10-15
Unexpected function call (in partial context)


woodlandhunter:
See the attached fiddle. `someFunc` is called (with undefined as the argument) when `foo` is set to `null`

http://jsfiddle.net/j3mjdf0m/

Rich-Harris:
Some breadcrumbs: it turns out the function is being called indirectly as a result of a `clearUnresolveds` call, via [here](https://github.com/ractivejs/ractive/blob/gh-2224/src/model/Model.js#L176). The thing that's unresolved is the name of the partial. It looks like the resolver gets created to try and find a value for `test-partial` even though it's a) a named partial, and b) the name wouldn't be a valid reference.

Have opened a branch with a failing test: https://github.com/ractivejs/ractive/tree/gh-2224

evs-chris:
Looks like the partial with context bit can be more-or-less fixed, but since it's just a `{{#with}}` wrapped around the partial, any other unresolved inside a side-effecting `{{#with}}` is going to have the same issue.

It looks like if we made the expression proxy aware of its rendered state somehow, we could avoid the recomputation since it should go out of scope with the wrapping conditional. I didn't see a quick and easy way to do so though, since those are a sort of hybrid model fragment beastie. My inner-functional-snob say just don't side-effect like that :smile:

At any rate...

```diff
diff --git a/src/view/items/Partial.js b/src/view/items/Partial.js
index b4b4eaf..7c6371e 100644
--- a/src/view/items/Partial.js
+++ b/src/view/items/Partial.js
@@ -5,8 +5,6 @@ import getPartialTemplate from './partial/getPartialTemplate';

 export default class Partial extends Mustache {
        bind () {
-               super.bind();
-
                // keep track of the reference name for future resets
                this.refName = this.template.r;

@@ -16,10 +14,13 @@ export default class Partial extends Mustache {
                if ( template ) {
                        this.named = true;
                        this.setTemplate( this.template.r, template );
-               } else if ( ( !this.model || typeof this.model.get() !== 'string' ) && this.refName ) {
-                       this.setTemplate( this.refName, template );
                } else {
-                       this.setTemplate( this.model.get() );
+                       super.bind();
+                       if ( ( !this.model || typeof this.model.get() !== 'string' ) && this.refName ) {
+                               this.setTemplate( this.refName, template );
+                       } else {
+                               this.setTemplate( this.model.get() );
+                       }
                }

                this.fragment = new Fragment({
```

[...]

PR #2356, opened 2016-01-23
Defer unresolved to after update for expressions - #2224 - RFC

evs-chris:
See #2224. This is one of those chewy places where `ExpressionProxy` acts like a model and a virtual DOM item at the same time. Even though the item side is about to be removed, the model side is updated before that happens and triggers the computation again.

I tried to come at this from a few different angles:

1. Getting the computation to see if its fragment should be rendered by asking up the fragment hierarchy, but that was a big ugly change and was confusing to trace through at best.
2. Avoiding the call to `get` in `has`, but here there's no clear indicator as to when to skip.
  1. Using the cached `value` in `has` if the fragment is dirty. This breaks a few things too.
3. I finally landed on deferring unresolved clear for `ExpressionProxy`s to be deferred until after the round of fragment updates is processed if the fragment is dirty.

Number 3 seems to work fairly well, but I'm not positive it's the best approach. I had to add support for checking to see if there is an active batch to the `runloop` to avoid an infinite recursion scenario. Thoughts?

### DIFF
--- a/src/global/runloop.js
+++ b/src/global/runloop.js
@@ -46,6 +46,8 @@ const runloop = {
 		batch.transitionManager.add( transition );
 	},
 
+	active () { return !!batch; },
+
 	// synchronise node detachments with transition ends
 	detachWhenReady ( thing ) {
 		batch.transitionManager.detachQueue.push( thing );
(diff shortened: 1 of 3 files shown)


heavyk:
in the test, should you count the number of times y it's called, as well?

evs-chris:
It does indirectly with t.expect(2). If the function gets called twice, there are three assertions and the test fails.

heavyk:
clever :)

Issue #2321, opened 2015-12-16
Ractive 0.8 - deleting item from inside a loop that includes conditional statement with reference, crashes


kouts:
I'm sorry for the description, couldn't find a better way to describe it.
Here is the fiddle: http://jsfiddle.net/hhfpsm44/14/

You'll need the console open for this one.
Click on any of the delete buttons and you get:
Ractive.js: Failed to compute @(objects\.3\.characteristics\.indexOf(33051)>-1)?"":"inactive": Cannot read property 'indexOf' of undefined

On firefox you'll get a noticeable delay before the item gets deleted.

This works as expected with Ractive 0.73

danjford:
@kouts, I've just had a quick look and I think the problem is that you're doing an index of even when the item may not be defined. 

I've just added extra parts to your conditional statements so that it will check for the existence of the item before trying to perform an indexOf.

**Example**
```
{{ (characteristics && characteristics.indexOf(code) > -1) ? '' : 'inactive' }}
```

See my updated [jsfiddle](http://jsfiddle.net/hhfpsm44/15/).

kouts:
Ok thanks! But why does it work with Ractive 0.73?
Plus, "characteristics" is always defined in my example...

martypdx:
It's a bug because Ractive shouldn't be recalculating expressions in array members that have been removed.

danjford:
I've had this issue happen to me in a previous version. It may have been 0.7.3, it may have been under different circumstances though, I can't really remember.

evs-chris:
This is now fixed in edge.

kouts:
Thank you!

PR #2359, opened 2016-01-26
Allow expression proxies and object iteration sections to rebind - #2321 - RFC

evs-chris:
There are two bugs showcased in #2321.

~~The first is that computations are currently triggered again during shuffling for elements that should no longer be rendered afterwards. Since expression proxies are also kinda-fragments, I added a way to rebind them too and added a check in mustache that detects a rebindable model (isExpression and has a rebind) and calls it instead of replacing the model. That resolved the issue of calling unresolvable computations for the test case. It also seemed to render the discard checks during shuffle obsolete, so I pulled them out.~~ Yep... caused by the second one...

The second is that object iteration, as opposed to array iteration, was not allowed to rebind because it was only looking for array-like thinks being iterated. Object aren't arrays, so they were being skipped. The simple solution is to always allow rebind in repeated fragments and use the `fragment.key || fragment.index` as key on which to rebind.

Thoughts?

### DIFF
--- a/src/model/Computation.js
+++ b/src/model/Computation.js
@@ -1,3 +1,5 @@
+/* global console */
+
 import { capture, startCapturing, stopCapturing } from '../global/capture';
 import { warnIfDebug } from '../utils/log';
 import Model from './Model';
@@ -49,7 +51,7 @@ export default class Computation extends Model {
 		this.signature = signature;
 
 		this.key = key; // not actually used, but helps with debugging
-		this.isExpression = key && key[0] === '@'
+		this.isExpression = key && key[0] === '@';
 
 		this.isReadonly = !this.signature.setter;
 
@@ -65,7 +67,7 @@ export default class Computation extends Model {
 		this.boundsSensitive = true;
 		this.dirty = true;
 
-		// TODO: computations don't shuffle, but this is a bit hackish
+		// TODO: is there a less hackish way to do this?
 		this.shuffle = undefined;
 	}
 
--- a/src/view/resolvers/ExpressionProxy.js
+++ b/src/view/resolvers/ExpressionProxy.js
@@ -17,12 +17,29 @@ export default class ExpressionProxy extends Model {
 		this.template = template;
 
 		this.isReadonly = true;
-		this.isExpression = true;
 
 		this.fn = getFunction( template.s, template.r.length );
 		this.computation = null;
 
-		this.bind();
+		this.resolvers = [];
+		this.models = this.template.r.map( ( ref, index ) => {
+			const model = resolveReference( this.fragment, ref );
+			let resolver;
+
+			if ( !model ) {
+				resolver = this.fragment.resolve( ref, model => {
+					removeFromArray( this.resolvers, resolver );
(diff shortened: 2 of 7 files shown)


src/model/Computation.js
```diff
@@ -65,7 +67,7 @@ export default class Computation extends Model {
 		this.boundsSensitive = true;
 		this.dirty = true;
 
-		// TODO: computations don't shuffle, but this is a bit hackish
+		// TODO: is there a less hackish way to do this?
 		this.shuffle = undefined;
```
fskreuz:
```js
this.shuffle = void 0;
```
:trollface:

If this eventually becomes a function, a noop/empty function perhaps? That way, one can safely assume it's a callable all the time regardless if it does something or nothing.

martypdx: https://github.com/ractivejs/ractive/blob/dev/src/utils/noop.js

evs-chris: Unfortunately noop isn't an option because the shuffler uses presence of the shuffle method to decide how to proceed.

src/view/resolvers/ExpressionProxy.js
```diff
@@ -110,6 +94,33 @@ export default class ExpressionProxy extends Model {
...
+
+			return model;
+		});
+
+		this.bubble();
+	}
+
+	rebind () {
+		this.unbind();
+		this.bind();
```
martypdx: Under what conditions does the `rebind` for an expression happen? Is it when it is used within a section?

src/model/Computation.js
martypdx: Right, I now remember trying to go that route and running into that issue.

evs-chris:
@martypdx well dang :smile:. I got caught up fixing the first symptom and it turns out it was caused by the other part that I didn't find until after the bandaid was applied. I have ripped the bandaid back off.

src/model/Computation.js
```diff
@@ -1,3 +1,5 @@
+/* global console */
+
```
martypdx: we don't need `global console` here do we?

martypdx:
@evs-chris well now you're removing twice as many lines as you're adding, so :+1:

src/model/Computation.js
evs-chris: I only add those where my in-editor linter is whining because it makes other issues more visible. Should I remove it?

martypdx: i didn't realize this block was down below in the file:

```js
 			if ( console.groupCollapsed ) console.groupCollapsed( '%cshow details', 'color: rgb(82, 140, 224); font-weight: normal; text-decoration: underline;' );
 			const functionBody = prettify( this.signature.getterString );
 			const stack = this.signature.getterUseStack ? '\n\n' + truncateStack( err.stack ) : '';
 			console.error( `${err.name}: ${err.message}\n\n${functionBody}${stack}` );
 			if ( console.groupCollapsed ) console.groupEnd();
```

Since it _is_ there, then I would leave the pragma. though ultimately, that code should move into `log` util.

Issue #2292, opened 2015-11-30
multiple copies of Ractive on a single page causes CSS scoping errors...


Rich-Harris:
...because each copy keeps its own `data-ractive-css` tally

fskreuz:
> multiple copies of Ractive on a single page

Is there a use case for having multiple versions of Ractive on one page? Or is this for the new versions to gracefully catch accidentally putting in 2 copies on the page?

gaelicmichael:
Sorry for the naive question, but does this refer to creating a new instance of Ractive? If so, this needs to be supported -- I have an instance of Ractive running to manage the GUI but create another instance to control a dialog box, for example.

Rich-Harris:
> Is there a use case for having multiple versions of Ractive on one page?

Sort of... my day job often involves creating interactive widgets for [pages like this](http://www.theguardian.com/us-news/2015/dec/01/the-county-kern-county-deadliest-police-killings) – the widgets need to be developed and tested separately, but they all need to live on the same page. Cruel experience has taught us the folly of attempting to load shared libraries from a CDN or vendor bundle – everything is async, we don't control the page itself, we use different AMD loaders in different environments (web vs app etc), mucking around with the loader config causes all sorts of hairy problems that you *don't* want to debug on a deadline, we need fully reproducible builds, etc, etc. If we had more time we could solve some of these issues but the simplest and most reliable solution is simply to bundle each app up separately. 

> does this refer to creating a new instance of Ractive?

No, this is purely about 

```js
<script src='my-bundled-ractive-app.js'></script>
<script src='my-other-bundled-ractive-app.js'></script>
```

– you can have as many top-level instances as you like, and CSS scoping will work as expected.

fskreuz:
CSS from multiple Ractives should now be covered https://github.com/ractivejs/ractive/pull/2362

Issue #1127, opened 2014-08-20
Missing styles when rendering with toHTML()


PhilJ:
I would like to render an initial page (which consists of nested Ractive components) on the server with `toHTML()`. 

Loading the component with `ractive-load` and rendering works fine - however **the output of `toHTML()` lacks any css code** - both the styles of the root component as well the styles of the nested components are not present in the output.

This is fine for providing a server-rendered page with search crawlers, however this page would not work to pre-render Ractive views for users, since no styles would be present.

**What are my options to add the styles of all used components in the server-rendered output when using `toHTML()`?**

Including the styles via `<link rel="stylesheet" ..>` might be an option, but i'd like to avoid that, since this would mean that styles would no longer be present in the ractive component file. Furthermore the client would need to load all component stylesheets seperately.

The styles are present under the `.css` respectively `.components[COMPONENT_NAME].css` properties of the loaded component - so i could get and concat them from there. However the output of `toHTML()` also does not include the `data-rvcguid` properties in the markup to make use of Ractive's css scoping.

 Any ideas? Thanks for your support!

ElliotChong:
I've been thinking through what it might take to get this working on the server-side... Would a possible interim solution be to expose a `toCSS()` method that is outputting the composed CSS? It would then be possible to use this in a server-only parent template that is rendering the `<html><head><body>` tags and encapsulating the "traditional" content / Ractive component.

The other caveat would be that `toHTML()` would also have to start attaching the `data-ractive-css` property, though I'm guessing that should be pretty doable.

@Rich-Harris @martypdx is this a route you guys would be interested in seeing a pull request for?

Rich-Harris:
@ElliotChong sorry, I missed your comment - just turned up here because I'm experiencing the same problem! On my way I was thinking 'it'd probably make sense if we had a `toCSS()` method...

The wrinkle, of course, is that there needs to be some kind of mechanism to ensure that a) styles don't conflict (if you had HTML/CSS rendered by one process on the same page as HTML/CSS rendered by another, they might stomp each other's `data-ractive-css` IDs), and b) they sync up correctly at render time.

I'm not totally sure what the solution looks like, but keen for us to figure out what it is, and yes a pull request would be great!

ElliotChong:
Did a quick implementation of `toCSS()` that exposes how Ractive is currently constructing the CSS string. This is useful right away if your CSS is written in an Object-oriented (aka: manually namespaced) fashion and works with `noCssTransform: true`.

[...]

PR #2362, opened 2016-01-29
Adding toCSS method WIP - RFC

fskreuz:
Exposes the CSS generated by Ractive via a new toCSS method, plus a few cleanups and tests here and there. Since there wasn't a PR (just a commit on some fork), decided to create one.

ToDo:

- [x] Render CSS  https://github.com/ractivejs/ractive/issues/1127
- [x] Multiple copies of Ractive generating CSS https://github.com/ractivejs/ractive/issues/2292
- [x] Warn when using `css` option for instances https://github.com/ractivejs/ractive/issues/2365
- [x] `toCSS` to trace through instance components only https://github.com/ractivejs/ractive/pull/2362#issuecomment-179639832

### DIFF
--- a/src/global/css.js
+++ b/src/global/css.js
@@ -1,62 +1,53 @@
 import { doc } from '../config/environment';
-import noop from '../utils/noop';
 
-var css,
-	update,
-	styleElement,
-	head,
-	styleSheet,
-	inDom,
-	prefix = '/* Ractive.js component styles */\n',
-	styles = [],
-	dirty = false;
+const PREFIX = '/* Ractive.js component styles */\n';
 
-if ( !doc ) {
-	// TODO handle encapsulated CSS in server-rendered HTML!
-	css = {
-		add: noop,
-		apply: noop
-	};
-} else {
+// Holds current definitions of styles
+let styleDefinitions = [];
+
+// Flag to tell if we need to update the CSS
+let isDirty = false;
+
+// These only make sense on the browser. See additional setup below.
+let styleElement = null;
+let styleSheet = null;
+let styleProperty = null;
+
+function add( styleDefinition ){
+	styleDefinitions.push( styleDefinition );
+	isDirty = true;
+};
+
+function toCSS(){
+	return PREFIX + styleDefinitions.map( s => `\n/* {${s.id}} */\n${s.styles}` ).join( '\n' );
+}
+
+function apply(){
+
+	// Apply only seems to make sense when we're in the DOM. Server-side renders
+	// can call toCSS to get the updated CSS.
+	if( !doc || !isDirty ) return;
+
+	styleElement[styleProperty] = toCSS();
+
+	isDirty = false;
+}
+
+// If we're on the browser, additional setup needed.
+if(doc && (!styleElement || !styleElement.parentNode)){
+
 	styleElement = doc.createElement( 'style' );
 	styleElement.type = 'text/css';
 
-	head = doc.getElementsByTagName( 'head' )[0];
+	doc.getElementsByTagName( 'head' )[0].appendChild(styleElement);
 
-	inDom = false;
-
-	// Internet Exploder won't let you use styleSheet.innerHTML - we have to
-	// use styleSheet.cssText instead
 	styleSheet = styleElement.styleSheet;
 
-	update = function () {
-		let css = prefix + styles.map( s => `\n/* {${s.id}} */\n${s.styles}` ).join( '\n' );
-
-		if ( styleSheet ) {
-			styleSheet.cssText = css;
-		} else {
-			styleElement.innerHTML = css;
-		}
-
-		if ( !inDom ) {
-			head.appendChild( styleElement );
-			inDom = true;
-		}
-	};
-
-	css = {
-		add ( s ) {
-			styles.push( s );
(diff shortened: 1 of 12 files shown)


fskreuz:
@Rich-Harris I remember components had uuid as identifiers. Why the switch to numbers?

fskreuz:
Tracking https://github.com/ractivejs/ractive/issues/395

evs-chris:
It looks like this would also fix #2292, since the uid goes from monotonic to uuid. Or have I misread that entirely?

Does this also cover the case of adding css using `new Ractive({ ... css: ... })` and `new SomeComponent({ ... css: ... })`?

I haven't really made use of component css up to this point, but this looks like it should do nicely. There are a few minor formatting inconsistencies with spacing around ifs and method defs. Sorry. I'm the whitespace version of a grammar nazi, even though the Ractive style-guide doesn't really agree with my personal aesthetics.

fskreuz:
> It looks like this would also fix #2292, since the uid goes from monotonic to uuid. Or have I misread that entirely?

Yep. Using uuid should prevent two copies Ractive from colliding scoping ids. It's just that I don't know how to test two instances of Ractive on one page/server.

> Does this also cover the case of adding css using new Ractive({ ... css: ... }) and new SomeComponent({ ... css: ... })?

This currently only works with `Ractive.extend`. I asked about that in https://github.com/ractivejs/ractive/issues/2365 if it was intended behavior for it to only work on `Ractive.extend` or if it should work on both cases. If it should work on both, then there's more work to be done. :smile: 

> There are a few minor formatting inconsistencies with spacing around ifs and method defs.

Yep, working on my editor to do that for me. ESLint and editorconfig are quite new to me. :smile:

martypdx:
> This currently only works with Ractive.extend. I asked about that in #2365 if it was intended behavior for it to only work on Ractive.extend or if it should work on both cases

Added a comment on that issue.

> Yep. Using uuid should prevent two copies Ractive from colliding scoping ids. It's just that I don't know how to test two instances of Ractive on one page/server on the unit tests.

Add two divs to fixture and assign them to respective instance:

```
var div1 = document.createElement('div');
var div2 = document.createElement('div');
fixture.appendChild( div1 );
fixture.appendChild( div2 );

var r1 = new Ractive({ el: div1 });
var r2 = new Ractive({ el: div2 });
```

martypdx:
BTW - UUID's where removed from instances because with large numbers of component instances becomes costly. Totally fine on per Component definition.

fskreuz:
> UUID's where removed from instances because with large numbers of component instances becomes costly

Ahh, makes sense. 

> Add two divs to fixture and assign them to respective instance

That's a bit different from https://github.com/ractivejs/ractive/issues/2292 where the case is about multiple copies of Ractive, each recording their own ids, and possibly colliding because they're numbers. :smile:

fskreuz:
Updated to test isolated instances of Ractive by loading them up in iframes and building components with them. Appveyor failing though.

evs-chris:
> Appveyor failing though.

Yeah... every now and then the AppVeyor CI will decide to work. We mostly just go by whatever Travis has to say at this point. :smile:

src/global/css.js
```diff
@@ -1,62 +1,53 @@
...
+let styleSheet = null;
+let styleProperty = null;
+
+function add( styleDefinition ) {
+	styleDefinitions.push( styleDefinition );
+	isDirty = true;
+}
+
+function toCSS() {
+	return PREFIX + styleDefinitions.map( s => `\n/* {${s.id}} */\n${s.styles}` ).join( '\n' );
```
fskreuz: This implementation of `toCSS` (as a prototype method) doesn't seem to make sense when this `toCSS` returns all the styles defined. So...

- Either make `toCSS` a static function that returns all CSS defined in Ractive.
- Gather all the IDs from the instance down to the descendants, and grab only their CSS.

evs-chris:
In the context of a long running server process that is potentially munging lots of Ractive instances, I'd go with the latter. I think that's one of the primary use cases for this, right?

evs-chris:
@fskreuz this has to be the most tested PR I've seen :smile:. Is it ready to go?

The only question I came up with whilst perusing the source was if it would be possible to handle the `toHTML()` case server-side for `enhance`ment purposes, but I couldn't think of a way to do so.

fskreuz:
> this has to be the most tested PR I've seen

Taking no chances :D

> Is it ready to go?

Looks ok to me. Just waiting for eyes to take a look at it and say it's good.

> if it would be possible to handle the toHTML() case server-side for enhancement purposes

If you mean merge this to the `toHTML()` method or something, I'll take a look.

martypdx:
> If you mean merge this to the toHTML() method or don't re-apply when existing, I'll take a look.

What's the thought on this? that `toHTML()` should include a `style` with the css? Or should we just leave that to the caller? ``` `<style>${ractive.toCss()}</style>${ractive.toHtml()}` ```

Everything else LGTM. :+1:

evs-chris:
I would just leave it to the caller. Was just wondering if there was an obvious way to handle it together that I couldn't noodle out.

martypdx:
done like dinner

fskreuz:
:+1:

Issue #2188, opened 2015-10-02
Conditional attribute with `on-` method call events don't work


JonDum:
Is this supposed to work? :)
    
```html  
<button
{{#if thing}}
on-click='set("animal", "dog")'
{{else}}
on-click='foo("cat")'
{{/if}}>Click #1</button> 
```
----
This way doesn't work either
```html
<button
on-click='{{#if thing}}set("animal", "dog") {{else}} foo() {{/if}}'
>Click #2</button> 
```
----
Nor does this. Ractive evaluates it like a normal expression and tries to insert the result instead of calling the method, but I kind of expected that already, just wanted to try it.
```html
<button
on-click='{{thing ? set("animal", "dog") : foo() }}'
>Click #3</button> 
```
----
http://plnkr.co/edit/aaAqQC?p=preview

Madgvox:
At the very least No.2 should be working. I know you're able to add or remove attributes as in No.1, but I guess it doesn't extend to events.

heavyk:
why not just do:

```html
<button on-click="click">Click me</button>
```

```js
onconfig () {
  this.on('click', (event) => {
    if (this.get('thing')) this.set('animal', 'dog')
    else this.foo('cat')
  })
}
```

JonDum:
@heavyk Of course that's an option still, and I've used that many many times. This was more of a question of whether this is working as intended or not.

Also there's so many times where I would rather have it declarative then switch to imperative event handlers. As a real world example, in one of my apps there are multiple supported `languages` that may or may not be available and I need to have a button open a modal if there are multiple languages or call a method on the prototype of ractive instance if there's only a single `language` available. In this specific scenario it's a lot cleaner for me to write:

```html
on-click="{{#if languages.length > 0}} set("modal", "newProgramModal") {{else}} reset() {{/if}}"
```
If it was anymore complicated than that I would switch to a proxy event, but it's annoying making a bunch of `this.on('')`'s all over the place when I already have the methods I need on the prototype.

evs-chris:
See #1270 and #1496.

I have a way to do it, but it's not exactly ideal. I can port #1496 to edge if the masterminds of Ractive think it'll do, but I still haven't cobbled together a fragment based approach to avoid the double parse.

JonDum:
@evs-chris Ah, that answers my question. Would you prefer I leave this open or close and direct conversation to #1496?

martypdx:
[...]

Issue #1270, opened 2014-09-27
Conditional attribute blocks don't work for directives


martypdx:
```html+handlebars
<p {{#if foo}}decorator="drag"{{/if}}></p>
```
This would solve having conditional events for method directives:
```html+handlebars
<div {{#if active}}on-click='set("foo", bar)'{{/if}}></div>
```

evs-chris:
Fixed in edge 😄

Issue #2366, opened 2016-01-31
perf benchmark: ractive blows hot and cold


sabob:
In a nonsensical benchmark described here:
http://www.stefankrause.net/wp/?p=191

they measure 1000 rows created, updated and a row removal.

Interestingly Ractive is the fastest at updating 1000 rows and slowest at removing a row ;-)

Thoughts?

evs-chris:
I haven't looked in-depth, but I do know that edge was way faster than 0.7.3 on several of our internal rendering benchmarks last time I checked.

fskreuz:
> Ractive.js is quite extreme. It’s the fastest when all rows are updated and the slowest in the “remove row” benchmark.

Possibly attributed to the fact that Ractive "surgically" updates the DOM. It tries its best to recycle DOM and only update what's necessary. An earlier wave of benchmarks were done using DB monster. @Rich-Harris did one for Ractive here https://github.com/Rich-Harris/ractive-dbmonster

> ## Why is Ractive so fast?

> Ractive uses similar techniques to Glimmer - it parses the template into a structure such that it's very easy to identify which DOM nodes need to be updated when data changes. This is different to how React (for example) handles DOM updates, which involves re-rendering everything and then running a diff.

Here's some demos:
- Ember https://dbmonster.firebaseapp.com/
- React http://run.plnkr.co/plunks/Wwgjjpl9NHMO5Nd1TUyN/
- Ractive http://www.rich-harris.co.uk/ractive-dbmonster/
- Paperclip (afaik, the fastest) http://paperclip-dbmonster.herokuapp.com/

On the other hand, implementation isn't usually "equal". There's a possibility that some are optimized for the task while some aren't. For instance, Ractive's rows are defined inline to the main component, while React's are defined in an entirely different component. What if Ractive used a component for the rows? What if React used inline rows? It could make the difference.

I'd say the DBMonster benchmarks were closer to "fair" as the framework authors themselves wrote the apps towards the same goal using a common data source. They could have thrown in everything to make their version really fast.

sabob:
Running the test locally in edge shows edge is quite a bit faster than 0.7.3, but only in the things 0.7.3 is already good at, updates ;-)

Initial rendering of 1000 rows and deleting a row is where ractive doesn't shine so brightly.

Although nonsensical, sometimes these micro benchmarks highlights low hanging fruit. Whether it is worth pursuing those gains is another matter of course.

[...]

PR #2369, opened 2016-02-02
Second swing at conditional directives: Revenge of the Fragment - RFC

evs-chris:
__Targeting 0.9 - ~~no tests yet~~__

This is my second attempt at getting directives to play nicely with conditional attributes - the first being #1496. It resolves such issues as #1270 and #2188.

### Status

- [x] Parse tests passing
- [x] Current tests passing
- [x] Tests for new functionality
- [x] `Attribute` can live anywhere in the attributes fragment
- [x] non-fragment-based conditional attributes still work as before
- [x] `EventDirective` can live anywhere in the attributes fragment
- [x] `Decorator` can live anywhere in the attributes fragment
- [x] `BindingFlag` (new - twoway or lazy) can live anywhere in the attributes fragment and recreates the two-way binding when it changes
- [x] `Transition` can live anywhere in the attributes fragment
- [x] Attributes can be used to map data for components
  - [x] Conditional attributes re/un-map data for components
- [x] `as-${decorator}` decorators work properly
- [x] `${transition}-${in|out|in-out}` transitions work properly
- [x] Random commented code cleaned up

### The issue

As of edge pre-0.8, conditional attributes are still implemented with the wonderful hack that uses a div (or svg) to parse them and transfer them to the target element at render. Unfortunately, without adding a second full parse, that precludes any directives (on-*, intro, outro, lazy, twoway bindings, etc) from being included in conditionals. My previous attempt at this did introduce an additional parse step to handle this, but it turned out to be pretty brittle and cumbersome. @martypdx suggested a fragment-based approach, and with the re-org done for models around #1740 and #2022, I think I can comprehend things enough to do so.

### The approach

To achieve unrender-able directives, this starts by modifying the parser to no longer distinguish between plain attributes and directives. They are all now attached to the conditional attributes (`m`) portion of an element as distinct nodes. There are a few new types (BINDING_FLAG, DECORATOR, EVENT, TRANSITION) beside the existing ATTRIBUTE to accommodate the new distinction. When an element is created, it creates items from its `m` template member alongside its normal fragment. Each item then finds its parent element by looking up the fragment hierarchy for the first element. The element then uses any attribute-ish things it needs to set up binding, etc as usual.

Conditional attributes don't actually ever render their fragment in edge now. Since directives need to be part of the normal render cycle, this adds an `inAttributes` flag to `ConditionalAttribute`, which still holds normal non-attribute nodes, that `Interpolator` and `Text` check before actually trying to render to avoid rendering stuff into the DOM that's meant to be contained in the `ConditionalAttribute`. The `ConditionalAttribute` fragments still use the `div` hack to get non-fragment attributes out of the rendered strings.

So far, all this seems to work pretty well, with the only really weird one being `Transition`.

### Addendum

I have also started adding `as-${decorator}` support here because I'm already messing around with all of the requisite bits to do so.

### Issues

~~I still haven't written any tests for new functionality, and there are about 30 parse tests and 35 other tests failing. I still haven't quite nailed down resetting for forms.~~

I wanted to get some feedback on the approach before I get too much farther along to make sure I'm not off in the rough too far. There are probably some use- and corner- cases that I've completely overlooked, too. So, if you have a few minutes, comments, suggestions, criticism, and pushes would be most welcome!


martypdx:
@evs-chris awesome. and challenging :grin: as there are a lot of touch points. 

I noticed you're supporting use on components too.

Are you supporting any type of section? Or just conditional/with?

> I have also started adding as-${decorator} support here because I'm already messing around with all of the requisite bits to do so.

you rock

evs-chris:
> I noticed you're supporting use on components too.

They turn out not to be too difficult in that only `Attribute`s and events matter with attributes being used only for mapping. As an extension, I have also considered it might be useful to expose the current list of mapping keys on the component for people who want to do validation.

> Are you supporting any type of section? Or just conditional/with?

Anything goes as long as it gets parsed in the right context, which basically leaves out true conditional support for partials. Partials are still supported in the old way<sup>__tm__</sup>. I could add a parsing flag that would tell partials to parse as if they were `inTag === true` and then you could use those partials as conditional attributes safely, but I haven't really tried to chase that rabbit yet.

Also, repeated sections, while technically supported, make almost no sense at this point. A future enhancement might be name-bound attributes e.g. `<a {{myAttr}}={{myValue}}>` where object iteration could be handy: `<a {{#each props as v:k}}{{k}}={{v}}{{/each}}>`. I don't know though, as events and friends still probably wouldn't be supported.

evs-chris:
All of the existing tests are now passing, excepting the parser tests which I have temporarily disabled. I'm pretty sure I need to bump the template version to 4 and make it an incompatible change. I'll clean up the parse tests when that happens.

I ended up having to defer some things with runloop tasks to make sure that attrs were in the right place during binding and event attachment and the like. I also added some attr reordering to make sure type, min, and max are first value is last. Can anyone think of a better approach?

evs-chris:
FWIW, this branch runs consistently faster on my machine  (470 - 580ms vs 515 - 700ms) than edge on the initial render test in #2366. The re-render is slightly faster too, but everything else is about the same.

sabob:
@evs-chris can confirm this runs faster for creates and deletes. Is the speedup intentional or just a byproduct of the changes you made?

evs-chris:
@sabob it's entirely coincidental. I suspect it's to do with reducing some of the complexity of element creation by flattening all of the attributes and directives into a single template chunk.

In other news, I've bumped the template version and fixed the parsing tests. I've also added some tests for the new functionality. All that remains now is sanity checking, handling conditional mappings for components (input welcome), and finishing up the attribute parsing for as-${name} directives.

For the conditional mappings, I was planning on handling it basically like a rebind, though I'm not 100% sure that's the best approach. It makes sense for a mapping to change in certain circumstances, but I can't really noodle out a reason for mappings to pop in and out of existence. Especially since using a mapping while it's not mapped would cause the local model to grow a keypath and override the mapping in the future.

sabob:
The current release cycle for ractive seems quite long, Since this PR is almost complete, why not push it into 0.8?

On that note, is there a feeling for a 0.8 release date? haven't seen @Rich-Harris here lately, hope he isn't suffering from [open source fatigue] (https://twitter.com/dan_abramov/status/684141711598092288)

fskreuz:
@sabob 

> I'm afraid we adhere to the John Carmack 'when it's done' philosophy of release cycles...
> @Rich-Harris , https://github.com/ractivejs/ractive/issues/2077#issuecomment-136003028

---

> Instead, when asked for a release date on a new title, Carmack will usually reply that the game will be released "when it's done."
> https://en.wikipedia.org/wiki/John_Carmack#Personal_life_and_philosophy

sabob:
@fskreuz makes sense, I was trying to frame the question such as not to annoy, but in hindsight that question is always annoying especially in open source communities. Apologies for that. Instead I'll try and get hands dirty fixing outstanding bugs!

heavyk:
@sabob thanks for posting that link to sebastian mckenzie's blog. I have recently been kinda burned out myself and reading that was quite nice. cheers

evs-chris:
@sabob my reasoning for targeting 0.9 is that I think we're trying to limit the number of breaking changes with 0.8. There are some fairly major architectural changes underneath 0.8 from 0.7 and we were shooting for making the transition as painless as possible. That said, I suppose we're not to 1.0, which is where I consider stable and semantic to kick in, so for my part, I guess it wouldn't be awful to go ahead and knock out the breaking changes ASAP. There's also the handful of bugs that need to get squished before 0.8 can be a thing. And the docs...

There's also cdn linking to consider for breaking changes because once 0.8 hits, it will be http://cdn.ractivejs.org/latest/ractive.js.

At any rate, this is one of the last big things on the way to what I consider feature completeness for a 1.0 release. @Rich-Harris has a ton of awesome open source pots on his range, and I have no idea how he manages to have a day job at the same time. He'll probably get some more time for Ractive in the next month or two, and in the meantime, anything we can knock out bug and doc wise will make it that much easier to hit 0.8.

evs-chris:
I have added support for named transition directives alongside named decorator directives. Should the `decorator="..."` and `intro="..."` versions be deprecated here as well?

I also added support for conditional component mappings, which is basically handled as a remap. There was a TODO to handle that, so I just added a check to see if a mapping that isn't to the same model already exists. If the mapping exists, then the component is unregistered. Does that sound right?

If those are good and no one sees any obvious tests I've missed, then I think this one is pretty much done.

martypdx:
> I also added support for conditional component mappings, which is basically handled as a remap. There was a TODO to handle that, so I just added a check to see if a mapping that isn't to the same model already exists. If the mapping exists, then the component is unregistered. Does that sound right?

Holy crap, that's pretty impressive. I hadn't even consider that side of things.

BTW - I assume everything in the "attribute fragment" is properly torn down when it is removed?

evs-chris:
Everything with the exception of mappings is torn down. I suppose they should be unmapped, but that means that if there are two-way bindings on top of them, the viewmodel will grow new keypaths and any future mapping will be ignored. I'm not quite sure what to do with that at this point... I suppose just unmap  and roll on?

evs-chris:
I got the feels that moving away from mustache args for directives is good, so I have added deprecation notices for decorator and the intro, outro, and intro-outro directives. Once the deprecation cycle is over, a pretty decent chunk of code could be put out to pasture.

JonDum:
Do you think depreciation is necessary? There's already so many changes in
0.8 at this point, I guarantee any 0.7.3 apps are not going to upgrade
cleanly.

It may be better long run to just make note of it somewhere and add it to
the big list of breaking changes come 0.8 release. Wouldn't that make the
code leaner for 0.8 if you didn't have to worry about these deprecated edge
cases now?

evs-chris:
@JonDum I think if #2331 lands, most 0.7 stuff will come straight through after setting the flag. There were some things deprecated in 0.7 that are throwing on edge for me in components I wrote for 0.5, but beyond that, _I_ haven't seen any serious incompatibilities. There are quite a few bugs that are resolved in edge, though, so I think a minimally incompatible 0.8 would be appreciated by anyone with any significant amount of <= 0.7 code out there.

I don't think there's any reason we couldn't have a fairly short 0.8 -> 0.9 cycle once any bugs hiding out there are discovered. Something between six and twelve weeks seems ideal to me.

In other news, I got unmap (as opposed to remap) working for component attributes. I think that's the last thing left for this aside from a sanity check. Unfortunately, this is a heavier (+500 non-test loc) PR that touches lots of places so a sanity check isn't exactly easy. On the plus side, after deprecation, I think this may be a little better than a wash for code size.

evs-chris:
... and Travis has, once again, left the PR. Does anybody know how to get him to pick this back up? It _does_ pass an `npm run build` locally, but I'd like at least one of the CI flags to be a check.

MartinKolarik:
@evs-chris not sure where's the problem, but I restarted it [here](https://travis-ci.org/ractivejs/ractive/builds/110553597) and it passed.

evs-chris:
@MartinKolarik thanks, but unfortunately it looks like it's still only running against the last commit three days ago (0231ba9). Looking at the [requests](https://travis-ci.org/ractivejs/ractive/requests), it's saying 'branch not included or excluded'. I'm not sure exactly what I did or didn't do to cause that.

MartinKolarik:
@evs-chris nah, that's probably another Travis issue. It's using the latest code from GitHub:

```
Cloning into 'ractivejs/ractive'...
remote: Counting objects: 5339, done.
remote: Compressing objects: 100% (2212/2212), done.
remote: Total 5339 (delta 3670), reused 4311 (delta 3074), pack-reused 0
Receiving objects: 100% (5339/5339), 3.08 MiB | 0 bytes/s, done.
Resolving deltas: 100% (3670/3670), done.
Checking connectivity... done.

$ cd ractivejs/ractive
$ git fetch origin +refs/pull/2369/merge:
remote: Counting objects: 269, done.
remote: Compressing objects: 100% (98/98), done.
remote: Total 269 (delta 221), reused 217 (delta 170), pack-reused 0
Receiving objects: 100% (269/269), 36.37 KiB | 0 bytes/s, done.
Resolving deltas: 100% (221/221), completed with 59 local objects.
From https://github.com/ractivejs/ractive
 * branch            refs/pull/2369/merge -> FETCH_HEAD

$ git checkout -qf FETCH_HEAD
```

MartinKolarik:
I noticed the "branch not included or excluded" stuff too, but don't know what that means.

evs-chris:
I think I got this beat back into shape after merging, and I also managed to get mappings split out to make attribute/mapping logic easier to follow.

evs-chris:
I think I got everything merged successfully with the event expression stuff that just hit master.

evs-chris:
Well, event expressions are a template version breaking change, though I missed that aspect for that PR, and this is too. With that in, I've come around to @JonDum's point of view that we may as well break the things that will eventually be broken now, so any objections to going ahead and merging this now?

heavyk:
👍 from me. I think the changes are really good

evs-chris:
@heavyk thanks for reading through, as it's not a particularly light read :smile:

I'm planning on merging this on Sunday, dealing with the inevitable fallout next week while updating the edge docs, and then trying to get an alpha out for 0.8 week after next.

MartinKolarik:
LGTM

JonDum:
This is so insanely cool, Chris. One of my coworkers was trying to do exactly this the other day. He just *expected* it to work and was surprised it didn't. Awesome job!

zlamma:
Hi. Thanks @evs-chris for the great work.

I just noticed this while trying to use current ractive-load with the latest Ractive because the support for `<link>` is now broken.
I guess [this is not working](https://github.com/ractivejs/rcu/blob/063ced98cc826ed0f35ba95d84a868e278cf28f2/src/parse.js#L110) because of the move from `a` to `m`, and the different structure.

Just thought to ask here before raising the Issue with rcu: doesn't this make you guys consider adding the `a` back for the simple cases of unconditional attributes as a backwards compatibility?

NB: `ractive-load` and any other uses of rcu will also need a release to get the rcu changes.

zlamma:
Aaah, scratch that! Yo've already been in rcu and fixed it. Then all that's needed is re-releasing the ractive-load!

JonDum:
@evs-chris @martypdx Quick question. What does the `as` in `as-${decorator}` stand for? How do you read it in your head? I haven't really grasped it yet.

martypdx:
@jondum "as" is just the English word as. `<span as-tooltip>foo</span>`. Decorate this span _as_ a tooltip. 

It made enough logical sense plus it fot a pattern of short preposition prefix: `on-click`, `as-calendar`.

PR #2372, opened 2016-02-03
Avoid exporting all reachable keypaths from rootmodel when using has - #2366 - RFC

evs-chris:
One of the things that @sabob dug up in #2366 was that lots of resolution in a heavily iterated template was unnecessarily creating full `RootModel` exports just to check for presence of keys. This works around that by implementing the whole `has` method for `RootModel`.

I was going to make the full export opt-in only by adding a flag to `RootModel.get` and possibly `Ractive.get`, but that breaks the assumption that `model.get` can be called for any model anywhere. Thoughts?

### DIFF
--- a/src/model/RootModel.js
+++ b/src/model/RootModel.js
@@ -5,6 +5,9 @@ import Model from './Model';
 import { handleChange, mark } from '../shared/methodCallers';
 import RactiveModel from './specials/RactiveModel';
 import GlobalModel from './specials/GlobalModel';
+import { unescapeKey } from '../shared/keypaths';
+
+const hasProp = Object.prototype.hasOwnProperty;
 
 export default class RootModel extends Model {
 	constructor ( options ) {
@@ -63,7 +66,7 @@ export default class RootModel extends Model {
 
 		this.extendChildren( ( key, model ) => {
 			result[ key ] = model.value;
-		})
+		});
 
 		return result;
 	}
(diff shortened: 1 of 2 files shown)


evs-chris:
Also, I tried this out on an a heavily iterative template with lots of expressions on one of my projects and it cut render time from 13s to 1s for ~1500 big funky chunks of template.

martypdx:
@evs-chris I've stared at your changes for 5 minutes but I'm not seeing what is getting short-cutted. How is moving the code into the RootModel different than calling `super.has`?

Also, AFAIK `prop in obj` is same as doing the crawl up the prototype chain, so can we just use that?

martypdx:
Oh, Model calls `this.get()` and Root is doing `this.value`

martypdx:
Yeah, I think we should implement a module function that does:

```js
function hasProp( obj, key ) {
    return obj && key in obj;
}
```

I've read that use of `in` can mess with optimization, so if we put in own function it isolates that lookup from its use.

evs-chris:
> Also, AFAIK prop in obj is same as doing the crawl up the prototype chain

That actually came up twice in the original PR (#2192) :grin:

evs-chris:
> I've read that use of `in` can mess with optimization, so if we put it own function it isolates that lookup from its use.

I haven't seen the Chrome profiler flag any of our uses of `in` so far. I haven't really researched any of the other browsers optimization fences, so I have no idea about them. That said, a quick grep shows 68 instances of `.*(if|for).*\w+ in \w+` in `src`, so if it would help performance, I'd guess 1/5 of those are in warm to hot code paths. So, should I go ahead and add a `hasProp` to this as well?

PR #2375, opened 2016-02-07
Gh 2366

sabob:
Exit early from Model.adapt if there are no registered adaptors. This provides small performance boost in tight loops when setting modifyArrays to false.

### DIFF
--- a/src/model/Model.js
+++ b/src/model/Model.js
@@ -57,6 +57,7 @@ export default class Model {
 		const adaptors = this.root.adaptors;
 		const value = this.value;
 		const len = adaptors.length;
+		if ( len === 0 ) return;
 
 		// TODO remove this legacy nonsense
 		const ractive = this.root.ractive;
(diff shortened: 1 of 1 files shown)


src/model/Model.js
```diff
@@ -57,6 +57,7 @@ export default class Model {
 		const adaptors = this.root.adaptors;
 		const value = this.value;
 		const len = adaptors.length;
+		if ( len === 0 ) return;
```
fskreuz: Being a little nitpicky, might want to pull this up just after `adaptors`. The `value` assignment becomes useless if it bails out early.

Might also want to look at those 2 hacks down below. It seems your performance boost relies on avoiding `getKeypath`. If those comments tell the truth, there could be another way of doing this function, avoiding all the hacks in this sector of the code. :smile: 

Other than that, :+1:

sabob:
@fskreuz I've no idea what the legacy code is or how rewrite it.

evs-chris:
@sabob the hacky code referred to a few lines down is that the ArrayWrapper holds its own special blessed reference to the Ractive model that holds the value it is wrapping. I think that change is out of the scope of this PR. One of the longtime TODOs is to move the provided wrappers (array and magic) to their own projects and refactor Ractive's APIs such that they can still continue to work without special treatment.

PR #2376, opened 2016-02-08
Allow keypath-taking methods to use an element for context for resolution - RFC

evs-chris:
This is an attempt to make life outside of template context (method events, decorators, and other strange and dark corners of javascript) by allowing the keypath-taking methods of Ractive instances to take an element reference as a context parameter used for resolving the keypath. Here's an example:

## Example

```js
const r = new Ractive({
  template: `<ul>{{#each people as person}}
    {{#person.details.address}}
      <li {{#if .picked}}class="picked"{{/if}} on-click="toggle('.picked', event.original.target)">
        {{address}}, {{city}}, {{state}} {{zip}}
      </li>
    {{/}}
  {{/each}}</ul>`,
  ...
});
r.set({ '.city': 'Hell', '.state': 'MI' }, r.find('ul li'));
console.log(r.get('person', r.findAll('ul li')[2]));
``` 

In this example:

1. Clicking on one of the `li`s `toggle`s the `picked` flag on the context of the first `ul li`, which happens to be the first item in the `people` list. The benefits to this become lots more compelling if you have contexts with aliases that pop around different objects, such as looking up an object in a root context lookup e.g. `~/states[.state].directory` inside a user `li`.

2. The city and state for the first address are changed to 'Hell, MI' without the intervening keypath gubbins, which in this case are admittedly minor but can become much more complex in the wild.

3. The third item in the `people` array is logged to console by looking it up from the `li` to which it is attached.

## Changes

This adds an extra optional parameter for context in the following methods:
```
set( keypath, value [, context] )
set( object [, context] )
get( keypath [, context] )
add( keypath [, number] [, context ] )
subtract( keypath [, number] [, context] )
toggle( keypath [, context] )
splice( keypath, [context,] index, drop, ...add )
push( keypath, [context,] ...values )
pop( keypath [, context] )
shift( keypath [, context] )
unshift( keypath, [context,] ...values )
sort( keypath [, context] )
reverse( keypath [, context] )
update( keypath [, context] )
updateModel( keypath [, cascade] [, context] )
animate( keypath, [context,] value [, options] )
link( keypathSrc, keypathDest [, context] )
```

If a keypath has a wildcard, context is not supported.

## RFC

First off, am I the only one who has ever wished for this? Second, I tried to make the API as consistent as possible with the current methods. With the exception of `set`<sub>1</sub>, `add`, and `subtract`, you can just add a context after the keypath. Should those be adjusted to match the others? I did them first, and it felt more natural to have the context trailing there. I suppose the context could be shifted to trail on the varargs methods, and that would make it a little more consistent too.

Another alternative would be to supply an `options` object with a `context` flag for each of the relevant methods.

## Addendum

I also considered another special ref (heh), `@element` that would grab the node of the nearest parent element for use in events. Another option would be to add the appropriate value to the `event` object that is available during method event resolution e.g. `event.el` or `event.element` or `event.target`.

## TL;DR

Use keypaths outside of templates as if they were in template by passing an element from which to start resolution.

Thoughts?

### DIFF
--- a/src/Ractive/prototype/add.js
+++ b/src/Ractive/prototype/add.js
@@ -1,5 +1,5 @@
 import add from './shared/add';
 
-export default function Ractive$add ( keypath, d ) {
-	return add( this, keypath, ( d === undefined ? 1 : +d ) );
+export default function Ractive$add ( ...args ) {
+	return add( this, args, 1 );
 }
--- a/src/Ractive/prototype/animate.js
+++ b/src/Ractive/prototype/animate.js
@@ -1,9 +1,10 @@
 import runloop from '../../global/runloop';
 import interpolate from '../../shared/interpolate';
-import { isEqual } from '../../utils/is';
+import { isEqual, isRactiveElement } from '../../utils/is';
 import { splitKeypath } from '../../shared/keypaths';
 import easing from '../../Ractive/static/easing';
 import noop from '../../utils/noop';
+import resolveReference from '../../view/resolvers/resolveReference';
 
 const noAnimation = { stop: noop };
 const linear = easing.linear;
@@ -26,7 +27,15 @@ function getOptions ( options, instance ) {
 	};
 }
 
-export default function Ractive$animate ( keypath, to, options ) {
+export default function Ractive$animate ( keypath, to, options, overflow ) {
+	let model;
+
+	if ( isRactiveElement( to ) ) {
+		model = resolveReference( to._ractive.fragment, keypath );
+		to = options;
+		options = overflow;
+	}
+
 	if ( typeof keypath === 'object' ) {
 		const keys = Object.keys( keypath );
 
@@ -40,7 +49,7 @@ ${keys.map( key => `ractive.animate('${key}', ${keypath[ key ]}, {...});` ).join
 
 	options = getOptions( options, this );
 
-	const model = this.viewmodel.joinAll( splitKeypath( keypath ) );
+	if ( !model ) model = this.viewmodel.joinAll( splitKeypath( keypath ) );
 	const from = model.get();
 
 	// don't bother animating values that stay the same
--- a/src/Ractive/prototype/get.js
+++ b/src/Ractive/prototype/get.js
@@ -1,26 +1,32 @@
+import { isRactiveElement } from '../../utils/is';
 import { splitKeypath } from '../../shared/keypaths';
 import resolveReference from '../../view/resolvers/resolveReference';
 
-export default function Ractive$get ( keypath ) {
+export default function Ractive$get ( keypath, context ) {
 	if ( !keypath ) return this.viewmodel.get( true );
-
-	const keys = splitKeypath( keypath );
-	const key = keys[0];
-
 	let model;
 
-	if ( !this.viewmodel.has( key ) ) {
-		// if this is an inline component, we may need to create
-		// an implicit mapping
-		if ( this.component && !this.isolated ) {
-			model = resolveReference( this.component.parentFragment, key );
+	if ( isRactiveElement( context ) ) {
+		model = resolveReference( context._ractive.fragment, keypath );
+		if ( model ) return model.get( true );
+	} else {
+		const keys = splitKeypath( keypath );
+		const key = keys[0];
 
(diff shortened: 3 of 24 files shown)


evs-chris:
I forgot to mention that I also added a `result` and `ractive` property to the `Promise` returned by array methods, because sometimes the synchronous result is useful and chaining.

evs-chris:
I liked `event.el` so much that I went ahead and added it. It also has an alias `element` to match a (near) future `element` alias for Ractive's `el` config option... or is `target` better? I'm think I prefer `element`.

evs-chris:
Not sure why Travis is not keen on running this PR, but it passes a local `npm run build`...

I completely missed a few methods where context is handy: animate, link, update, and updateModel. Link, in particular, benefits from this because one of the main use-cases for it is master-detail views where selecting a master links in the details. Updating the initial post to include the missed methods...

guilhermeaiolfi:
> First off, am I the only one who has ever wished for this? 

I think so :) 

But couldn't the result (context) of ```r.find()``` be a mini-ractive instance (or a wrapper) so you could use: 

```
r.find("ul li").set({ '.city': 'Hell', '.state': 'MI' }); 
```

Other than that, the ```options```seems better for accommodate future ideas and keep compatibility.

martypdx:
> First off, am I the only one who has ever wished for this?
> > I think so :)

I'm not really following the need myself. It seems antithetical to reactive programming to be mucking with elements.

I'm thinking what you're struggling with is getting the full keypath from the context so you know what to pass to `keypath` parameter of `set`, `toggle`, etc.. 

What if the `@keypath` had methods:

```html
<li {{#if .picked}}class="picked"{{/if}} on-click="toggle( @keypath.join('picked') )">
        {{address}}, {{city}}, {{state}} {{zip}}
</li>
```

Does that make more sense than going the round-about way of passing an element to get the context to get the keypath?

evs-chris:
> I think so :)
> I'm not really following the need myself.

I guess it is just me :smile:

> But couldn't the result (context) of r.find() be a mini-ractive instance

Well, right now, it's a DOM Node, and I don't know of a way merge those without things getting weird for one side or the other.

> Other than that, the options seems better for accommodate future ideas and keep compatibility.

I started out adding an options param and using it to pass a `context` key, but it didn't feel quite write in the test templates. Also, this is fully compatible with the current methods.

> What if the @keypath had methods

If the methods account for parent refs and aliases, then that would certainly solve the reference side. `@keypath` would have to be blessed in the parsing process or change into an object with the current behavior moving to `@keypath.current` or `@keypath.string` or something.

The reason I went with an element for context was lowest common denominator for both in and out of the template. I don't really see much use for using `find` to get the element and work from there, but that was the easiest way to demonstrate. If you're in a decorator or doing some sort of other non-Ractive thing with your Ractive app, you can usually get a reference to an element pretty easily. The element is also available to events pretty easily. 

> I'm thinking what you're struggling with is getting the full keypath from the context so you know what to pass to keypath parameter of set, toggle, etc..

That's it exactly. I usually end up doing some funky string manipulation on the `@keypath` to get what I'm after. If the children of computeds thing gets resolved in a writable way, my string mangling will tend to fall apart without becoming significantly more complex.

If the general consensus is that this isn't useful/wise as it stands and that souping up `@keypath` would be, I'll pursue that and see where it ends up. If not, I think I've explored far enough that I can come up with a reasonable external library to do keypath resolution against a Ractive-owned DOM node - it would just be using private APIs.

evs-chris:
Another use-case for element-based keypath resolution would be delegated event handling. If, for instance, you had a 5000 row table and wanted a handful of click events on each, delegation would actually be beneficial for performance. With reference lookup from the target element, you could accomplish that with much less headache.

I think I'm going to sunset this one, as it seems it is less cowabunga than I thought. I was given a few ideas for a hopefully more palatable way to accomplish the same thing, though. I'm leaning towards a `Ractive.resolve(path, el)` for API use and `@keypath(relativeRef)` for template use, where `@keypath(...)` would be similar to `ReferenceExpressionProxy`.

guilhermeaiolfi:
I think ember uses event delegation for lists by default. Would that be possible for ractive?

Issue #2005, opened 2015-06-21
Weird array bug in magic mode


atomiomi:
Hi! While implementing nested list I found out some weird bug. Here is [test](http://jsfiddle.net/pvtabau9/1/). In this case I want to remove some item from array and put it to another item (`children` field). If one item is added to another everything is ok but after adding another one, children list of parent item is cloned to next item. Just click two times on `Move` button in test and you will see that `third` and `fourth` have same children array. Without magic mode everything works fine.

MartinKolarik:
The original bug seems to be fixed, but the fiddle demonstrates two new bugs in edge. With `magic: true`, items are not removed from `parents` array, and there's an error in the console saying `Uncaught Error: TODO magic adaptor reset`, with `magic: false`, second `move()` fails with `Uncaught TypeError: Cannot set property 'index' of undefined` inside `RepeatedFragment#updatePostShuffle()`

evs-chris:
I think I may be missing a step on adaptors and shuffling. In edge right now, I don't see where adaptors get notified if their keypath changes from a shuffle. That's mostly fine for the edge array wrapper that gets a reference to a model and just uses that, but it's not good anything else that has to interact with the instance using its initial keypath. It doesn't look like there's really a good way around this other than to add some sort of `shuffled(newKeypath)` support to the adapator API so that it can moved without having to be torn down and re-assembed.

For reference, I bumped into this whilst trying to rewrite the array adaptor such that it doesn't rely on private internals and could thus be extracted at some point in the future.

sabob:
@evs-chris I came to the same conclusion here: https://github.com/ractivejs/ractive/pull/2380#issuecomment-191365419

I briefly checked how 0.7.3 does it and there seems to be a step for updating keypaths after a shuffle which is missing in edge.

evs-chris:
@sabob oops, I missed (or more likely misread) that comment amidst other discussion.

After having slept on it, and looking more into what is expected of adaptors from existing docs, I came to the conclusion that they should probably be reset with the value _and_ keypath if they support reset. Otherwise, they should be torn down and set back up as usual. For the top-level adaptor, it doesn't really matter, but if an adaptor happens to be downstream from another adaptor, it is left as-is.

There is another deeper question here too: should adaptors be reset on every change that updates their value? I feel like they should, but right now, `mark`ing a model may change its value without `reset`ing the adaptor.

[...]

PR #2378, opened 2016-02-11
Add a splice-oriented handler for updatePostShuffle - RFC/Review

evs-chris:
For background, see #2366

There are two scenarios to arrive in `updatePostShuffle`: `merge` and `splice`(ish things). The current algorithm is a sort-of hybrid for handling merges and splices that ends up detaching and re-attaching everything past a certain point. The detaching and re-attaching is required for merges, which tend to be more of a jumble, and the 'past a certain point' is aimed at splices, which operate on a single contiguous region.

For the delete test on which Ractive performs poorly, part of the performance issue seems to be caused by detaching and re-attaching _every_ element in the repeated fragment. This adds a check to see if everything remains sequential, and if it does, avoids detaching and re-attaching iterations that are already in place. New iterations are aggregated and stuffed in before the next existing iteration or the end of the repeated fragment, whichever comes first.

I was going to try to use the same loop to handle both merges and splices, but I couldn't come up with a fast way to make sure the existing fragments were in the proper order for the merge. I figured the existing method would probably end up being faster than trying to sort in the DOM. I also thing that splices are probably more common than merges.

On latest, the test takes ~700ms for the first run on my machine and settles down to ~300ms. On edge, the test takes ~160ms every time, but deleting the _second_ element takes ~280ms. From there, the closer you get to the end of the list, the faster it runs. With this change, the test still takes roughly ~160ms (it might be a few ms better, but is close enough I'll call it the same), but deleting further elements also clocks in at ~160ms.

For further reference, the actual shuffling that Ractive does takes ~70ms. I'm assuming that the rest of the time is taken by the browser, which would mean that some of the other libraries in the test do the delete in ~10ms or less.

### DIFF
--- a/src/view/RepeatedFragment.js
+++ b/src/view/RepeatedFragment.js
@@ -357,11 +357,18 @@ export default class RepeatedFragment {
 		// This algorithm (for detaching incorrectly-ordered fragments from the DOM and
 		// storing them in a document fragment for later reinsertion) seems a bit hokey,
 		// but it seems to work for now
-		let previousNewIndex = -1;
-		let reinsertFrom = null;
+		const len = this.context.get().length;
+		let i, maxIdx = 0, merged = false;
 
 		newIndices.forEach( ( newIndex, oldIndex ) => {
 			const fragment = this.previousIterations[ oldIndex ];
+
+			// check for merged shuffles
+			if ( !merged && newIndex !== -1 ) {
+				if ( newIndex < maxIdx ) merged = true;
+				if ( newIndex > maxIdx ) maxIdx = newIndex;
+			}
+
 
 			if ( newIndex === -1 ) {
 				fragment.unbind().unrender( true );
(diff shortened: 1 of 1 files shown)


sabob:
Great to separate the two concepts. Investigating #2005 I stumbled upon the updatePostShuffle function and didn't even realize it did both merge and splice.

With this change on my machine, deleting first row takes 250ms, deleting last row 95ms. Before this change both operations took ~250ms

My ignorance is probably showing but it feels as if there is a way to drastically optimize deletes. It is faster to update the entire table of 1000 rows (120ms) than to delete the first row (250ms).

When deleting the first row, is the shuffle done in the virtual dom, and then ractive figures out only the first row in the browser DOM should be deleted, or does ractive copy every row content to the one above it and end up with an empty row at the bottom, which it then deletes from the browser DOM?

evs-chris:
I'm looking for another way to shave some time off of shuffling. Unfortunately, profiling isn't much help (in Chrome) because most of the time is spent outside of Ractive's code. This was just the first thing that I tried that didn't fail a bunch of tests and actually reduced time spent shuffling :smile:

I'm looking at changing how `rebind` happens too, as shuffling the children of an array model instead of rebinding got the time on my machine down to ~95ms, but it also breaks a bunch of tests. The tests that fail mostly seem to be index and keypath reference tests, so rejiggering `IndexModel` and `KeypathModel` may help that work.

There's something a little weird about this scenario anyhow, because on the blogger's machine, which seems to be considerably more powerful than mine, the lowest time for this test is 80ms. Most of the times are closer to 120ms.

evs-chris:
Chrome's timeline is invaluable when trying to figure out what's happening. It basically combines js profiling and browser render profiling in one place.

I found a way to knock Ractive's part of this down to ~45ms, but it's a fairly major rework of rebinding. It might be possible to pull it down a little more with some optimization. Looking into it a bit more, most of the time is taken by Chrome recalculating styles, which probably has much to do with Bootstrap being used for styles - striped tables are apparently expensive. Even using `document.querySelector('tr').remove()` takes ~100ms on my machine, where using Ractive's takes ~140ms (between 127 and 300, clustering around the 140s).

Based on those numbers, I'd say the lowest time is just above Chrome's style recalculation time. The rebinding rework would put Ractive between 110ms and 140ms on the bloggers machine.

This also answers the question of why removing an element is so much slower than replacing the whole list - when replacing the whole list, the elements get re-used and don't appear to trigger style reapplication, whereas removing an element does.

heavyk:
chris have you seen this?

https://github.com/wilsonpage/fastdom

I've been using it very effectively to reduce a lot of the things you're noticing. I bet if you grouped all of the measures and mutates like fastdom helps you do, I think you can get those numbers down even more. it really works wonders.

evs-chris:
I haven't seen that. Interesting. I'm not sure how it would work with Ractive's synchronous nature, though.

Just for giggles, I removed the bootstrap stylesheet and added a little inline `<head>` styling to make the remove buttons show up. With that plus the rebind changes I mentioned earlier, removes in the benchmark drop to between 50 and 100ms and cluster around 70ms. It goes from me being able to see the delay between the click and the removal to it looking like the removal happens at the end of the *click* made by my touchpad.

evs-chris:
I have pushed a new branch based on this one with the rebinding experiment in it here: https://github.com/ractivejs/ractive/compare/splice-shuffle...splice-shuffle-2

heavyk:
> I'm not sure how it would work with Ractive's synchronous nature, though.

it doesn't. that's why I haven't made a pr to ractive to use fastdom. however, what can be used is the technique of grouping modifications into one tick. so, for ractive I postulate that if you queue up the dom modifications and then, before reading (measureing) the dom, that queue would need to be emptied first. you know the code way better than me but I imagine grouping these up will save tons of time.

sabob:
@evs-chris numbers are looking good. Thanks for the hint to chrome's timeline, that is a useful tool.

@heavyk doesn't ractive's runloop batch dom operations, similar to fastdom?

martypdx:
@evs-chris fyi In #1740, I implemented slice handling [here in the section](https://github.com/ractivejs/ractive/blob/keypaths-ftw/src/virtualdom/items/Section/blocks/EachBlock.js) and [here in the model](https://github.com/ractivejs/ractive/blob/keypaths-ftw/src/viewmodel/context/BindingContext/shuffle.js)

heavyk:
@sabob not really. the reason is obvious. ractive is synchronous so you could easily be doing some of either in your components. I just started experimenting with it and you can definitely do fastdom in your components (and it helps a lot) but for example, if your components are modifying in the template and measuring during events they will definitely be mixed.

PR #2379, opened 2016-02-11
Fix server-side "window is not defined" error

monoblaine:
Fixes "window is not defined" error when Ractive.js is used in a server-side library (i.e. [ClearScript](https://clearscript.codeplex.com/))

### DIFF
--- a/src/model/specials/GlobalModel.js
+++ b/src/model/specials/GlobalModel.js
@@ -4,7 +4,7 @@ import Model from '../Model';
 class GlobalModel extends Model {
 	constructor ( ) {
 		super( null, '@global' );
-		this.value = typeof global !== 'undefined' ? global : window;
+		this.value = typeof global !== 'undefined' ? global : win;
 		this.isRoot = true;
 		this.root = this;
 		this.adaptors = [];


monoblaine:
Well, actually I've just realized that there's some other problem here: The variable `global` is always `undefined` because it's not defined globally. Am I wrong?

evs-chris:
`global` is the nodejs equivalent of `window`. If you need to support an environment where the global object is not named `global` or `window`, you can update the conditional to `typeof global !== 'undefined' ? global : typeof ${otherGlobal} !== 'undefined' ? ${otherGlobal} : window` or change it to a less confusing set of `if else`s.

Is the global object for ClearScript named `win`?

monoblaine:
No, the global object in ClearScript is not `win`. I was just referring to the declaration [here](https://github.com/ractivejs/ractive/blob/3de5b6e347b3e267498824e63e948558f93fe1e0/src/config/environment.js#L3). Apperantly there's no such thing as a global object in ClearScript. In fact, [someone suggested](https://clearscript.codeplex.com/workitem/31) executing some code like the following:

```js
window = this;
```

In my case, I would happily execute `global = this;` hoping it would be OK.

Issue #1342, opened 2014-10-13
bug: magic mode and list manipulations cause trouble


dagnelies:
Hi,

Here is the fiddle:
http://jsfiddle.net/bjmmnq7s/

Without 'magic mode', it works more or less fine(*), but when 'magic:true' is enabled, it breaks.

The example is my attempt to have generic functions on list to move entries up&down, delete and add them. All you need for a basic table operations handling actually. However, in combination with magic mode, it runs wild and throws "TypeError: fragment is undefined" errors.

(*) Without magic mode, it runs fine except when moving down the last-1 item and adding a row afterwards, then a "down" button will be missing.

Cheers,
Arnaud

PR #2380, opened 2016-02-11
Gh 2005

### LINKED ISSUE #2005: Weird array bug in magic mode

sabob:
This PR is an attempt to fix #2005.

Currently magic.js#reset throws an error and a comment that this method is unlikely to be called. Seems it is called after all. Here I removed the error thrown and return false. Magic.js#reset is invoked from Model.js#applyChanges. Returning false from magic#reset will forces a teardown and adaption of the new value. Not sure what the correct solution here is, but the tests pass...

The next two errors faced was in RepeatedFragment#shuffle and RepeatedFragment#updatePostShuffle where fragments are looked up but could be undefined. So I added null checks to guard against executing the functions further. Feels like treating the symptoms rather than the cause though?

I've also uncommented the test #1342 which seems to work now.

### DIFF
--- a/src/view/RepeatedFragment.js
+++ b/src/view/RepeatedFragment.js
@@ -210,6 +210,10 @@ export default class RepeatedFragment {
 			if ( newIndex === -1 ) return;
 
 			const fragment = this.iterations[ oldIndex ];
+
+			// guard against null fragment - #2005 (feels like treating the symptom, not the cause?)
+			if ( fragment == null ) return;
+
 			iterations[ newIndex ] = fragment;
 
 			if ( newIndex !== oldIndex && fragment ) fragment.dirty = true;
@@ -362,6 +366,9 @@ export default class RepeatedFragment {
 
 		newIndices.forEach( ( newIndex, oldIndex ) => {
 			const fragment = this.previousIterations[ oldIndex ];
+
+			// guard against null fragment - #2005 (feels like treating the symptom, not the cause?)
+			if ( fragment == null ) return;
 
 			if ( newIndex === -1 ) {
 				fragment.unbind().unrender( true );
--- a/src/Ractive/static/adaptors/magic.js
+++ b/src/Ractive/static/adaptors/magic.js
@@ -98,7 +98,7 @@ class MagicWrapper {
 	}
 
 	reset () {
-		throw new Error( 'TODO magic adaptor reset' ); // does this ever happen?
+		return false;
 	}
 
 	set ( key, value ) {
--- a/test/browser-tests/plugins/adaptors/magic.js
+++ b/test/browser-tests/plugins/adaptors/magic.js
@@ -49,6 +49,64 @@ try {
 		t.htmlEqual( fixture.innerHTML, 'Fozzie: bear' );
 	});
 
+	test( 'Test that splice array works in magic mode', t => {
+		const items = [
+			{ name: 'a' },
+			{ name: 'b' },
+			{ name: 'c' }
+		];
+
+		new Ractive({
+			template: '{{#each items}} {{name}} {{/each}}',
+			el: fixture,
+			data: { items: items },
+			magic: true
+		});
+
+		t.htmlEqual( fixture.innerHTML, 'abc' );
+
+		arr.splice( 0, 1 );
+
+		t.htmlEqual( fixture.innerHTML, 'bc' );
+	});
+
+	test( 'test moving one array to another #2005', t => {
+
+		var parents = [
+		    { name: 'first', children: [] },
+			{ name: 'second', children: [] },
+			{ name: 'third', children: [] },
+			{ name: 'fourth', children: [] }
+		],
+		third = parents[2];
+
+		new Ractive({
+			template: '{{#parents}}<li>{{this.name}}</li><ul>{{#this.children}}<li>{{this.name}}</li>{{/}}</ul>{{/}}',
+			el: fixture,
+			data: { parents: parents },
+			magic: true
+		});
+
+		t.htmlEqual( fixture.innerHTML, '<li>first</li><ul></ul><li>second</li><ul></ul><li>third</li><ul></ul><li>fourth</li><ul></ul>' );
+
+		third.children.push( parents[0] );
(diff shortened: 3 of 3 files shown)


src/view/RepeatedFragment.js
```diff
@@ -210,6 +210,10 @@ export default class RepeatedFragment {
 			if ( newIndex === -1 ) return;
 
 			const fragment = this.iterations[ oldIndex ];
+
+			// guard against null fragment - #2005 (feels like treating the symptom, not the cause?)
+			if ( fragment == null ) return;
```
evs-chris: Yeah, that's a little odd. Is something in the shuffle creating a null iteration? I haven't had time to look yet, but I'd guess it has something to do with the model and wrapper getting out of sync during the shuffle from the model side.

sabob: @evs-chris Indeed, the wrapper and model gets out of sync. I spent some debugging and found the following sequence:

We start by doing childrenArray.push(...)
1. we call a function, ```processWrapper```, which retrieves it's associated ractive model through the hack: ```var __model = wrapper.__model;```
2. we then shuffle the model, shuffle the fragments and eventualy call ```RepeatedFragment.updatePostShuffle``` to rebind the model to the iterated fragments
3. we render the new fragment

Next we call parentArray.splice(...)
1.  again we call processWrapper, which retrieves it's associated ractive model through the hack: ```var __model = wrapper.__model;```
2. we then shuffle the model and shuffle the fragments. Because this is a splice we move fragments to different models.
3. we don't seem to update the ```wrapper.__model``` to it's new model, so ```wrapper.__model``` points to it's old model and old fragment

Next we call childrenArray.push(..) again:
1. same story, ```processWrapper``` retrieves it's associated ractive model through the hack: ```var __model = wrapper.__model;```. This model is still the old model pointing to the old fragment
2. we then do shuffling and eventually ```RepeatedFragment.updatePostShuffle``` to rebind the model to the iterated fragments. However because we have the wrong model, we also have the wrong RepeatedFragment instance (this fragment has no iterations) and retrieves the null fragment.

Hope that makes sense.

So the fix requires us to update the ```wrapper.__model``` once shuffling occurs. But I'm not sure where in the code that should occur?

The wrapper.__model is hacky anyway, so probably needs a bit of a overhaul to be more robust?

evs-chris: Nice sleuthing! Definite yes to the hacky part. There's a sort of long standing TODO on tearing ArrayWrapper and MagicWrapper out into their own plugins. Magic is a different can of worms, but I think the first step to moving ArrayWrapper would be to have it delegate to the `Ractive` array methods instead of managing its own runloop turn and model. That should eliminate the `__model` hack and clear up this issue as well.

I think I have a stash somewhere that does the delegation. I can open a PR with it if you'd like, or if you'd rather dig in and handle it yourself, I'd be happy to help in whatever ways I can.

sabob: I think I'm too much of a noob to tackle the delegation, which sounds like the right approach. For example, how would one build the keypath to invoke the native methods? From the template? Sorry, you have alot on your plate already, but it's probably better if you do a PR. I'll help where I can.

fskreuz: Code review time! Use `===`. :smile:  I believe I had a discussion somewhere here before where one should always use strict comparison, and loose only in special cases.

src/view/RepeatedFragment.js
```diff
@@ -363,6 +367,9 @@ export default class RepeatedFragment {
 		newIndices.forEach( ( newIndex, oldIndex ) => {
 			const fragment = this.previousIterations[ oldIndex ];
 
+			// guard against null fragment - #2005 (feels like treating the symptom, not the cause?)
+			if ( fragment == null ) return;
```
fskreuz: And here too :smile:

src/view/RepeatedFragment.js
sabob: I normally use == to compare against null as my intent is to include undefined as well. Not sure how folks compare for null+undefined?

heavyk: you can put `if (!frament)` instead. however, that will also catch `0` and `false`

there is no way to check both `null` and `undefined` with strict equality.

MartinKolarik: I'd say go with `==` if you want to check for both.

webraptor: why not
`if ( fragment === undefined || fragment === null ) return;`

martypdx: I'm on board with using `foo == null` represent a check against a primitive or multi-type value for not existing (`null` or `undefined`). When I see that  `== null` then I know that's exactly the condition of that variable - it could exist but have a possible falsey value like `0`.

However, in this case `fragment` is either an object or not, so `if ( !fragment )` should be sufficient and it communicates that expectation.

martypdx:
General note: With the transition to `mapped` data between components, strictly speaking the Array and Magic Adaptors no longer need to maintain multiple ractive instances. That would likely greatly simplify the implementations.

sabob:
@evs-chris delegation doesn't seems too hard after all. The keypath is already available to the wrapper here:
https://github.com/ractivejs/ractive/blob/9085396013226e409bd2a9f6b3bbfd01991a0692/src/Ractive/static/adaptors/array/index.js#L14

We can use that keypath to delegate, however I'm unsure how or rather where to update that keypath once the model shuffles.

For example, in #2005 the keypath could be "parents[2].children". Say we perform a splice, the keypath might have to be updated to "parents[1].children". How can we update the keypath reliably?

evs-chris:
I'm pretty sure this is handled in edge now. Big :+1: to @sabob.

PR #2382, opened 2016-02-15
ractive.animate should always return a promise, even if no animation

Rich-Harris:
`ractive.animate(...)` was failing to return a promise in cases where it wasn't possible to interpolate values, causing unexpected behaviour. This PR fixes it


evs-chris:
LGTM. OK to merge?

Rich-Harris:
:shipit:

Issue #2311, opened 2015-12-07
Children of component arguments ($1.foo)


Rich-Harris:
Just realised you can't access properties of arguments passed through to components: http://jsfiddle.net/rich_harris/6n58rwg2/

Filed under nice-to-have

evs-chris:
Available in edge :smile:

Issue #1396, opened 2014-10-22
Multiple handlers on inline events? Is it possible?


arxpoetica:
Consider the following:

`<button on-click="save(.)" on-custom-click-event="track(.)">Save</button>`

It would be nice if I could not have to create a separate click event just to handle two separate function calls. Is this possible? Obviously the following doesn't work, but is their an equivalent?

`<button on-click="save(.);track(.)">Save</button>`

Is it possible to run multiple calls on events? Does this make sense?

evs-chris:
This is now possible on edge with event expressions 😄

PR #2386, opened 2016-02-16
Event expressions - RFC

### LINKED ISSUE #1172: Discussion: should we deprecate the old proxy event syntax?

evs-chris:
This is my attempt to resolve #1172, #1396, #2311, and partially #2176. It is also somewhat tied to #2369.

### What this does

This takes event handling and unifies it as an expression instead of having special parsing for handling method events. This opens event handling up to everything possible with an expression in Ractive. Here're a few examples:

* call data methods e.g. `on-click=".loadStuff()"`
* call more complex structures e.g. `on-click="@this.reset.partial('foo')"` or `on-click=".foo().and.bar(., 'yep')"`
* call parent methods for tightly integrated (container) components e.g. `on-click="@this.parent.doAThing(.)"`
* chain multiple actions in an event handler e.g. `on-click="@this.toggle('foo') && @this.fire('fooed')"` or since these are parsed as a list `on-click="@this.toggle('foo'), @this.fire('fooed')"`
* the code is a lot smaller and generally easier to follow as a result

I have also taken the spread args and made it available anywhere, which is what ES2015 allows e.g. `on-foo="fn('hi', ...arguments, 42)"` is perfectly legal. It is also extended to any other call expression with any reference e.g. `{{ doThing(foo, ...bars, baz, ...bats) }}`. I implemented this by half-shimming with an IIFE based on Babel's output. There may be a better way, but this didn't require an extensive refactor.

Additionally, dollar args (`$1`) and `arguments` refs are now allowed to have children as per #2311. These are handled in the same way as `event` refs with children.

### Comments specifically requested

**Updated**

1. ~~Method events are handled in their current state my detecting basically `^someName(.*)\s*$`, prepending `@this.`, and handling as an expression. This seems to work pretty well in practice, but is it confusing that `toggle('foo')`, `.toggle('foo')`, and `~/toggle('foo')` mean potentially very different things? With the confusion surrounding where method events actually call, should they just be deprecated and eventually removed in favor of `@this.method(args)`? Deprecating and removing plain method call events would eliminate the remaining inconsistency in template reference resolution.~~ Consensus is that non-specific calls should be removed, so they are deprecated here and would be removed later.

2. ~~I haven't addressed proxy events in this PR yet, because I'd like some feedback first. I would prefer to support only no-args events (option 3) for auto-proxying, which would be super simple to support here. Is that what most people would like to do? The only other thing I consider tenable from a parsing (code and mental) viewpoint is to support `on-click="nameExpression:argsList"`. That would make `on-click="{{ foo ? 'bar' : 'baz'}}: {{bop}},bip,{{ { oh: 'my' } }}"` become `on-click="( foo ? 'bar' : 'baz' ): bop, 'bip', { oh: 'my' }"`.~~ Argumented proxy events are deprecated here and would be removed later.

3. ~~I also haven't addressed decorator and transition args yet, because I'd like some feedback first. I would prefer to drop mustaches in the directives entirely and just move to comma separated expression lists for decorators and transitions. That combined with `as-decoratorName` and `transitionName-in`/`transitionName-out`/`transitionName-in-out` (or something similar) and the changes in #2369 would allow that. This would be more consistent with event directives. Another proposal has been to treat them as method calls instead e.g. `decorator="decoratorName(arg1, arg2)"` which seems reasonable, but I think it's less consistent and could get weird with transitions with their in/out/in-out facets.~~ This is addressed in #2369.

4. ~~How should deprecation be handled?~~ I'd kinda like for this to land for 0.8 now, since it's still compatible with 0.7 and would let the big cleanup start immediately after 0.8.

### Status

* [x] Handle method events as expressions
* [x] Account for spread, `arguments`, and dollar args in expressions
* [x] Allow children of dollar args and `arguments`
* [x] Allow spread args in any invocation (it's there, so may as well use it, no?)
* [x] Tests ~~, tests, and more tests~~ - the deprecation path leaves the old proxy handling code in place, so only expression-related stuff is tested beyond the existing event tests
* [x] Finalize deprecation path
* [x] ~~Clean up code~~ mostly deferred until after deprecation cycle - there's lots of code that can go away and/or be nicely refactored


fskreuz:
> With the confusion surrounding where method events actually call, should they just be deprecated and eventually removed in favor of `@ractive.method(args)`?

I'm in favor of this one. But I'd suggest `@instance`. `@ractive` looks like a reference to the `Ractive` global. Thinking something like `@ractive.version` and stuff like that.

> I haven't addressed proxy events in this PR yet, because I'd like some feedback first. I would prefer to support only no-args events (option 3) for auto-proxying, which would be super simple to support here. Is that what most people would like to do?

Method calls all the way! Burn the old event syntax! :fireworks: :fire: :fire_engine: 

> Another proposal has been to treat them as method calls instead e.g. decorator="decoratorName(arg1, arg2)" which seems reasonable, but I think it's less consistent and could get weird with transitions with their in/out/in-out facets.

I method calls, possibly just args I guess?

```xml
<div transition="transitionName({ in: { duration: 5 }, out: { duration: 10 }})></div>
```

...or maybe multiple setup calls

```xml
<div transition="transitionName('in', 5); transitionName('out', 10);"></div>
```

Not sure how transitions are written, just throwing in ideas.

> How should deprecation be handled?

Partial support in 0.8 with warning log, remove from code by 0.9. Isn't that the usual mo?

> Even though it's longer, @ractive.toggle('foo') is much more clear to me when looking at a template.

Explicit is always better than implicit. :smile:

guilhermeaiolfi:
What @fskreuz said. ```@ractive``` doesn't seem to imply the current instance. ```@this``` was discarded? ```@current``` maybe?

heavyk:
I really `@instance` to me, it seems much more intuitive than `@ractive` -- which I would assume would be the same as `Ractive` - as in `Ractive.extend()`

evs-chris:
I originally went with `@ractive` because just about every piece of code I've seen in the wild using Ractive has `var ractive = new Ractive(...)` somewhere towards the top. I hadn't considered that `@ractive` would point to the constructor because it's all lower case. I would think that if that were the case, it would be `@Ractive` based on convention (jshint, eslint, etc). Also, I think of `@instance` as a bit generic, so it doesn't stand out as strongly as a _Ractive_ instance to me.

At any rate, if that is to be changed it needs to be soon, because `@ractive` has already been out in edge for about two months but would only have been used by people who watch PRs pretty closely. This is probably as good a place as any since it is proposed to make the special ref mandatory for event methods... anyone have any strong preference? After some consideration, I still prefer `@ractive`.

sabob:
For me <on-click="foo()"> is nicer than <on-click="@ractive.foo()">, but only because it has been entrenched that foo()  sits on the Ractive instance.

I understand the need to align events to resolve similar to references, so I think this is a good change overall.

Once folk gets used to calling @ractive.foo() maybe we can revisit shortening it to just @foo(). I appreciate the arguments against adding symbols to the template, but once @ractive.xxx is entrenched, maybe we would appreciate a shortcut more ;-)

heavyk:
actually, I kind of like @sabob 's idea that all `@` would reference the instance which is similar to coffeescript's convention to use `@foo` -> `this.foo`. but since the `@` is reserved for special references (like `@global`), maybe instead what if it were `@@` so, `@@foo()` would be the same as (what is now) `@ractive.foo()`

fskreuz:
@evs-chris I'm convinced :smile: Noted down in the docs repo to document the different `@` things.

@sabob @heavyk The use of `@` isn't native to JS so it might be weird to newcomers. Not everyone writes CoffeeScript either. I think it should be left as `@something` instead of standalone `@` to give it more meaning.

heavyk:
yeah I agree with `@something` convention now that I think about it.. I guess my my order of preference would probably be:
1. `~/foo()`
2. `@self.foo()` or `@instance.foo()`
3. `@ractive.foo()`

martypdx:
1. Using `@something` to refer to component instance 

    a. I agree with @evs-chris that `var ractive = new Ractive(...)` provides a cognitive connection to `@ractive`.  

    b. Current method events, while now familiar to us, don't fit with the rest of data refs in ractive. Looking at: `on-click="clicked( foo )"` why would I know that `foo` referred to data context but `clicked` referred to component context? When `@ractive` means the current component instance _across the board_ then it should be _easier_ for new people coming to ractive (though a bit of a migration road bump for us old-timers :) ): `on-click="@ractive.clicked(foo)"`.

    c. We will get tired of typing `@ractive.foo` and `@ractive.root.user` and it will become an eyesore in our templates to have `@ractive` all over the place, which is why I introduced #2345 to introduce `@foo` as a shortcut for `@ractive.foo` and `@@foo` as a shortcut for `@ractive.root.foo`. Yes, it makes the built-in `@keypath`, `@index` a bit odd - but reserved keywords are nothing new.

2. Proxy-event shortcut: Agree that having `on-click="handler"` as a shortcut for `on-click="@ractive.fire( 'handler', event )"` (assuming I'm understanding @evs-chris correctly) makes a lot of sense. Events bubble and fit a set of use cases different than method handlers. With the use of `this.event`, a very large percentage of events can be dispatched this way. Having it be a shortcut for `@ractive.handler()` doesn't really save you much, especially if you can write `@handler()`

3. Plugin syntax
    a. Agree that if all element plugins use the attribute name to communicate the plugin type and name it makes things more consistent: `on-click`, `as-tooltip`, `fade-in`.
    b. Event handlers _are_ fundamentally different than other plugin attribute values because they are  arbitrary expression to _respond to an event_, whereas decorators and transitions are _intialization values_ for the plugin. It will make decorators and transitions more terse if the attribute value is interpreted as an argument list: `as-thing='expression'` where `'[' +  expression ']'` is what gets "interpreted" in the parser and passed to `fn.apply( ractive, args )` when actually evaluated. It enables both single or double value passing via `="foo"` or `=" qux, 'BAR' " as well as `options` pattern js: `="{ foo, bar: 'BAR' }".

4. Begin the deprecation warnings!

heavyk:
@martypdx hmmm, I actually never cognitively connected that inconsistency with the data and instance methods (and I've written a lot of ractive now!!). I wonder how I missed that.

so what you're proposing then is for it to instead be `@clicked( foo )`?

TBH, I could dig that... I do like that syntax introduced in #2345, a lot :+1:

evs-chris:
@martypdx regarding 2, after the deprecation, yes, I would have the parser rewrite the argument of `on-click="([^(]+)"` as `@ractive.fire($1, ...arguments)`. In other words as long as it's obviously not a call, it would reproxy the named event with any arguments passed (including `event`, if present).

guilhermeaiolfi:
@fskreuz considering the syntax proposed in: #2345. I think ```@this``` would make a lot of sense instead of ```@ractive```. The same way that ```this.foo``` would ref the ```foo``` in the context and ```@foo``` would ref the ```foo``` in the ractive instance. ```this``` would ref the context and ```@this``` would ref the instance. Consistency!

martypdx:
@guilhermeaiolfi I'm starting to like to the idea of `@this` too

heavyk:
let's just alias `@this` to `@ractive` and call it good. everyone wins and no deprecations for us old-timers :)

JonDum:
I've got to third the `@this`.

JonDum:
Although, If we do go with `@this`, do we even need the `@`? 

call data methods: `on-click="loadStuff()"` or `on-click="../loadStuff()"` or `on-click="~/loadStuff()"`
call proto methods: `on-click="this.loadStuff()"`
call complex structures `on-click="this.reset.partial('foo')"` or `on-click="this.foo().and.bar({{this}}, 'yep')"`
call parent methods `on-click="this.parent.doAThing({{this}})"`
chain multiple actions `on-click="this.toggle('foo') && this.fire('fooed')"`

It kinda works, and avoids the `@` completely, but uses `{{ }}` mixed in to pass expressions... which may be inconsistent with how expressions are used in other places (but this inconsistency already exists < 0.8).

martypdx:
> Although, If we do go with @this, do we even need the @?

Mixing mustaches inside an expressions has been a PITA, both from an implementation perspective and IMO as a developer because you end up in "mixed-mode" javascript and html template land. The goal is to use mustaches for html and text interpolation, allow javascript expressions _inside_ mustaches and  inside plugin attribute directive values.

The other benefit to use a marker instead of assuming the first token of an event handler applies a method name is that the expression itself can more more dynamic as multiple actions can be invoked:

```html
on-click="@this.set( 'x', foo ), @this.set( 'y', bar )"
on-click="@this.push( 'items', newItem ).then( _ => @this.set( 'newItem' ) )"
```
[note: AFAIK, neither of those examples is currently supported even with this PR, @evs-chris?]

While arguably there is an initial learning in having both `this` and `@this`, `@` would be recognizable as a special keyword. Ultimate syntax test would be something like: 

```html
{{#each items}}
<li on-click="@this.select( this )">{{this.name}}</li>
{{/each}}
```

I was going for succinctness in shortening that to `@select` but I could live with `@this.select` and it avoids the javascript weirdness of concatting `obj.prop` down to `objprop`.

evs-chris:
@martypdx no, but `on-click="@ractive.set('x', foo) && @ractive.set('y', bar)"` is. I don't know what @RichHarris' stance on arrow functions in the template is, but plain functions where explicitly excluded. I've found myself wanting simple, single statement arrows for mapping and the like before, but it's usually better to pull those out of the template. In-template functions are an entirely different 50 gallon drum of worms, though.

martypdx:
> no, but `on-click="@ractive.set('x', foo) && @ractive.set('y', bar)"` is

That's what I thought, but that's a bit of a hack and feels error prone when people use them with functions that have a falsey return. Maybe allow the comma operator so this expression can be properly, um, expressed: `on-click="@ractive.set('x', foo), @ractive.set('y', bar)"`

> In-template functions are an entirely different 50 gallon drum of worms, though.

yep. and future drum of worms for sure.

evs-chris:
It's not exactly the comma operator per se, but I added support for a list of expressions. The last value in the list is the one used for cancelling.

evs-chris:
After looking at it for a bit, I decided `@this` is probably not too confusing and reads and types a little better than `@ractive`. I personally like #2345, but I can't really gauge it's overall reception at this point. I could see how `@this` makes the possibility of it not landing a bit more palatable.

I also added deprecation messages for old-style method and proxy-with-args events, as the consensus seems to be to move towards consistency and the existing method events are not consistent. Also, dropping method events and making proxy events simple enough to easily parse them into an expression would let a fair amount of code be put out to pasture after the deprecation cycle.

dagnelies:
I strongly agree with @fskreuz's comment:

>  `this` means the current context in templates (synonymous with `.`). Using `@this` to mean the current instance can make it confusing.

Making `this` and `@this` with totally different meanings is just confusing!

sabob:
```@this``` and ```this``` doesn't sit well with me either.

If my template ends up as: ```<li on-click="@this.select( this )">```, I would go back and remove both ```this``` if possible to ```<li on-click="@ractive.select( . )">```

Certainly @this is shorter to type than @ractive, but it means I have to stop and think about what the different ``this`` things mean in my template (or someone else's)

Also it seems once we switch to ```@this``` and ```.this```, ```this``` becomes meaningless and the useful bits are really the ```@``` and ```.```, ```this``` can be removed.

I'm still in favour of the longer @ractive and see where things go from there. There are a couple of releases before v1 to iron things out.

**Comment by evs-chris** [comment]
I like `@this` because it gels with the JS meaning. It _is_ a bit close to `this`, but then any special can have a same-named reference in scope at any given point in time. I've found that the `@` is a darn good visual cue on meaning because it really stands out in templates. The only other suggestion that looked ok to me was `@self`. I sometimes escape anonymous function binding issues when I'm ES5-ing it with `var me = this;`, so `@me` looks fairly natural to _me_ too.

So what would the vote be for the Ractive instance special ref: `@this`, `@ractive`, `@self`, `@me`, or something else? I like `@this`, then `@me`, with a tie for third with `@ractive` and `@self`.

**Comment by skeptic35** [comment]
For me the answer lies in the question:
> So what would the vote be for the Ractive instance special ref

`@instance` as suggested by @fskreuz 
It's barely longer than `@ractive` and it's totally clear what it means (as opposed to any other suggestion, imho)

**Comment by sabob** [comment]
My order of preference is @ractive.something then @this.something, and if all goes well, at a later stage just @something

As the joke goes "if two developers agree during a meeting, that is a majority".

At the end of the day once a decision is made and documented I don't think it will matter much which keyword we use.

**Comment by dagnelies** [comment]
**Summary**

Problematic names which may cause confusion:
- `@this`: too close to `this` used to denote the current context in expressions (same as `.`)
- `@ractive`: one may think it refers to the `Ractive` global instead of the instance

Suggested so far (in alphabetical order to not make anyone jealous ;) ):
- only `@` or `@@`
- `@current`
- `@instance`
- `@me`
- `@self`

My vote goes to `@self` and `@instance` (tie).

Side note of existing symbols (as far as I know, independent of the context):
- `@index`
- `@keypath`
- `@global`
- `@key`

**Comment by guilhermeaiolfi** [comment]
What the problem of ```@this``` being too close to ```this```? Both are pretty close in meaning too. 
For people voting, have you read #2345? 

I know that it's a little too radical (and maybe late) and have a hard chance of acceptance but I would like to share two ideas:

1) ```this``` (without @) to always ref the ractive instance and let just ```.foo``` to ref the context. It's more aligned with OOP. ```@this/@ractive/@me/whatever``` seems a little off considering others symbols like ```@index``` and ```@key```. 

2) Since v0.8 it seems the recommendation seems to use ```{{.foo}}``` to avoid unnecessary lookups to get better performance. But using the context data seems the most used case in templates. So why not make it the default option? So you'd have ```{{foo}}``` (without any special char) to always be a reference to the context and something like ```{{^foo}}``` an expression that goes up to the tree looking for references. That way you know when you can get data from outside you component. That would be easier for people coming from others mustache based frameworks too and IMO easier to read.

**Comment by dagnelies** [comment]
> What the problem of `@this` being too close to `this`? Both are pretty close in meaning too. 

They are not close at all. `@this` would denote the current ractive instance, while `this` denotes a nested object belonging to `data`.

> `this` (without @) to always ref the ractive instance and let just .foo to ref the context.

The only issue is that `this` is part of ractive since years. Changing it would be a major (read catastrophic) compatibility break.

**Comment by evs-chris** [comment]
FWIW `this.foo` and `.foo` parse to exactly the same thing.

**Comment by guilhermeaiolfi** [comment]
> They are not close at all. @this would denote the current ractive instance, while this is an already existing keyword denoting a nested object belonging to data.

So, same idea, different scope.

> The only issue is that this is part of ractive since years. Changing it would be a major (read catastrophic) compatibility break.

I know. That why I think it's hard to be accepted. But heyy, 0.8, now is the time to change things before 1.0. Let's make it right.

> FWIW this.foo and .foo parse to exactly the same thing.

Sure. I was purposing to change that.

I can leave with any of the options. But I think we can do better if we are willing to break compatibility. Thinking in the long run, now is the time. Let's decide for the future and not for the present.

**Comment by evs-chris** [comment]
I think it would be best to stay consistent with references, where `@` means "this is a meta reference that has something to do with where I am in a template construct and nothing to do with data". I think swapping `this` to mean `@this` would be considerably more confusing to newcomers based on issues that I've seen over the years. Having two distinct classes of refs is a bit easier to track:

| things that are in my data | things that are semi-magical meta-references injected by the template | 
|---|---|
| `.` - current context | `@index` - the current index of the nearest iteration |
| `this` - more visible alias for `.` | `@key` - the current key of the nearest object iteration |
| `foo` - foo somewhere in data - or - an ambiguous reference that may resolve with an alias, a named index reference, or at any point up the context's hierarchy | `@keypath` - the current context keypath relative to the component |
| `.foo` - foo property of current context | `@rootpath` - the current context keypath relative to root |
| `../foo` - foo property of context parent | `@global` - the global object for the environment |
| `~/foo` - foo at root of data | `@this` or `@ractive` or `@something` - the Ractive instance for the nearest component or just the root instance if none |
| ??? | |
| Profit! | |

At any rate, would it be better to make the `@ractive` special ref configurable so that each person/team can swap it for whatever is more meaningful/visible/obvious to them? Personally, I'm opposed to anything longer than `@ractive` for purposes of concision in templates. I frequently have elements that have multiple event handlers and I'd like to avoid getting past column 200 whenever possible.

**Comment by guilhermeaiolfi** [comment]
> At any rate, would it be better to make the @ractive special ref configurable so that each person/team can swap it for whatever is more meaningful/visible/obvious to them?

I prefer it to be something that I don't like now but are consistent between projects than have to keep this one more thing in my mind when going from one project to another because the previous team decided that ```@mypuppy``` thought it would be cool.

**Comment by JonDum** [comment]
> Problematic names which may cause confusion:
> 
> * `@this`: too close to this used to denote the current context in expressions (same as .)
> * `@ractive`: one may think it refers to the Ractive global instead of the instance

`@ractive` seems so weird. If it's derived from `var ractive = new Ractive(...)`, that `ractive` var is something used outside of the prototype/template, elsewhere on the page—possibly with other instances of Ractive. By that logic we could also call it `@main` or `@app` since those are frequently used (`var main = new Ractive()`). And that would be _really_ weird. 

I still think `@this` is most succinct, but `@instance` isn't bad either. `instance` is also used all over the docs and in example decorator code, `var instance = this`.

screen shot 2016-02-23 at 2 12 48 pm

Makes sense. When you're trying to call an "instance" method, you'd use `@instance`. 

`@this` also makes sense because something on the "instance" is really on the prototype, and if you're familiar with prototypal calls, you use `this` to reference something. 

I guess my vote would be for both `@this` & `@instance` but not `@ractive`? 

¯\_(ツ)_/¯

**Comment by aharpervc** [comment]
`@this` is too confusion-prone. What about `@context`?

**Comment by fskreuz** [comment]
@aharpervc Then this question comes up: "context to what?". The word "context" also has special meaning in Ractive, i.e. ["the context stack" in templates](http://docs.ractivejs.org/latest/mustaches#restricted-refs).

**Comment by heavyk** [comment]
this is going nowhere fast :) -- what if it were `@specimen` instead? while examining my code, I like to think of myself as a detective examining specimens of my work :D

**Comment by dagnelies** [comment]
oh, come on folks! Let's keep it concise!

.@evs-chris : for the sake of sanity, please just pick one you think adequate.

**Comment by aharpervc** [comment]
@fskreuz yeah, I was intentionally paralleling the use of "context" as in the right side column in @evs-chris's table

**Comment by evs-chris** [comment]
Coming back to it after a bit of time has elapsed, `@this` still looks good, so I'm going to leave it at that.

I've also updated the intro comment with the result of the other bits of discussion here. The general consensus seemed to be that this is a good way forward, so any thoughts on getting this in for 0.8 or 0.9 (or not at all)?

**Comment by guilhermeaiolfi** [comment]
So much goodness to wait until 0.9. Since it's ready, ship it. 0.8 from me.

**Comment by JonDum** [comment]
0.8, ho!

**Comment by fskreuz** [comment]
0.8 is good from this end :+1:

**Comment by martypdx** [comment]
I've been coming back to something @guilhermeaiolfi said: 

> 1) this (without @) to always ref the ractive instance and let just .foo to ref the context. It's more aligned with OOP. @this/@ractive/@me/whatever seems a little off considering others symbols like @index and @key.

If we put aside the BREAKING CHANGE aspect, I prefer an API where `{{this}}` always referred to the ractive instance, both in mustache references *and* in event handlers.

You could still use `{{.}}` to mean the model context.  In fact you could argue it's more consistent `{{@this}}` means current context, like current @index, @key, @keypath. Plus with aliases in `0.8` you have other options for naming model contexts beyond "this".

Thoughts?

**Comment by JonDum** [comment]
@martypdx No complaints about that. I concede to whatever Chris and you like there.

If you follow that train of thought, shouldn't all data/context refs start with `@` and everything else be reserved keywords?

So...

`@global` -> `global`
`@root` -> `root`
`@this` -> `this`
`this` ->  `.`'

To remain consistent

**Comment by evs-chris** [comment]
@martypdx well I'm not really for it, but I'm not really against it either. I like that `@` stuff stands out as 'not in my data' in the template. I typically use the `.` form of scoped references e.g. `.foo` rather than `this.foo`, so it doesn't really affect me on that front. I don't know how often people use `this` in the wild, but it's not in either mustache or handlebars as far as I can tell. From what I've seen from issues, more people tend not to use scoping at all. It looks like `this` was added in b74fb4df0d8189be46b7472dd675ea4ae937161f so it's been around for a while with its current meaning.

All that said, I think it would be pretty easy change to make. Anyone else (especially @Rich-Harris)?

**Comment by martypdx** [comment]
@evs-chris, @JonDum for the record, I'm not actually advocating `@this` needs to get added under that scheme, just pointing out the option exists if people think `{{.}}` is inadequate.

**Comment by martypdx** [comment]
btw - `self` is the apparently the preferred way to get to global scope across browser, web and service workers, and node. I wonder if we might keep things simple via these mapped changes:

```
this -> .
@global -> self
@this -> this
@root -> this.root
```

Only first item would be breaking change as others have not been introduced.

**Comment by heavyk** [comment]
good point (especially the service workers), I think that mapping to is quite reasonable.

is there some way that could be combined in some way with your state idea in #2345 too? maybe we can define our own `@` vars? or, heaven forbid a `@state` variable? (prolly not) I've had multiple occasions now, where having access to state would have been a more desirable approach (even though it's simple enough to add state from an external object into the template)

**Comment by fskreuz** [comment]
Thread becoming long, adding just one more to hopefully seal the deal. Let's all agree to [this proposal](https://github.com/ractivejs/ractive/pull/2386#issuecomment-187807506)):

- Anything **without **`@` means it's a valid JS expression.
- Anything **with **`@` means it's something Ractive-specific.

It becomes clear when you put it like this:

> I like that @ stuff stands out as 'not in my data' in the template.
>
> https://github.com/ractivejs/ractive/pull/2386#issuecomment-215620950
>
> this is a meta reference that has something to do with where I am in a template construct and nothing to do with data
>
> https://github.com/ractivejs/ractive/pull/2386#issuecomment-187807506

That should be enough of a guideline to avoid confusion. We should steer away from keyword-specific to a broader view (had to re-read the thread to realize). A doc page for `@`-things should also be sufficient, sort of like a glossary.

Really looking forward to 0.8 (as if I weren't using edge all this time).

**Comment by martypdx** [comment]
@evs-chris @fskreuz `@this` it is!

I think we should merge this now, because it changes or supports a number of other features that are already in edge (like `@ractive`)

**Comment by kouts** [comment]
Is the '@this' prefix required from now on?
Does writing ```on-click="fire('eventNameToBeFiredOnThis', event)"``` still work?

**Comment by guilhermeaiolfi** [comment]
@kouts yes, it works without ```@this```: http://jsfiddle.net/p9hbej0k/

**Comment by martypdx** [comment]
@kouts, @guilhermeaiolfi Though it is deprecated if you check the console: 

```
Ractive.js: Unqualified method events are deprecated. Prefix methods with '@this.' to call 
methods on the current Ractive instance.
```
whereas default firing by name w/o args is still supported:

```
on-click="eventNameToBeFiredOnThis"
```

**Comment by kouts** [comment]
So in order to be future proof should I add the '@this' to all my event calls inside my templates?
Wouldn't it be better if the '@this' part was optional?

**Comment by guilhermeaiolfi** [comment]
Yeah, I think I said that somewhere. Not just for events, for properties in general it would be nice to avoid using '@this'. I would like to avoid those features that differ ractive from pure javascript. The exception being when there is analog way of doing that in javascript.

**Comment by JonDum** [comment]
I've been going through all my components preparing them for this change.

Now having used it extensively... I'm *really* not a fan of all the clutter. 90% of the time I'm just doing a `set()`. It's far more uncommon for me to have methods in my data.

> Wouldn't it be better if the '@this' part was optional? 

+1 to that. If you did want to call a method on your data, can't you use `.foo()`, `../foo()`, `~/foo()` anyways? If we don't deprecate the old syntax and just let it remain everything works pretty dandy:

http://jsfiddle.net/ddxgxupr/

(except for the `{{#with foo}} {{ .test() }} {{/with}}` that doesn't appear to work, may just be a bug)

**Comment by evs-chris** [comment]
I'm personally opposed to anything that makes the same reference resolve to different things in different contexts. My designer buddies have a hard enough time with one set of rules :laughing:

I think we should land something like #2345 that, among other things, makes `@set('foo', bar)` equivalent to `@this.set('foo', bar)`. Would that get close enough?

**Comment by kouts** [comment]
I've already switched to @this, but an alternative is always nice to have.

**Comment by dagnelies** [comment]
well, I'm not fan of `@this` because I still think people will confuse it with `this`, which have *totally* different meanings.

Now you can do this:
```
<button on-click="@this.do(this)">foo</button>
<button on-click="this.do(@this)">bar</button>
```
If I'd be a noob, I'd just be confused :astonished:

heavyk:
I'm a huge fan of the #2345 PR (but more for the state part about it). the `@` operator feels natural to me cause I've spent a lot of time writing livescript. it's also the same in coffee-script: `@set 'lala', 2` is shortcut for `this.set('lala', 2)`

that  being said, I believe @JonDum brings up a good point. in my code, I almost never use `@` variables in the template (cept when I need to access global state) and I've always used `~/foo()` syntax instead. so, perhaps the `@` variable isn't really needed so much. I dunno... personally, I think `~/foo()` is the most 'natural' for me, cause `~/` makes me think of referring to `$HOME` in bash.

Issue #1980, opened 2015-05-30
warn on unresolved references


Rich-Harris:
Doing some performance optimisation for a large project with several contributors, I found it to be very useful to add some logging in [here](https://github.com/ractivejs/ractive/blob/dev/src/global/runloop.js#L126-L155) that warns me about unresolved references, since these need to be checked on each data change. If you have many of them it can therefore slow things down a fair bit. The problem might go away with #1740 (@martypdx?) but if not then it would be good to add some proper logging around this, with guidance for developers that helps them understand what's going on and how to optimise their apps

martypdx:
> The problem might go away with #1740 (@martypdx?)

Yes, I changed how unresolved refs work because the runloop was trying to re-resolve _every_ unresolved on _every_ runloop end which, as you point out, leads to perf problems with large number of unresolved.

In `0.8` (#1740) there's the ability to add watchers to binding contexts. An unresolved reference will add a watcher to each context in its context stack that will be called if a matching key gets added to that context (first match wins, others watchers in the stack are removed once resolved). So now instead of trying all unresolved refs on every data change, a data change that adds a child to a context with a matching key _notifies_ the unresolved with its resolved context.

Watchers are also used for pattern observers and for object hash iteration to notify on any new child add.

Been meaning to add some perf tests to measure this, would welcome good test case example for current perf problems, though I suspect adding a deeply nested section with couple hundred items in a list that each have an unresolved (or share same unresolved for that matter) would do the trick.

ADantes:
There is value in warning about ambiguous references, independent of the performance impact.  I've had more than a handful of cases where things didn't work quite as expected, unless my data was properly initialized, so it would be a very useful tool for debugging.  It should also make code substantially easier to hand off when working on a large team.

evs-chris:
There's now a flag, `warnAboutAmbiguity`, on edge.

PR #2394, opened 2016-02-21
Housekeeping logging WIP

fskreuz:
Mostly just moving around stuff.

- Moved error and warning messages to `src/messages` for easier lookup and update of messages and so that verbiage changes for `better error` issues can easily be addressed.

- While I was at it, moved out constants to `src/constants` because being in `config` was confusing. Even stuff in `config` is confusing because they're not configuration. They're more of environment checks.

- Also updated pre-flight checks. There were 3 functions that were not being checked (`reduce`, `bind` and `every`) and also checking for things that could potentially be pulled out into an optional polyfill.

- Turned `PROMISE_DEBUG` warnings into `warnIfDebug` so they look consistent with other warnings as a `console.warn` instead of `console.log`.

- Debug mode welcome messages now show everything instead of a collapsed warning. That way, users can easily see that there are help links, enabled debug modes, etc. etc.

ToDo:

- [x] Updated pre-flight checks (required methods, debug mode warnings etc.)
- [x] Messages in one location
- [x] Constants in one location
- [x] `log` only used in `PROMISE_DEBUG`, replace with `warnIfDebug`.
- [ ] Convert `throw new Error` into `fatal`  or convert `fatal` into `throw new [SpecificErrorConstructor]`
- [ ] Convert `console.warn` into `warn` calls
- [ ] Look for instances where the other logging functions could be used
- [ ] Warn on unresolved references https://github.com/ractivejs/ractive/issues/1980

<sub>Was supposed to be trying to remove legacy and polyfills, but this looked like an easy win. :smile: </sub>


fskreuz:
Rethinking about inline errors and stuff.

PR #2398, opened 2016-02-23
use local variable with for...in to avoid deopt

Rich-Harris:
Chrome was barking at me about `Ractive.prototype.set` being deoptimised ('ForInStatement with non-local each variable'). This fixes it

### DIFF
--- a/src/Ractive/prototype/set.js
+++ b/src/Ractive/prototype/set.js
@@ -10,9 +10,9 @@ export default function Ractive$set ( keypath, value ) {
 	if ( isObject( keypath ) ) {
 		const map = keypath;
 
-		for ( keypath in map ) {
-			if ( map.hasOwnProperty( keypath) ) {
-				set( this, keypath, map[ keypath ] );
+		for ( const k in map ) {
+			if ( map.hasOwnProperty( k) ) {
+				set( this, k, map[k] );
 			}
 		}
 	}


evs-chris:
:+1:

src/Ractive/prototype/set.js
```diff
@@ -10,9 +10,9 @@ export default function Ractive$set ( keypath, value ) {
 	if ( isObject( keypath ) ) {
 		const map = keypath;
 
-		for ( keypath in map ) {
-			if ( map.hasOwnProperty( keypath) ) {
-				set( this, keypath, map[ keypath ] );
+		for ( const k in map ) {
+			if ( map.hasOwnProperty( k) ) {
+				set( this, k, map[k] );
 			}
```
martypdx: Same as?

```js
Object.keys( map ).forEach( k => set( this, k, map[k] ) );
```

Rich-Harris: ah, merged before I saw this note. I *think* the `for...in` is microscopically faster; I could be mistaken – anyone know for sure?

evs-chris: Definitely is for polyfilled situations, and barring some interesting optimization voodoo, should be everywhere else too. Fewer ops.

Rich-Harris: Huh – seems the fastest might actually be a combo, as you avoid both the function overhead and the cost of calling `hasOwnProperty`:

```js
const keypaths = Object.keys( map );
let i = keypaths.length;
while ( i-- ) {
  set( this, keypaths[i], map[ keypaths[i] ] );
}
```

Via https://jsperf.com/object-keys-vs-for-in-with-closure/3

martypdx: When last I measured (couple years ago) it was a wash when you consider then `.hasOwnProperty` call. 

Here is info on optimization blocking: https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#5-for-in

martypdx: > const keypaths = Object.keys( map );
> let i = keypaths.length;
> while ( i-- ) {
>   set( this, keypaths[i], map[ keypaths[i] ] );
> }

yes, that should be the fastest :) - consistent with all the testing I did around keys and iterations

MartinKolarik: Yes, `Object.keys()` and `for`/`while` is definitely the fastest, because native `forEach` is very slow compared to normal `for`. That said, we use `forEach` all over the place, so it probably doesn't really matter.

Rich-Harris: Maybe there's some value in replacing `forEach` at some point then, particularly in hot code paths where speed > readability. Though catching other places where functions are being deoptimised is probably lower hanging fruit

heavyk: :+1: I think it would make a difference, honestly

heavyk: one could find out pretty easily to kinda get an idea how much of a difference actually by first redefining the `Array.prototype.forEach`

just a minute, I'm gonna try it :)

heavyk: totally unscientific, but one of our functions spends about 6-7ms to update formally. however swapping out the prototype method reduces time spent to around 5.1ms.

certainly an improvent.

heavyk: put this at the top of your main file to see for yourself:

```js
Array.prototype.forEach = function (fn) {
  for (var i = 0; i < this.length; i++) {
    fn(this[i], i)
  }
}
```

additionally converting to a normal for-loop will reduce the number of functions needed drastically, further improving performance.

MartinKolarik: @heavyk IIRC I've tried that a while ago and the difference was rather small, but feel free to give it a shot.

@Rich-Harris we could actually have both speed and readability in this case, if we used our own `forEach` implementation.

heavyk: doh! you guys are talking about for-in loops... uhhhh lol

well, this optimisation still exists nonetheless for normal arrays

MartinKolarik: @heavyk `for-in` can be replaced by `Object.keys() + for`, so it applies to that as well.

heavyk: @MartinKolarik it does make a huge difference for me. if I count how many times that function is called:

```js
window.foreach = 0
Array.prototype.forEach = function (fn) {
  window.foreach++
  for (var i = 0; i < this.length; i++) {
    fn(this[i], i)
  }
}
```

and then look at `window.foreach` after the page loads I get the value 5460 -- so on our codebase it's called quite a lot. if I specifically record the number of times it's called in that one update function, it's called 192 times for that 20% speed improvement.

Issue #2397, opened 2016-02-23
Numbers lower than 10 undefined when passed into partial's "as" context


woodlandhunter:
See fiddle http://jsfiddle.net/478xdvqb/

```xml
<main></main>

<script id='template' type='text/ractive'>
    {{>test-partial 1 as numa, 10 as numb, 20 as numc}}
</script>

<script id='test-partial' type='text/ractive'>
	<div>A:  {{numa}}</div>
	<div>B:  {{numb}}</div>
  <div>C:  {{numc}}</div>
  <div>C1:  {{.numc}}</div>
</script>
```

```js
var ractive = new Ractive({
    el: 'main',
    template: '#template',
});

```

Also a secondary question regarding "as" context and using restricted `{{.keypath}}` (see the C1 line in the fiddle). Should that work?

evs-chris:
Hmmm... `numa` should work. You've run across one of those parsing bugs I keep making a note to hunt down and keep forgetting. It just won't parse a single digit literal int. You can usually work around it by adding a `+` e.g. `+1`, but when you try to do that as the first expression in a partial alias, it confuses the parser. Using `{{>partial 10 as numb, 20 as numc, +1 as numa}}` is a possible workaround.

`.numc` is not resolving as intended. Aliases don't change the context. They just add names to check on the way to resolving an ambiguous reference and `.numc` is entirely unambiguous.

evs-chris:
This is really weird... I've found the cause and I'm not sure exactly how to fix it. I didn't realize that Ractive supports this `{{#with someArray}}{{1}}{{/with}}`. So basically, numbers are always treated as references to an array member. I think that's pretty surprising, but it was done on purpose. The reason `10` and `20` work is that the `refineExpression` test for array indices is a little wonky and ignored numbers ending in `0`.

So, how to fix?

1. Fix the regex and keep the old behavior.
2. Drop the old behavior and call number literals number literals.
3. Flag the reference as a potential number literal and use the literal if not in the context of an array.

I vote 2 because on the odd (in my opinion) chance that you'd want to use a number literal as a reference in an array, you can still do `{{#with someArray}}{{.[1]}}{{/with}}`. 3 supports both existing and expected behavior, but also adds a fair amount of complexity during ambiguous resolution and creating expressions on the fly for the special reference-but-not-necessarily case. So not really csp friendly. I suppose we could add global number models to handle that, but again, complexity.

fskreuz:
[...]

PR #2401, opened 2016-02-24
BREAKING: integer literals are probably not a reference - #2397

evs-chris:
This is a breaking change in that this specific case is no longer supported `{{#with someArray}}{{1}}{{/with}}`. There was a bug with the original support for that anyways that let numbers ending in `0` slip through as expressions.

If for some reason the original behavior is needed, it can be had with `{{#with someArray}}{{this[1]}}{{/with}}` or some more terse equivalent.

Any objections?


fskreuz:
Unless I'm missing anything obvious, :+1:

martypdx:
Intra-reference numbers like `{{ this.1 }}` are also okay?

evs-chris:
Yes, non-number-literal numbers take a different path as references, so `this.1` is still valid. Only places where just a number literal is parsed are affected.

heavyk:
what about `{{.1}}`?

I'm also :+1:

evs-chris:
`{{.1}}` is the same as `{{0.1}}` because numbers. I think that's always been the case, but then again I always write the leading `0`.

heavyk:
you're right. I just tried it for myself. I also always write the leading 0 for numbers, but I did try it for index access once in a loop and it didn't work. (seems obvious now)

I think that `{{0}}` is confusing to begin with.

fskreuz:
> {{.1}} is the same as {{0.1}} because numbers

So this means `{{ .1 }} !== {{ this.1 }}`? Hmm... a potential gotcha.

Issue #2400, opened 2016-02-24
Array methods not updating array index for getNodeInfo.keypath


sabob:
In this fiddle:
http://jsfiddle.net/sabob/96x93jve/

we have a button to add "bar" items to array.

When a new item is added we lookup the last node:
```var nodes = document.querySelectorAll("#items li");
var node = nodes[nodes.length- 1];```

Then we print the keypath:

```
var nodeInfo = Ractive.getNodeInfo(node);
console.log("keypath", nodeInfo.keypath);
```

In the console, note that the keypath array index of the last item is not updated. Foo keypath always prints:
```
items.0
```

In 0.7.3 it works correctly and "foo" keypath is updated when item is added:
```
items.1
```

add another item and foo keypath becomes:
```
items.2
```

etc.

Problem exists for "splice" and "unshift", probably other methods too?

Issue #2399, opened 2016-02-24
array event index not updated


sabob:
This fiddle: http://jsfiddle.net/sabob/m8325w9a/2/ shows that the event's index references passed from the template is not updated after array items are removed.

In the fiddle, click on the Delete button next to the top item, in the console note the index is 0.
Click Delete again of the new top item, in the console note the index is 1, instead of 0. Also, instead of deleting "John", "Steve" was deleted.

Works in 0.7.3

fskreuz:
BUG! A workaround for now would be to get the index from the keypath `e.keypath.split('.').pop()`.

http://jsfiddle.net/m8325w9a/4/

```js
var view = new Ractive({
	el: 'body',
  template: `
  	<ul>
      {{#each items:i}}
      <li>
        {{name}}
        <button on-click="remove(event, this)">Delete</button>
      </li>
      {{/each}}
    </ul>
  `,
  data: {
    items: [
      {name: "Bob"},
      {name: "John"},
      {name: "Steve"}
    ]
  },
  remove: function(e, obj) {
    console.log("index: %s keypath: %s", e.index.i, e.keypath);
    this.splice("items", e.index.i, 1);
	}
})
```

woodlandhunter:
Another workaround could be to pass the index directly: `on-click="remove(event, this, i)`

sabob:
Thanks, I'm actually trying to get this decorator to work in edge as I use it in 0.7.3:
https://github.com/ractivejs/Ractive-decorators-sortable

The event index not updating is one bug, but it seems there are more issues. I'll update as I go along.

sabob:
Can we tag this issue under milestone 0.0.8?

sabob:
fixed by #2402

PR #2402, opened 2016-02-24
PRIVATE BREAKING: Look up template refs for node info and events on the fly - RFC

evs-chris:
Template refs for indices and keys are currently forwarded along through the fragment hierarchy as it gets created. This seems like basically trading performance for memory, which is fine, but it also introduces a maintenance burden for splicing. That maintenance is not currently being done per #2399 and #2400.

This PR drops the forwarding of refs through and instead looks them up when needed e.g. when an event is fired or when `Ractive.getNodeInfo` is called. I looked briefly at trying to keep the `indexRefs` and `keyRefs` objects up to date when a shuffle happens, but it looked like more code than it's worth for the two places they're actually used. The only potential downside is that gathering the refs for a very very deep template on event firing might add a very slight delay to the event handler being called.

I also added the `keys` to the `event` object because it was free. That brings around the questions of whether there should actually be separate objects for indices and keys, and whether aliases should also be thrown in, since that would be basically free with this approach.

Thoughts?

### DIFF
--- a/src/Ractive/initialise.js
+++ b/src/Ractive/initialise.js
@@ -41,9 +41,7 @@ export default function initialise ( ractive, userOptions, options ) {
 		ractive.fragment = fragment = new Fragment({
 			owner: ractive,
 			template: ractive.template,
-			cssIds,
-			indexRefs: options.indexRefs || {},
-			keyRefs: options.keyRefs || {}
+			cssIds
 		}).bind( ractive.viewmodel );
 	}
 
--- a/src/Ractive/static/getNodeInfo.js
+++ b/src/Ractive/static/getNodeInfo.js
@@ -1,15 +1,16 @@
-import { extend } from '../../utils/object';
+import gatherRefs from '../../view/helpers/gatherRefs';
 
 export default function( node ) {
 	if ( !node || !node._ractive ) return {};
 
 	const storage = node._ractive;
+	const { key, index } = gatherRefs( storage.fragment );
 
 	return {
 		ractive: storage.ractive,
 		keypath: storage.keypath,
 		rootpath: storage.rootpath,
-		index: extend( {}, storage.fragment.indexRefs ),
-		key: extend( {}, storage.fragment.keyRefs )
+		index,
+		key
 	};
 }
--- a/src/Ractive/static/getNodeInfo.js
+++ b/src/Ractive/static/getNodeInfo.js
@@ -4,12 +4,14 @@ export default function( node ) {
 	if ( !node || !node._ractive ) return {};
 
 	const storage = node._ractive;
+	const ractive = storage.fragment.ractive;
 	const { key, index } = gatherRefs( storage.fragment );
+	const context = storage.fragment.findContext();
 
 	return {
-		ractive: storage.ractive,
-		keypath: storage.keypath,
-		rootpath: storage.rootpath,
+		ractive,
+		keypath: context.getKeypath( ractive ),
+		rootpath: context.getKeypath(),
 		index,
 		key
 	};
--- a/src/view/Fragment.js
+++ b/src/view/Fragment.js
@@ -23,8 +23,6 @@ export default class Fragment {
 
 		this.context = null;
 		this.rendered = false;
-		this.indexRefs = options.indexRefs || ( this.parent ? this.parent.indexRefs : [] );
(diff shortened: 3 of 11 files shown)


martypdx:
> also introduces a maintenance burden for splicing. That maintenance is not currently being done per #2399 and #2400.

Definitely correct splicing is top priority.

> Thoughts?

The `eventsOrInfo.index.i` api of the event/nodeInfo was always a bit hackish as it required intimate knowledge of the template (though I think the feature needs to be there for nested templates) and requires an index to be defined to get the index (no `@index` equivelent).

What if we move forward with what you have and add a deprecation warning that `.index` and `.key`, and also new api to be `.getIndex( [alias] )` and `.getIndex( [alias] )` respectively.

Then you can write:

```js
this.event.getIndex()  //default to @index !
this.event.getIndex( 'i' ) //gets the index alias defined by i
```

__AND__ it would have the added benefit of being lazy about running `gatherRefs` (and could do so in a more targeted fashion fetching just the index or key being requested).

evs-chris:
Interesting! My only concern with the lazy evaluation would be that stashing it for later might cause issues if stuff disappears in the interim. Beyond that, it basically looks like pre-anchored reference resolving from an element, so you could drop the index or key specificity and just make it `get` to find any ref available from that point in the template. See what I did there? :stuck_out_tongue_winking_eye:

Also, I'm not sure how to do a deprecation for property access unless we leave IE8 out shivering in the cold... where it belongs.

martypdx:
> My only concern with the lazy evaluation would be that stashing it for later might cause issues if stuff disappears in the interim

Does it need to be cached at all?

> so you could drop the index or key specificity and just make it `get` to find any ref available from that point in the template. See what I did there?

:astonished: 

> Also, I'm not sure how to do a deprecation for property access unless we leave IE8 out shivering in the cold... where it belongs.

Silly :rabbit:, deprecations are for developers not users. We can use `defineProperty` like we do [here](https://github.com/ractivejs/ractive/blob/dev/src/Ractive/construct.js#L37). Which only needs to work on dev machine.

evs-chris:
> Does it need to be cached at all?

Not from our end, but I was thinking the dev may stash the event object for some reason. Probably not a terribly reasonable thing to do or worry about.

> :astonished:

Heh. Maybe I'm going off the rails further here, but I think it'd be pretty slick to make the event and node info objects basically a pointer into the data with its own `get`, `set`, `splice`, etc.

> Silly :rabbit:, deprecations are for developers not users.

Well, dang. :smile:

sabob:
Thanks Chris! I was looking for where indexrefs were updated, now I know, they aren't ;-)

To fix the keypaths for #2400 should updatePostShuffle update the "fragment.key" as well as "fragment.context.key" to the index value? Or it's not so straightforward?

evs-chris:
@sabob, no, the `getNodeInfo` call should just be pulling the keypath on the fly too. I misread #2400, so I thought I had it covered. On the plus side, I think we can eliminate a little more memory footprint :smile:

evs-chris:
Since we already have everything we need for node info in the fragment, I just dropped the context and keypath stuff out of `_ractive`. It seems like more unnecessary work during an already busy process to keep those up to date during rebind. I also dropped the ractive reference because it is readily available from the fragment. Is that fair?

If this is all good, I think I'll punt deprecating and replacing the index and key maps to another PR.

martypdx:
> Heh. Maybe I'm going off the rails further here, but I think it'd be pretty slick to make the event and node info objects basically a pointer into the data with its own get, set, splice, etc.

Actually that was the "astonished" emoji - meaning it's a great idea

evs-chris:
:smile: It also neatly solves 2/3s of the issue I was trying to address with anchored set, splice, etc. I'm imagining `<button on-click="event.splice('../', @index, 1)">`, but I don't know that that's entirely reasonable.

martypdx:
Just hang the methods of `@keypath` (or add `@context` if you prefer):

`<button on-click="@keypath.splice('../', @index, 1)">`

sabob:
Should we also return the "binding" property as part of nodeInfo?

I've seen binding used in decorators such as:
https://github.com/Prezent/ractive-decorators-select2/blob/master/ractive-decorators-select2.js

where the binding and keypath is looked up via:
``` var binding = node._ractive.binding;
var keypath = binding.keypath ? binding.keypath : binding.model.key;```

Issue #2395, opened 2016-02-23
Recursive looping on firstNode of fragment with nested if blocks


dagnelies:
Hi,

I encountered a bug, which I couldn't isolate so far. I'm posting it nevertheless to keep track of it. I'll perhaps be able to track it down when I have more time... or perhaps some of you already have an idea. Following snippet on edge causes a timeout (probably because of an endless loop):

```
{{#each mylist.users}}
	{{#if !~/users.filter || .displayName.indexOf(~/users.filter) >= 0 }}
		{{#if !~/users.showSelected || .selected}}
			...
		{{/if}}
	{{/if}}
{{/each}}
```
Funnily enough, inverting the two `if`s makes it run seamlessly again.

The recursive loop is connected to:
```
findNextNode@ractive.js:6877:1
firstNode@ractive.js:13486:12
findNextNode@ractive.js:6879:18
findNextNode@ractive.js:13441:12
firstNode@ractive.js:13486:12
firstNode@ractive.js:7212:29
firstNode@ractive.js:13480:13
findNextNode@ractive.js:6879:18
...
```

dagnelies:
Never mind, errors happen independently of the `if` order. Apparently when more results have to be rendered.

fskreuz:
I assume this is edge?

dagnelies:
yes, edge

evs-chris:
@dagnelies I think I can see where infinite recursion might occur and blow the stack, but I can't seem to reproduce it. Can you give me a little more context from the template e.g. what surrounds the iterative section and the contents of the iterations?

dagnelies:
@evs-chris the content doesn't appear to play a role. Even with solely `...` it blows up. Sadly, I wasn't able either to reproduce it in a simple example. I'm currently also quite busy at work right now, but I'll come back later to it. Just give me some time.

dagnelies:
Ok, I've lastly nailed it down better. One reason why I couldn't reproduce it with a small example is because it was not an endless loop. It rather appears to require exponential time to resolve.

Here is the fiddle:
http://jsfiddle.net/ey0LxzL0/

"Instructions":
- check/uncheck the "show only selected" box, everything is fine
- change the `addUsers(20)` into `addUsers(25)` and do the same, it'll take quite a while
- change to `addUsers(30)` and the browser dies

evs-chris:
[...]

PR #2405, opened 2016-02-25
Avoid exponential firstNode search - #2395 - Review

evs-chris:
When searching for the next node in a complex repeated fragment in order to insert a new iteration, edge is currently checking each element from the parent, each element is checking the parent, and so forth until the browser gives up.

This adds a flag to `firstNode` that, when set, tells it not to punt to the parent if it can't find a node for itself. That avoids repeated sections calling children that call it back.

Thoughts?

_I fat-finger tab-enter submitted this a little early... sorry_


evs-chris:
Also managed to forget - fixes #2395

heavyk:
no thoughts. seems correct to me

PR #2407, opened 2016-02-25
Fragments and bubbling while also updating - #2406 - Review

### LINKED ISSUE #2406: nasty tricky bug of the view not being updated (I think)
Hi,

here one more bug introduced in edge (works in latest). It took me *ages* to track down / simplify.

http://jsfiddle.net/ba8ny5mu/6/

Instructions:
- click many times on `show selected users`: broken, nothing happens
- click many times on `show selected groups`: works, but why? it's the same!
- click many times on `show selected users`: oh magic, now it works too!

The reason why it is tricky is because it apparently can't be simplified further.
- removing the `ractive.set('lists', lists_data)` and initializing the data directly => it'll work!
- removing the `container` around => it'll work!
- removing the whole groups part => it'll work!

...but combine all this stuff together, and suddenly, out of nowhere, you're screwed.

```xml

<container>
  Show only selected users:
  <input type="checkbox" checked="{{showSelectedUsers}}" />
  {{#each lists.users}}
  {{#if !~/showSelectedUsers || ~/lists.selected[.name] }}
  <div>
    <input type="checkbox" checked="{{~/lists.selected[.name]}}" /> {{.name}}
  </div>
  {{/if}}
  {{/each}}
</container>
<container>
  Show only selected groups: 
  <input type="checkbox" checked="{{showSelectedGroups}}" />
  {{#each lists.groups}}
  {{#if !~/showSelectedGroups || ~/lists.selected[.name]}}
  <div>
    <input type="checkbox" checked="{{~/lists.selected[.name]}}" /> {{.name}}
  </div>
  {{/if}}
  {{/each}}
</container>
```

```js
Ractive.components['container'] = Ractive.extend({
	template: "<div>{{yield}}</div>"
})

var lists_data = {
	users: [{name:'Alice'},{name:'Bob'},{name:'Charlie'}],
	groups: [{name:'Team 1'},{name:'Team 2'},{name:'Team 3'}],
	selected: {
		'Bob': true
	}
}

var ractive = new Ractive({
	el: 'body',
	template: '#template',
	data: {
		//lists: lists_data
	}
})

ractive.set('lists', lists_data)
```

evs-chris:
This is a really funky one... basically, the binding causes a bubble while inside update. Since the fragment is still dirty during the update, the bubble doesn't pass the updating fragment. Then the update completes and marks the fragment as clean, but it now has dirty children that won't get registered until something causes them to bubble again.

Making the dirty flag get reset at the beginning of the update fixes that, but it also breaks some assumptions about how things flow in the runloop. So, this takes the existing process, which allows new fragments to be registered _after_ the fragments are processed, and lets new fragments (or the same ones) be registered _during_ the processing. That messes up assumptions about how observers will fire, so the same thing is done with them.

There have been a couple of issues pop up related to fragments and observers and bindings messing with each other that had terribly complex scenarios that couldn't be reduced. I think this would probably solve them without the need to get extra-crazy like #2284, but I can't really try it because I couldn't reproduce it.

This fixes #2406, but I'm not positive it's the best way to do so. Thoughts?

If this is good, I'll try to cook up a slightly simpler test to stave off regressions in the future.

### DIFF
--- a/src/global/runloop.js
+++ b/src/global/runloop.js
@@ -77,15 +77,20 @@ function dispatch ( observer ) {
 }
 
 function flushChanges () {
-	batch.immediateObservers.forEach( dispatch );
+	let which = batch.immediateObservers;
+	batch.immediateObservers = [];
+	which.forEach( dispatch );
 
 	// Now that changes have been fully propagated, we can update the DOM
 	// and complete other tasks
 	let i = batch.fragments.length;
 	let fragment;
 
+	which = batch.fragments;
+	batch.fragments = [];
+
 	while ( i-- ) {
-		fragment = batch.fragments[i];
+		fragment = which[i];
 
 		// TODO deprecate this. It's annoying and serves no useful function
 		const ractive = fragment.ractive;
@@ -94,11 +99,12 @@ function flushChanges () {
 
 		fragment.update();
 	}
-	batch.fragments.length = 0;
 
 	batch.transitionManager.start();
 
-	batch.deferredObservers.forEach( dispatch );
+	which = batch.deferredObservers;
+	batch.deferredObservers = [];
+	which.forEach( dispatch );
 
 	const tasks = batch.tasks;
 	batch.tasks = [];
@@ -110,5 +116,5 @@ function flushChanges () {
 	// If updating the view caused some model blowback - e.g. a triple
 	// containing <option> elements caused the binding on the <select>
 	// to update - then we start over
-	if ( batch.fragments.length ) return flushChanges();
+	if ( batch.fragments.length || batch.immediateObservers.length || batch.deferredObservers.length ) return flushChanges();
 }
--- a/src/model/Computation.js
+++ b/src/model/Computation.js
@@ -75,9 +75,9 @@ export default class Computation extends Model {
 		if ( shouldCapture ) capture( this );
 
 		if ( this.dirty ) {
+			this.dirty = false;
 			this.value = this.getValue();
 			this.adapt();
-			this.dirty = false;
 		}
 
 		return this.value;
--- a/src/view/Fragment.js
+++ b/src/view/Fragment.js
@@ -305,8 +305,8 @@ export default class Fragment {
 
 	update () {
 		if ( this.dirty ) {
-			this.items.forEach( update );
 			this.dirty = false;
+			this.items.forEach( update );
 		}
 	}
 
--- a/src/view/items/Alias.js
+++ b/src/view/items/Alias.js
@@ -87,10 +87,9 @@ export default class Alias extends Item {
 	}
 
 	update () {
-		if ( !this.dirty ) return;
-
-		this.fragment.update();
-
-		this.dirty = false;
+		if ( this.dirty ) {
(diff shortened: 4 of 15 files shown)


heavyk:
from my perspective, the code is clearer with your changes. you clean the dirty flag then make the changes. 

also, seems more correct to update the observers in the runloop like that. there won't be any toe stepping. immediate observers then fragments, then deferred.

I think this is the best way to do it, especially when compared to #2284

martypdx:
Reading through the code changes, looks like intent is to allow fragments, observers and deferred observers to be queued during the runloop process while fragments, observers and deferred observers are being processed.

If that's so, LGTM,

sabob:
In angular 2 is the idea that  "change detection gets stable after a single pass" which means if there are any additional side effects after the first run, angular will throw an error. These type of restrictions is often annoying but it does help users stay within the boundaries of what is supported by the library.

My reason for bringing this up is perhaps ractive is too lenient and accommodating in this regard? 
Perhaps ractive should also impose restrictions on how and when users can update and add fragments and components?

heavyk:
I'm not sure I understand. are you talking about if one of the observers throws an exception and it exits the the runloop with the dirty flag having been set already to false and not not yet all of the fragments updated?

sabob:
My comment is more a general observation, not specific to this PR.  Sounds as if angular2 has a rigid structure or boundary in place, and if you step over it you get an error.

Ractive seems to allow one to "just keep going", which is a good thing. For example, in this PR we batch changes and perform recursive flushchanges if further changes are detected. If I understand angular2 correctly, it would have thrown an error and the user would have to change their code to accommodate the framework. I'm sure Angular2 has valid reasons in place, perhaps there are performance benefits etc.

I'm not saying Ractive is wrong in this regard, just checking how far should Ractive stretch and bend to accommodate users, because it adds complexity to the codebase.

heavyk:
the major change in this PR is before, the dirty flag was set to false only after the the updating all of the fragments. however this created a problem if the update caused something to become dirty again (during the update). so, all double dirty fragments would not get updated until something caused them to be dirty again (which is why these elements were hanging around in list even after the filter checkbox was set).

this PR changes that to set each fragment to be clean before making the update, so if the update causes it to get dirty again (through one of the children) it would be correctly result in a dirty state instead of being set to clean at the end of the update (even though one of its children made it dirty again).

I'm not sure what error could be thrown there. it's more similar to circular dependencies. I guess I'll give an analogy:

*you take your car to the car wash and they vacuum and clean the interior, but while the outside is being cleaned the commotion knocks your oversized starbucks coffee out of the beverage tray and it spills on the seat. they give you your car and tell you it is clean. you point out the coffee situation and they claim that the interior had to have been cleaned, so the coffee spill is probably how you want your car.*

lol..

evs-chris:
As nature abhors a vacuum, two way binding abhors ambiguity. When a piece of template gets rendered that has a two way binding and the binding causes a new keypath to grow on the model, it can cause more chunks of template to render. That's what was causing the issue in this case, though occasionally an observer will cause the same thing.

Issue #2408, opened 2016-02-25
Regression: `things.*.stuff.*.foo` observers don't work inside {{#with things[index] }} 


JonDum:
http://jsfiddle.net/ggb2517q/

```js
data: {
    currentThing: 0,
    things: [
        { messages: [ {message: "Apple"} ] }
    ]
},
```

Ractive doesn't like observing properties in arrays in a dynamic with :crying_cat_face: 

```html
{{#with things[currentThing] }}
```
Changing that to makes the observers work though
```html
{{#with things[0] }}
```

```js
this.observe('things.*.messages.*', _ => {
    console.log('messages changed');
})

// doesn't work in edge with `{{#with things[currentThing] }}`
this.observe('things.*.messages.*.message', _ => {
    console.log('--message changed');
})
```

JonDum:
Chris, I hope you know I'm naming my first born after you.

Tested it out and it works!

PR #2409, opened 2016-02-25
Keep primary model in sync with reference expression children - #2408 - Review

evs-chris:
This is another slightly funky one. Basically, binding to a child of a reference expression creates a branch off the reference expression proxy that doesn't notify its source when it changes. This adds a special `ReferenceExpressionChild` model that finds its source model during `applyValue` and calls `applyValue` on it too.

Thoughts? I couldn't come up with a better way™, but that certainly doesn't mean that there isn't one.

### DIFF
--- a/src/view/resolvers/ReferenceExpressionProxy.js
+++ b/src/view/resolvers/ReferenceExpressionProxy.js
@@ -5,7 +5,45 @@ import resolveReference from './resolveReference';
 import resolve from './resolve';
 import { unbind } from '../../shared/methodCallers';
 import { removeFromArray } from '../../utils/array';
+import { isEqual } from '../../utils/is';
 import { escapeKey } from '../../shared/keypaths';
+
+class ReferenceExpressionChild extends Model {
+	constructor ( parent, key ) {
+		super ( parent, key );
+	}
+
+	applyValue ( value ) {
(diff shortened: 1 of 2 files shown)


martypdx:
> Thoughts? I couldn't come up with a better way™, but that certainly doesn't mean that there isn't one.

Yes, this is right approach IMO

JonDum:
I'm now getting a `undefined` exception in `ArrayWrapper` [here](https://github.com/ractivejs/ractive/blob/dev/src/Ractive/static/adaptors/array/index.js#L57). It's something to do with `ractive-select` setting its `_items` internally from a regular non-ractified array [here](https://github.com/JonDum/ractive-select/blob/master/src/ractive-select.js#L177).

I added a simple `if(!storage) return` in `ArrayWraper.teardown()` and that made the error go away, but I'm not sure if that has any other repercussions. Thoughts?

evs-chris:
I didn't think about adaptors when calling the regular model method. The issue is the the wrapper is being torn down from both sides, which can't really be done. I think the solution is to stop calling `super.applyValue` or add a flag to the model to skip adapting. I'm a little fuzzy on adaptors though...

I'll look at it harder when I get a little time later tonight.

PR #2418, opened 2016-02-28
using removeChild(firstChild) instead of innerHTML = ""

brakmic:
According to several tests innerHTML is not only slower than removeChild but also less performant than innerText: https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent

I don't know how helpful my pull request is but the tests seem to be unambiguous.

http://jsperf.com/innerhtml-vs-removechild/15
https://jsperf.com/innerhtml-vs-removechild/96

Maybe there's a better explanation on how to properly use innerHTML/removeChild?

I hope the sources are properly formatted now. I had to revert the patch because of Tabs requirements from guidelines. Sorry.

### DIFF
--- a/src/Ractive/prototype/render.js
+++ b/src/Ractive/prototype/render.js
@@ -14,7 +14,7 @@ export default function Ractive$render ( target, anchor ) {
 
 		// make sure we are the only occupants
 		if ( !this.enhance ) {
-	      	//removeChild() is faster than innerHTML = ''
+			//removeChild() is faster than innerHTML = ''
 			//test1: http://jsperf.com/innerhtml-vs-removechild/15
 			//test2: https://jsperf.com/innerhtml-vs-removechild/96
 			while ( target.firstChild ) {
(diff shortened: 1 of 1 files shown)


fskreuz:
This looks good to me, but I'm unsure. Waiting for more eyes to take a look.

MartinKolarik:
`textContent` is not supported by IE 8. I personally couldn't care less, but since we still maintain the legacy version, this is a problem.

MartinKolarik:
Not the mention both linked benchmarks are completely broken. [Here's one that should be correct](http://jsperf.com/innerhtml-vs-removechild/376), and it's in line with my previous testing.

brakmic:
Wow, thanks! Learned something new. I must admit I can't recognize why the other tests are broken but after some time I'll uncover the secret.

Many thanks.

I think I should close this request, ok?

MartinKolarik:
>I must admit I can't recognize why the other tests are broken but after some time I'll uncover the secret.

Because `Benchmark.prototype.setup()` which is used to populate the element is called once before each test, not before every iteration, so once the children are removed, the element remains empty during the whole test.

brakmic:
Oops, now I understand it. :smile: Thanks for the clarification.
I'll now close this superfluous pull request. 

Regards,

Issue #2420, opened 2016-02-29
bug: observers observe too much and not enough


dagnelies:
Hi,

Thanks for all your effort. 

Here is an update of the fiddle last time:
http://jsfiddle.net/ey0LxzL0/2/

As you can see, there are *two* problems:
- the observers for *ALL* items are triggered, instead of just the one that changed
- the "old" value is simply wrong

One more day, one more bug, sorry about that :/

evs-chris:
I think I may be missing something here, but it looks like the observer is behaving as intended. It is being asked to watch `list.users.*` which includes 0-19 in this case. When the observer is created, it is not passed `{ init: false }` to keep it from firing on setup, so it checks all of its matching paths against the previous value (`undefined`) and fires for each instance that has changed (all of them went from undefined to an object).

dagnelies:
sure, just set a checkbox on/off, you'll see

evs-chris:
Aha! I was missing something :smile:... looking into it.

martypdx:
@evs-chris it doesn't appear that the PatternObserve call to the model for child values does any discrimination based on whether the child model is actually dirty or not.

evs-chris:
I think this is now covered.

The old values will be the same as the new because they are the same values (object identity) and ractive doesn't do any defensive cloning.

dagnelies:
Uh ...still misleading to see identical old/new values. What a pity, but I can live with that. I just remember that in the past I did extra checks that new != old because there was another bug leading to the observers triggered too often. Ideally, it'd be nice to see the "correct" old value.

evs-chris:
There's some ongoing discussion about that facet of observers in #2357.

PR #2421, opened 2016-03-02
Make sure pattern observers only fire changed paths - #2420 - RFC/Review

evs-chris:
As @martypdx pointed out in #2420, pattern observers aren't currently checking to make sure a path actually changed before firing. Instead they were just going off an equality test, which is fine for primitives and strict equality, but less so for objects that are never considered equal.

I initially tried to use the viewmodel changes registry to test the keypaths coming into handleChange, but that wasn't playing out as well as I had hoped. Setting an object at a keypath where one of the keys in the set object would match a pattern is trickier to deal with. I kinda think it would cause issues with components too. It still might be the better approach, but I went a different direction with this...

Instead of looking up and getting new values for every matching key when the base model notifies of a change somewhere, this registers an object with the matching target models that handles the new value op of the observer. The base model is still checked for new keys. I'm not positive this covers all of the weird little corner cases that observers have to deal with.

I was thinking that maybe a capture/bubble cycle for `set`s of object values may be a good way to handle pattern observers and the change event. That would also allow deeply nested observers e.g. `foo.bar.**.stuff.*` which would probably help out people trying to watch a keypath for change tracking purposes.

Any thoughts?

### DIFF
--- a/src/Ractive/prototype/observe.js
+++ b/src/Ractive/prototype/observe.js
@@ -243,7 +243,6 @@ class PatternObserver {
 			runloop.addObserver( this, this.defer );
 
 			if ( this.once ) this.cancel();
-
 		}
 	}
 }
--- a/src/Ractive/prototype/observe.js
+++ b/src/Ractive/prototype/observe.js
@@ -155,18 +155,30 @@ class PatternObserver {
 			this.newValues[ model.getKeypath( this.ractive ) ] = model.get();
 		});
 
+		baseModel.register( this );
+
+		this.models = models;
+		this.touchedKeys = {};
+		let observer = this;
+		this.registrar = {
+			handleChange( model ) {
+				observer.newValues[ model.getKeypath( observer.ractive ) ] = model.get();
+			}
+		};
+		let i = models.length;
+		while ( i-- ) models[i].register( this.registrar );
+
 		if ( options.init !== false ) {
 			this.dispatch();
 		} else {
 			this.oldValues = this.newValues;
+			this.newValues = {};
 		}
-
-		baseModel.register( this );
-
 	}
 
 	cancel () {
 		this.baseModel.unregister( this );
+		this.models.forEach( m => m.unregister( this.registrar ) );
 	}
 
 	dispatch () {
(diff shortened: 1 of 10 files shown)


JonDum:
Rebased comp-bind on top of this. Couldn't really test because I got an infinite loop with `oninit` of a component getting called again and again. :|

Good news though was that I was able to reproduce it! http://plnkr.co/edit/zHGOcz6kDLplWCc6DP0D

Careful with the preview. It'll infinite loop as soon as you open. Suggest adding a `debugger` in the ractive source or trying to pause execution with devtools already open.

evs-chris:
@JonDum thanks for taking it for a spin!

This one got really weird on me, and I was in the midst of trying to come up with a better solution based on the end of the opening comment, when the ole day job dumped a ton of stuff on me. I managed to get tests passing on the alternate fix, but there is probably still some weirdness about it.

The new approach makes all models that notice a change during a runloop turn register with the runloop. Then when the observer portion rolls around, the observer pulls the change list and looks for matching models. The change event does more or less the same and is no longer based on having a fragment update and no longer maps changes at the root model level.

Attribute locking was causing an observer-as-validation test to fail, so I removed the locks. It looks like the whole purpose of locking was to avoid updating a value whilst typing for validation scenarios, and I think laziness is probably a better approach. I haven't fully explored what sort of breakage this could cause, but it seems like just checking the value coming in to the attribute delegate to see if it has actually changed is a reasonable approach.

TL;DR not sure if it will work yet, but I think there's a more flexible way to fix this that opens up some other possibilities for observation later on

evs-chris:
Just wanted to note that I'm still playing with this one. Some questions regarding change events have arisen from the existing tests. Basically, changes here and to the runloop to address other issues can cause multiple change events during a single turn. I think that's probably inevitable, but I haven't had time to trace it to conclusion yet.

Also, I think we should go ahead and do the breaking change for observers observing removal. It makes sense for people trying to track changes for rollback purposes.

evs-chris:
Closing in favor of #2502, which is considerably less complicated, and, more importantly, doesn't fail tests when merged with dev :smile:

PR #2425, opened 2016-03-06
Keep update from running in nested runloops for repeats and sections - #2424

### LINKED ISSUE #2424: if section loops if data is not found in select
Hi,

here is another impressive bug:

http://jsfiddle.net/xc2o8y08/24/

It's probably related to #2423 but I'm not sure. Something is seriously broken there.

...anyway, I wish you a good Weekend.

evs-chris:
This fixes #2424 and maybe a few others... __not ready - no tests__

In #2407, the dirty flag got bumped around such that happenings during an update that should cause a bubble actually caused a bubble. That was also keeping an update from running during an update, which apparently causes all sorts of bizarre behavior. This adds a `updating` flag to RepeatedFragment and Section to keep them from getting out of whack whilst updating content with bindings that cause further updates. I have to ponder this a bit more, but should all items have an update guard, or just fragments, or just what I have here?

### DIFF
--- a/src/view/Fragment.js
+++ b/src/view/Fragment.js
@@ -305,9 +305,11 @@ export default class Fragment {
 	}
 
 	update () {
-		if ( this.dirty ) {
+		if ( this.dirty && !this.updating ) {
 			this.dirty = false;
+			this.updating = true;
 			this.items.forEach( update );
+			this.updating = false;
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


martypdx:
If just adding it to `RepeatedFragment` solves immediate issues, I would be conservative about extend it to all items.

evs-chris:
Between Fragment and RepeatedFragment, it seems to be covered. Applying this only to RepeatedFragment didn't cover all of the issues, but if something crops up in the future that makes this unnecessary, it's now only six lines to adjust :smile:.

martypdx:
:+1:

Issue #2422, opened 2016-03-03
Bug: @index incorrectly 


dagnelies:
Hi,

in the process of chasing a bug (not yet able to isolate), I nevertheless managed to isolate another one:

http://jsfiddle.net/txptg3bz/4/

The line: `<span style="margin-left:{{50*(@index)}}px;">{{.}}</span>` "breaks" and appear to render based on a wrong index after a splice. This works with latest, but breaks on edge.

evs-chris:
I feel like this is probably my fault and the result of changes made to speed up splicing for big iterative sections. I'm a bit swamped right now, but I'll get round to it shortly. Sorry.

dagnelies:
No biggy, take your time. We all have just two hands and one head.

evs-chris:
I've tracked this back to an issue with rebinding attributes that depend on an expression with an index ref, so it wasn't me after all :smile:. What's happening is the attribute isn't a mustache, so it doesn't rebind directly. Instead it delegates to its fragment. Since it doesn't have a model to check for a change, it doesn't know that it changed. It seems like the best way forward is to check the attribute value on rebind to see if an update is in order, but I'm not sure. An alternate would be to have a proper rebind for attributes, which would get them to update with their fragment if something actually changes, though I suppose that's not really too different than just doing a value check.

dagnelies:
Well, you're the expert here. ;) I sometimes try to look into the sources but I'm quickly lost there. I'd just pick the safest, simplest, straightforward solution.

dagnelies:
thanks! :+1:

PR #2438, opened 2016-03-14
Check attribute values while shuffling to see if an update is in order - #2422 - RFC

evs-chris:
The test case from #2422 involves rebinding an attribute value that depends on an `@index` ref, but since attributes aren't sensitive to the change (they just wholesale unbind and bind), they never get notified of a change during the shuffle. This adds a step that grabs the value at the start of rebinding and checks it against the value after rebinding. If the value is not strict equal, it bubbles.

An alternate fix would be simply: `rebind() { if ( this.fragment ) this.fragment.rebind(); }`. It seems to work, but for some reason it didn't feel right (I think stepping through too much code may be throwing me). I'm not wild about calling `getValue` twice on every attribute shuffle, so if there's no reason not to do that, it would probably be better.

Thoughts?

### DIFF
--- a/src/view/items/element/Attribute.js
+++ b/src/view/items/element/Attribute.js
@@ -78,15 +78,7 @@ export default class Attribute extends Item {
 	}
 
 	rebind () {
-		const current = this.getValue();
-
-		this.unbind();
-		this.bind();
-
-		// the value may have actually changed if it depends on a template context (@index, etc)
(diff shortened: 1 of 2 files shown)


evs-chris:
After brief consideration, I think delegating to fragment rebind is the way to go. I'm gonna lean on the test suite here and go with if it's green, it's good.

Issue #2396, opened 2016-02-23
[0.8] - getNodeInfo is not returning the keypath


kouts:
I'm probably doing something wrong but here goes:

```xml
<div id="container"></div>

<script id="my_template" type="text/ractive">
  <div>
    <label>Input element</label>
    <input type="text" value="{{text_value}}" decorator="my_decorator" />
  </div>
</script>
```

```js
Ractive.decorators.my_decorator = function(node, props) {
  console.log(node);
  var node_info = Ractive.getNodeInfo(node);
  console.log(node_info);
  return {
    teardown: function() {}
  };
};


var ractive = new Ractive({
  el: '#container',
  template: '#my_template',
  data: {
    text_value: 'Sample text'
  }
});

```

In this fiddle http://jsfiddle.net/wwx3gj1e/2/ I'm trying to get the keypath from the DOM node (an input text element), but I get an empty object (check console).
Am I using getNodeInfo in the wrong way or is this a bug?

evs-chris:
No, you're using it correctly. It's just that there's no context at that point in the template. Context is only provided by non-`if` sections i.e. plain section mustaches, `each ` sections, and `with ` sections. Your context there is effectively the root model.

kouts:
Thanks @evs-chris , so is there any way to get the keypath associated with the DOM element.
In this fiddle's case it would be: text_value

evs-chris:
Ah, so you want the binding keypath. Unfortunately there's no official API for that, but as of #2376, I've been thinking about proposing a utility method to resolve keypaths relative to an element. I updated your fiddle with an incomplete version that supports getting a binding keypath as well as any element's keypath. Will that work for you?

It's missing the path resolution part because I'd have to copy private chunks of Ractive to get it to work at this point. Do you think this would be a reasonable addition to Ractive to sit next to `Ractive.getNodeInfo`?

martypdx:
> I tried wrapping it into a "values" section but I get "values" as the keypath.

Wrap the input in a `with` block and change the ref to `this` or `.`:

```html
{{#with text_value}}<input type="text" value="{{this}}" decorator="my_decorator"/>{{/with}}
```

evs-chris:
... or you could do what that, though it is a bit noisy. I'm determined that _someone_ else needs to resolve keypaths relative to an element :laughing:

Completely forgot to link to the updated fiddle: http://jsfiddle.net/wwx3gj1e/4/

martypdx:
[...]

PR #2442, opened 2016-03-17
Extended context helpers - RFC

evs-chris:
~~__Not ready for merge... no tests yet :smile:__~~

This is my attempt at making interaction with Ractive from the outside world a bit more comfortable and convenient, as previously discussed in such issues as #2402, #2376, and #2396. Here's what it does:

* Extends `@keypath` and `@rootpath` special refs to allow them to receive a reference and resolve the keypath relative to that. It uses a method-call-ish syntax and accepts a plain reference in the parens: `@keypath(../../foo)` in the context `some.nested.listy.271.deep.items.41.path.to.who.knows` would return the string `some.nested.listy.271.deep.items.41.path.to.foo`.
* Adds a handful of context helpers that use `resolveReference` to allow template-relative paths to be used with proxies for `get`, `set`, `add`, `toggle`, etc. The helpers get defined onto events and the result of `getNodeInfo` calls. This lets you use events and node info as a pointer into the data at the node's context in the template. This exposes index, key, special, and aliased refs (anything accessible from the template at that point) through `get` e.g. `info.get('.prop')` or `info.get('@index')` or `info.get('myHandyAlias')`.
  * `resolve` takes a relative keypath and returns the actual keypath to the model.
  * Binding helpers `isBound`, `getBindingPath`, `getBinding`, and `setBinding` are added for working with two-way bindings on form controls.
* Adds a property on the promise returned from splice methods with the result of the splice. I eventually would love to add await support for references, but until then, this makes triggering splices from events a bit friendlier.
* Adds getter support to the `defineProperty` shim, because *sigh* IE8.
* Deprecates the existing `context`, `keypath`, `rootpath`, `index`, and `key` properties on events and node info.
* Allows `../` and similar prefix-only refs because they're handy for dealing with lists.
* Checks the third arg to splice for numericality, and if it doesn't pass, sets it to the remaining length past the index.
* Removes the fragment ref from node `_ractive` storage, because it's redundant - everything is more easily reachable from the element proxy.
* Extends `Ractive.getNodeInfo` to take a string, which is passed to `document.querySelector` before continuing with the result as usual.
* Adds an instance version of `getNodeInfo` that takes either a node or a selector that is passed on to `find`.

~~I'll add some tests shortly, and I'm also thinking about letting `Ractive.getNodeInfo` take a selector string that looks up a matching node and returns the info.~~ I haven't implemented some of the map and wildcard forms of `set` and friends, but it should be possible without too much trouble if it's not YAGNI.

If you've got a few minutes to spare, I'd love some comments! Specifically:

1. I wanted to use `keypath` instead of `resolve`, but then I couldn't deprecate the existing property. Is there a better name than `resolve`? I also considered `getKeypath`, but that felt weird passing it a keypath.
2. Does the binding helper stuff make sense? I know it's useful for validating decorators and the like, but does it need to be adjusted?


fskreuz:
> I wanted to use `keypath` instead of `resolve`

In favor of `resolve`. It acts similar to Node's [`path.resolve`](https://nodejs.org/api/path.html#path_path_resolve_from_to).

> Deprecates the existing `context`, `keypath`, `rootpath`, `index`, and `key` properties on events and node info.

Probably might want to highlight in that same bullet point what the alternatives are when they're gone. I'm also kinda lost.

evs-chris:
Ah yes, for the deprecations, the replacement is in the warning message for debug, but they don't exactly stand out here :smile:

* `.keypath` becomes `.resolve()`
* `.rootpath` varies slightly but going with the event version: `event.resolve(event.ractive.root)`, which is a bit longer, but also more than a little corner-casey
* `.context` becomes `.get()`
* `.index.i` and `.key.k` (or whatever your various aliases are in the template) just become `.get('i')` or `.get('k')`, etc. The added bonus is that `.get('@index')`, `.get('@key')`, and `.get('someAlias.path.etc')` also work, so you can reference aliases from `{{#with foo.bar.baz[bat] as bop, 42 * yep as bip}}` (`.get('bip')` and `.get('bop')`) and you don't have add an alias if you are only interested in the nearest index.

> I'm also kinda lost.

I'm rarely not lost :smile:. Were you referring to the deprecation, the code, or generally what this replaces and/or is useful for?

evs-chris:
I went ahead and added support for `Ractive.getNodeInfo('query selector here')` in addition to the existing `Ractive.getNodeInfo(someNodeRef)`. It goes well with Chrome's 'Copy Selector' context menu. As a quick example, you could pop up the console and type in `Ractive.getNodeInfo('input').setBinding('Hello')` to set whatever keypath the first input field is bound to to 'Hello', assuming the first input on the page is managed by a Ractive instance.

evs-chris:
The tests are in. I didn't do a set specifically for events because they use all of the same code that the node info tests use, and testing event is more painful than having just a simple synchronous method call. To clarify though, anything you can do with the result of `Ractive.getNodeInfo(...)` you can also do with the provided event object in a Ractive event e.g. in `<a on-click="doSomething(event, value)">` `doSomething` will be passed an event object that has all of the `get`, `set`, `toggle`, etc methods that are set up for references relative to that point in the template, including aliases, named indexes and keys, and special refs. Better, if the event expressions in #2386 land, manipulating lists from the template becomes super simple - think removing items or moving them up or down the list.

```html
{{#with some.really.big.complex.context}}
  <ul>
  {{#each items}}
    <li>
      {{.name}}
      <button {{#if @index === 0}}disabled{{/if}} on-click="event.splice('../', @index - 1, 0, event.splice('../', @index, 1).result[0])">Move Up</button>
      <button {{#if @index - 1 === ../length}}disabled{{/if}} on-click="event.splice('../', @index + 1, 0, event.splice('../', @index, 1).result[0])">Move Down</button>
      <button on-click="event.splice('../', @index, 1)">Remove</button>
    </li>
  {{/each}}
  </ul>
{{/with}}
```

evs-chris:
Figured I should mention that that last example is possible without the extra methods just using the keypath expressions added in this PR:

```html
{{#with some.really.big.complex.context}}
  <ul>
  {{#each items}}
    <li>
      {{.name}}
      <button {{#if @index === 0}}disabled{{/if}} on-click="splice(@keypath(../), @index - 1, 0, @ractive.splice(@keypath(../), @index, 1).result[0])">Move Up</button>
      <button {{#if @index - 1 === ../length}}disabled{{/if}} on-click="splice(@keypath(../), @index + 1, 0, @ractive.splice(@keypath(../), @index, 1).result[0])">Move Down</button>
      <button on-click="splice(@keypath(../), @index, 1)">Remove</button>
    </li>
  {{/each}}
  </ul>
{{/with}}
```

fskreuz:
Nice! So if I understand correctly, 

- The object resulting from a `getNodeInfo` is a way to manipulate data relative to the anchored element.

- The event object will still be there for events, but will now be decorated with these helpers.

Sounds good to me. And this is probably the single most tested feature in history now. :smile:

evs-chris:
I was looking at this from a decorator perspective and realized that there's not a perfectly straightforward way to get node info from an instance. There are a fair number of scenarios, including those involving decorators, where you wouldn't necessarily want to rely on the global or funny things like `myInstance.constructor.getNodeInfo(...)`.

To address that, I've added an instance flavor of `getNodeInfo` that will accept an HTMLElement in the same way as the static version, but it will also accept a string selector that it will hand off to `find` instead of `document.querySelector`. That lets you easily find a node that's relevant to the instance that you're interested in.

```js
const r = new Ractive({
  template: '{{#with foo.bar}}<div class="interesting">{{.thing}}</div>{{/with}}',
  data: { foo: { bar: { thing: 'yep' } } }
});

const info = r.getNodeInfo('.interesting');
console.log(info.resolve()); // foo.bar
console.log(info.resolve('.thing')); // foo.bar.thing
console.log(info.get('.thing')); // yep
info.set('.thing', 'hey!');
console.log(info.get('.thing')); // hey!
```

evs-chris:
Should this be punted to after 0.8, never, or is it useful and non-invasive enough to go ahead? I've used it a bit and can say that for dev-mode especially, I find it super useful.

kouts:
Does this include the getKeypath method (or something similar)? I have used this and it is a life saver!

evs-chris:
It does, plus a few more. `getBindingPath` returns the keypath for the two way binding, `isBound` returns true if there is a two way binding, `getBinding` returns the bound value, and `setBinding` sets the bound value. The last two are equivalent to calling `get` and `set` with the binding keypath.

kouts:
+1 from me to merge this then!

evs-chris:
This is now back in sync with dev post event expressions.

evs-chris:
I was going to hang back on this to avoid extra fluff in 0.8, but since event expressions landed, I'm going to go ahead and merge this. It's fairly non-invasive, so it won't be hard to revert if necessary.

kouts:
I'm on 0.8, how do I use getBindingPath inside a decorator?
Right now I'm using getKeypath custom method:
```
Ractive.decorators.my_decorator = function(element){
    var keypath = Ractive.getKeypath(element, {binding: true});
    var $el = $(element);
    $el.autoNumeric('init',{
        aSep: ' ',
        aDec: ',',
        callbacks: {
            afterConvert: function(value){
                self.updateModel(keypath);
            }
        }
    });
    return {
        teardown: function(){
            $el.autoNumeric('destroy');
        }
    };
}
```

evs-chris:
Here's the binding portion of the API:
```js
var info = Ractive.getNodeInfo(element);
if (info.isBound()) { // this is an element with a binding
  console.log(info.getBindingPath()); // logs the keypath for the binding on this element
  console.log(info.getBinding()); // logs the value of the binding
  console.log(info.setBinding('foo')); // sets the binding to 'foo'
}
```

It looks like another approach you could take here would be to pass in a relative path to the decorator e.g. `<input decorator="numeric:{{'.someNum'}}" />` (which will hopefully get even cleaner as `<input as-numeric=" '.someNum' " />` soon), and then in the decorator:
```
Ractive.decorators.numeric = function(el, path) {
  ...
  afterConvert: function(val) { info.set(path, val); }
  ...
};
```

That would be a one-way (write-only) input, though.

kouts:
Thanks @evs-chris that's a great set of helpers!

PR #2449, opened 2016-03-18
Gh 2446

### LINKED ISSUE #2446: Transition processParams returns empty object when invoked without a default object
According to docs, processParams accepts an optional defaults:
http://docs.ractivejs.org/0.3.9/transitions

Given the template:
```
template: '<p intro="foo:{duration: 1000}"></p>',
```
this test should work:
```
transitions: {
    foo ( transition, params ) {
        t.equal( params.duration, 1000 );

        params = transition.processParams( params );
        t.equal( params.duration, 1000 );
```           

it fails at the second assertion because processParams calls Object.extend which returns an empty object if the a defaults isn't provided. In the test above, params is empty after calling transition.processParams.

Should be easy fix, I'll do a PR tomorrow.

sabob:
This PR fixes #2446 by changing object.extend while loop into a forEach loop which include all arguments.

transition.processParams calls objec.extends as follows:
```
return extend( {}, defaults, params );
```

The issue is in the object.extend loop:

```
export function extend ( target, ...sources ) {
    while ( source = sources.shift() ) {
```

If processParams calls extend with an undefined defaults, the while loop will stop at the second iteration when sources.shift returns the "undefined" default.

By using a forEach loop we iterate all arguments, even if they are undefined.


heavyk:
if you make it a `for(var i = 0; i < sources.length, i++) { ... }` type loop it'll be faster as well.

sabob:
Dunno, the whole ractive codebase uses the foreach notation. Also I'd imagine browser compilers would optimize away the differences anyway.

evs-chris:
:+1:

It doesn't really matter how you iterate unless you're in a hot path.

Issue #2404, opened 2016-02-24
Dynamically updated element class blows away classes added by a decorator


Ghazgkull:
If you have an element in a template with classes that are conditionally added based on bound data, Ractive will blow away any classes on that element which were added by decorators if/when the data changes.

Here's a working example that demonstrates the issue:
http://codepen.io/Ghazgkull/pen/EKYVxd

martypdx:
Looks like a variation of #1262. Basically the text fragment that corresponds to the dynamic attribute value doesn't know anything about value modifications happening elsewhere.

Probably mostly an issue for `class` and `style`. The solution may require adding style and class value parsing, or maintaining a before/after comparison.

evs-chris:
Classes and styles are no longer trampled in edge.

Ghazgkull:
Woo!

Issue #1262, opened 2014-09-25
style attributes with mustache references overwrite transition values


martypdx:
```html
<div intro='fade' style='width: {{width}}px;'></div>
```
```js
var r = new Ractive({
    el: document.body,
    template: '#template'
})
r.set('width', 200)
```

Rich-Harris:
There's probably limits to what we can do here. Maybe one or both of these could help:

## 1. parse style attributes

Rather than `node.setAttribute('style', newStyle)`, we could do this under the hood:

```js
styles = parseStyles('width: 200px;'); // [{ prop: 'width', value: '200px' }]

styles.forEach( style => {
  node.style[ camelCase( style.prop ) ] = style.value;
});

// plus some logic to handle styles that previously existed but don't any more...
```

This would make it possible to manually manipulate styles without Ractive breaking them, which might in any case be good for interop with libraries that do that kind of DOM manipulation. At the moment Ractive is liable to stomp their changes. Not sure what the perf implications would be with this approach.

Also, it's not immediately obvious how straightforward it would be to parse styles. Definitely not quite as straightforward as splitting on `;`...

## 2. warn if style attribute is affected by data changes during transitions

When a transition starts, a flag is set. If data changes, causing the style attribute to be updated, while the flag is still active, we print a warning.

martypdx:
Agree that looking into treating style attribute specially will good for __future__ milestone. 

It may actually have some perf benefit where style is being used to animate if we can directly target the individual style. There may need to be rules like no blocks across multiple styles like `style='{{#foo}}width: 200px; height;100px;{{else}}max-height: 50px;{{/}}'`. 

It could then be a parsetime additional thing that, if conditions are met, gets applied at runtime - otherwise falls back to full attirbute set.

martypdx:
Just ran across this again in a subtle hard to find way :) 

Had a new idea, what if we allowed style properties as attributes of elements:

```html
<div intro='fade' style.width='{{width}}px;' style.height='{{height}}px'></div>
```

evs-chris:
Classes and styles are no longer trampled in edge.

PR #2453, opened 2016-03-18
Special treatment for style and class attributes - RFC

evs-chris:
~~__Not ready yet... missing tests :smile:__~~

Transitions and decorators often need to twiddle classes or inline styles, but if the element also has class or style attributes, they are just blindly overwritten - #1262 #2404. This is my attempt to resolve those issues.

In both cases, the list of names/attributes set in the last run is kept around to see if anything that was set last time is no longer present and should be removed. There are a couple of tests failing now because the order of the constituents comes out a little different than just overwriting the attribute.

For the style side, the parser is just a char by char looped brute force reader. It has stood up to everything I've found to throw at it so far, though @MartinKolarik, as the wizard of textual state machines, could probably replace it all with a regex :smile:. Once the key/value pairs are pulled out of the string, they are iterated over and if the node's style object has a matching key, it is set to the value.

Thoughts?


heavyk:
there may be a faster way to lookup if the class exists, but this is the right way to go.

maybe if classList were used (polyfilled if unavailable), all decorators and transitions could use that and not step on each other's toes.

https://developer.mozilla.org/en-US/docs/Web/API/Element/classList

MartinKolarik:
>though @MartinKolarik, as the wizard of textual state machines, could probably replace it all with a regex :smile:.

Couldn't resist :smile: 
```js
// import camelCase from 'utils/camelCase';

const specials = { 'float': 'cssFloat' };
const remove = /\/\*(?:[\s\S]*?)\*\//g;
const escape = /url\(\s*(['"])(?:\\[\s\S]|(?!\1).)*\1\s*\)|url\((?:\\[\s\S]|[^)])*\)|(['"])(?:\\[\s\S]|(?!\1).)*\2/gi;
const value = /\0(\d+)/g;

function readStyle ( css ) {
	let rules = {};
	let values = [];

	css.replace( escape, match => `\0${values.push( match ) - 1}`)
		.replace( remove, '' )
		.split( ';' )
		.filter( rule => !!rule.trim() )
		.map( rule => rule.replace( value, ( match, n ) => values[ n ] ) )
		.forEach(rule => {
			let i = rule.indexOf(':');
			let name = camelCase( rule.substr( 0, i ).trim() );
			rules[ specials[ name ] || name ] = rule.substr( i + 1 ).trim();
		});


	return rules;
}
```

It should handle any valid CSS and is a little faster, so feel free to use it. Also, here's a valid CSS  that's failing with your parser: `background: url(/*xx);` (because URLs don't have to be inside quotes).

fskreuz:
> ```js
.forEach(rule => {
  let i = rule.indexOf(':');
  let name = camelCase( rule.substr( 0, i ).trim() );
  rules[ specials[ name ] || name ] = rule.substr( i + 1 ).trim();
});
```

One less outer variable :smile: 

```js
.reduce((rules, rule) => {
  let i = rule.indexOf(':');
  let name = camelCase( rule.substr( 0, i ).trim() );
  rules[ specials[ name ] || name ] = rule.substr( i + 1 ).trim();
  return rules;
}, {});
```

MartinKolarik:
@fskreuz :+1: updated my comment.

heavyk:
yeah but your version is quite a bit more difficult to read. I prefer chris' implementation ... it just needs to be updated for `url()`

evs-chris:
@MartinKolarik :+1:

I switched the parser because even though it's a bit harder to follow, it's considerably shorter and there's apparently a speed advantage. Plus if I cross my eyes just right it's easy to read :laughing:.

There are also a few tests and everything passes now that I've worked around weirdness in how phantom handles inline styles set from the DOM side.

evs-chris:
In an odd turn of events, appveyor is succeeding where travis is not. It looks combination of various node versions and phantom cause inline style attributes to be updated in different ways? At this point, I suppose the easy way out is to just remove all white space for the comparison...

evs-chris:
Any objections?

martypdx:
@evs-chris my only concern would be if the perf impact for frequently updated values. Consider:

```html
<div style="transform: translate3d({{x}}px, {{y}}px, 1);">...</div>
```

Where `x` and `y` are set via a mousemove event. I know the number of styles/classes would be relatively small, but it is a `O(n^2)` process given use of `indexOf`.

Not directly related to this PR, I've wondered whether we should introduce attribute-based class and style syntax:

```html
<div class-foo="conditional" style-transform="translate3d({{x}}px, {{y}}px, 1);">...</div>
```

as an augmentation for this PR to more succinctly write and apply individual styles or classes.

evs-chris:
> my only concern would be if the perf impact for frequently updated values

Fair enough :smile:

> I've wondered whether we should introduce attribute-based class and style syntax

I like it! - and it looks like it probably offers a way to address to performance concern, so done!

martypdx:
> so done!

@evs-chris you rock :grin:

PR #2455, opened 2016-03-20
Build updates and try to get older IE back

evs-chris:
*tears of joy, rage, and sadness*

I made the mistake of trying something using edge out on IE8. After a good bit of cursing, wailing, gnashing of teeth, and desk-to-face interaction, I've arrived at a build file that, while terribly slow, will get to the point that tests run and even mostly pass in IE8 (~500 failures, but hey, something runs). I've not yet had the strength to look at other versions.

I hope to address most of the open IE issues in this PR - and get the build back down well below 45 seconds, but for now I have to go take a shower and not look at Windows again for a day or three.


heavyk:
holy shit nice work. the babel/rollup update is good for something too. :+1:

evs-chris:
Well, by rolling all of the various test files into one during the build, the build time drops back down to around 30 seconds, which is still a bit excessive. The test page does load a bit faster though.

Once again, Travis has decided to take his ball and go home, but I promise that an `npm run build` completes successfully :smile:

evs-chris:
I never did come up with a way to get build times back down from orbit. The Babel 6 docs are a bit sparse (or I'm just not good at finding them), but I think if there was a way to run it without adding all of the helpers to every file then it could be run in a cache-friendly way to at least cut the rebuild time down to something more reasonable. I tried a handful of things, but they either didn't produce viable builds or took just as much time to complete :cry:.

MartinKolarik:
I wanted to merge this to prevent more merge conflicts, but I'm definitely going to take a look if we can't make it faster.

Issue #2461, opened 2016-03-22
Computations survive the destruction of the components that own them


Rich-Harris:
Normally this isn't observable, but in certain cases involving reference expressions the computation continues to update after the component that owns it has been destroyed: http://jsfiddle.net/6xnLt2jy/1/. (The fact that reference expressions involving computations – e.g. `{{foo[bar]}}` where `foo` is computed – cause the computation to be re-evaluated as their containing sections are being destroyed is probably the more important bug, since a dormant computation is presumably harmless, though to be honest having been out of this codebase a while I'm not certain)

evs-chris:
Would overloading `mark` on `ExpressionProxy` and `ComputationChild` to not actually force an update when `mark`ing (just set a dirty flag) work? Just mark the deps and children and let the actual change check happen as the deps ask for the value... that would effectively defer the `retrieve` that triggers the issue until after the section is removed, which would avoid the pull on the model from the interpolator. I think. :confounded:

Rich-Harris:
I had a similar thought, and tried adding a `mark` method to `ReferenceExpressionProxy` (which is the offender here, rather than `ExpressionProxy`, though perhaps both behave similarly), and immediately got *more* test failures, so shelved that line of enquiry for the time being. But it seems likely that the proper solution lies in that direction. Going to have a prod and see what's behind the new failures.

Curiously, `ExpressionProxy.prototype.mark` looks like this...

```js
mark () {
  this.handleChange();
}
```

...whereas `ReferenceExpressionProxy.prototype.handleChange` looks like this:

```js
handleChange () {
  this.mark();
}
```

Pretty sure I wrote both. Can anyone explain to me what I was *trying* to write? Nothing more humbling than being completely stumped by your own code

evs-chris:
I didn't have much time to poke, but, while almost certainly incomplete, this passes the tests:

```diff
diff --git a/src/view/resolvers/ReferenceExpressionProxy.js b/src/view/resolvers/ReferenceExpressionProxy.js
index 6523251..744875a 100644
--- a/src/view/resolvers/ReferenceExpressionProxy.js
+++ b/src/view/resolvers/ReferenceExpressionProxy.js
@@ -3,7 +3,7 @@ import { REFERENCE } from '../../config/types';
 import ExpressionProxy from './ExpressionProxy';
 import resolveReference from './resolveReference';
 import resolve from './resolve';
-import { unbind } from '../../shared/methodCallers';
[...]

PR #2462, opened 2016-03-22
Partial fix for #2461

Rich-Harris:
Couple of failing tests for #2461, and a partial fix – it ensures that a root model being torn down takes its computations with it, but doesn't solve the slightly thorny problem that a reference expression based on a computation causes the computation to recompute as its containing section is being destroyed.


evs-chris:
Fixed :smile:

PR #2471, opened 2016-03-28
Fix #1856

### LINKED ISSUE #1856: Change {{#with obj}} to not render if obj is falsey

MartinKolarik:
Doesn't change behavior of `{{>partial context}}`, as I think partial should render even if `context` is undefined. Thoughts?

### DIFF
--- a/src/parse/converters/mustache/handlebarsBlockCodes.js
+++ b/src/parse/converters/mustache/handlebarsBlockCodes.js
@@ -1,9 +1,8 @@
-import { SECTION_EACH, SECTION_IF, SECTION_IF_WITH, SECTION_WITH, SECTION_UNLESS } from '../../../config/types';
+import { SECTION_EACH, SECTION_IF, SECTION_IF_WITH, SECTION_UNLESS } from '../../../config/types';
 
 export default {
 	'each':    SECTION_EACH,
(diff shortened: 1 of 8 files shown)


evs-chris:
:+1: I agree.

Also, I feel like a partial with context should not have the context portion applied around the partial so that it doesn't interfere with resolution inside the partial expression. So instead of desugaring `{{>partial context}}` to `{{#with context}}{{>partial}}{{/with}}`, move the `{{#with}}` into the fragment of the partial when it's set up in the virtual dom so it would end up being something like `{{>partial}}{{#with context}}...partial content with 'partial' resolved in parent context...{{/with}}{{/>partial}}`. It doesn't interfere _too_ often, but when it does, it's really confusing.

I only mentioned that because I misread the opening comment and thought that's what you were saying :smile:.

MartinKolarik:
>Also, I feel like a partial with context should not have the context portion applied around the partial so that it doesn't interfere with resolution inside the partial expression.

That makes sense. I'm going to merge this, but we should probably address it in a separate PR.

Issue #2469, opened 2016-03-27
Bug: Can't unset keys in objects passed from parent


ADantes:
I need to unset a key from an object.  Unfortunately for me, Ractive's update only notices the change *if* it's at the root level.

This works just fine: http://jsfiddle.net/q493yc7b/2/
Moving all the relevant logic into a component breaks it: http://jsfiddle.net/q493yc7b/1/

evs-chris:
This is sort-of known issue, though I can't seem to find the relevant number at the moment. On edge, you can work around it by updating without the key e.g. `update('color')` instead of including the full path. I'm looking into fixing it fully on edge, but unfortunately I don't think there's going to be a 0.7.4 to fix latest. If it's something you don't mind hacking around until 0.8 emerges, you can do roughly the same thing as edge by using `this.parent.update(this.viewmodel.mappings['color'].keypath.str)` to force the parent to update as well.

Issue #2141, opened 2015-09-09
using update('some.key') changes not reflected (object updated outside of set())


woodlandhunter:
See the attached fiddle. On Edge, click on "Click me" and the change is not reflected. On latest, the change is reflected.

http://jsfiddle.net/y270cxxb/

(I know usually you would use `set()` for this, which works, but in my application I'm updating many objects in a loop and then doing an update() call once.)

I can work around it by using `update('some')` instead of `update('some.key')` but since the behavior is changed I thought it was worth checking into whether this was now expected.

evs-chris:
It looks like the issue on edge is that `sub` is an ambiguous reference. Changing the reference to `.sub` to remove the ambiguity causes the "Click me" to work. I think the unresolved process should handle this case from `update`.

PR #2474, opened 2016-03-30
Notify parent deps on update too - #2469

evs-chris:
This is a weird case wherein `update`ing a child keypath doesn't flag a section iterating the parent as dirty across a component boundary and perhaps also not across the boundary in certain circumstances. `applyValue` dirties upstream deps, so it seems reasonable that `update` should also mark upstream deps, does it not?

Linking #2469


evs-chris:
Finally found the related issue: #2141 and its fix #2448

PR #2478, opened 2016-03-31
Reduce memory use with this one simple trick! - #2477

### LINKED ISSUE #2477: memory issues with ractive edge
Hello,

I was testing component construction with ractive edge, and I noticed that memory utilization goes up as I load/reload the same data. Here is the fiddle: http://jsfiddle.net/ax61fysj/. When I alternate between the "load data" and "load blank" buttons, I see the JS heap increase from Chrome dev tools.

evs-chris:
It looks like this fixes #2477, and may fix part of #2461. The timeline view is a little wonky with what it shows based on when a gc happens, but I've done heap profiling and that is saying that this no longer has a gazillion detached nodes still hanging around.

1. Component mappings are not un-mapped when the component is torn down, which lets the component and all its related gubbins stick around in memory. This adds a teardown, which overlaps with the one in #2462, that handles unregistering mappings so that the components don't hang around. I haven't looked to hard, but it looks like maybe breaking the mapping link will also let the computations get gc'ed when the component is torn down too.

2. ExpressionProxies get a reference to their fragment, create their own Computation (always!), and don't ever unregister. This adds a check on the `unregister` of ExpressionProxy to see if there are any surviving deps, and if there are none, it is torn down. The teardown drops the ref to the fragment and also triggers a teardown on the Computation. That cleans up some of the extra computation stuff that would get overwritten on the next appearance of that signature anyways, and it lets the DOM owned by the fragment go as well.

Thoughts?

### DIFF
--- a/src/model/Computation.js
+++ b/src/model/Computation.js
@@ -157,4 +157,13 @@ export default class Computation extends Model {
 
 		this.dependencies = dependencies;
 	}
+
+	teardown () {
+		let i = this.dependencies.length;
+		while ( i-- ) {
+			if ( this.dependencies[i] ) this.dependencies[i].unregister( this );
+		}
+		if ( this.root.computations[this.key] === this ) delete this.root.computations[this.key];
+		super.teardown();
+	}
 }
--- a/src/model/RootModel.js
+++ b/src/model/RootModel.js
@@ -152,6 +152,16 @@ export default class RootModel extends Model {
 		return this.value;
 	}
 
+	teardown () {
+		const keys = Object.keys( this.mappings );
+		let i = keys.length;
+		while ( i-- ){
+			if ( this.mappings[ keys[i] ] ) this.mappings[ keys[i] ].unregister( this );
+		}
+
+		super.teardown();
+	}
+
 	update () {
 		// noop
 	}
--- a/src/view/resolvers/ExpressionProxy.js
+++ b/src/view/resolvers/ExpressionProxy.js
@@ -114,7 +114,27 @@ export default class ExpressionProxy extends Model {
 		return this.get();
 	}
 
+	teardown () {
+		this.unbind();
+		this.fragment = undefined;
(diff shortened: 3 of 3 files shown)


src/model/Computation.js
```diff
@@ -157,4 +157,13 @@ export default class Computation extends Model {
 
 		this.dependencies = dependencies;
 	}
+
+	teardown () {
+		let i = this.dependencies.length;
+		while ( i-- ) {
+			if ( this.dependencies[i] ) this.dependencies[i].unregister( this );
+		}
+		if ( this.root.computations[this.key] === this ) delete this.root.computations[this.key];
```
evs-chris: I used delete here because there are other sections that assume that all keys in computations have a value. Would it be better to fix those assumptions and just set undefined here? My understanding is delete is slow and can be a barrier to optimization.

fskreuz:
Click-bait-y title :+1:

fskreuz:
Quick look at the timeline. Seeing the familiar saw-like profile, looking good.

image

src/model/Computation.js
martypdx: From [v8 forum](https://groups.google.com/forum/#!topic/v8-users/zE4cOHBkAnY):

> We only track property additions, because these are by far the most common case. Deleting a property results in going to "slow mode", i.e. using a dictionary for the object's properties. So as a general rule of thumb, using 'delete' makes thing slower.

> When you are using an object as a hash map, using delete is OK, because then the object is very probably already in dictionary (= hash map) mode. Playing GC by hand is not necessary and almost always the wrong thing, anyway.

evs-chris: In this case, we're using it as a hash map, so it looks like it's ok?

Rich-Harris:
:tada:

src/model/Computation.js
martypdx: 👍

PR #2480, opened 2016-04-02
Lazy reference expressions, to avoid unnecessary recomputation

Rich-Harris:
**Description of the pull request:**:
Stealing @evs-chris's [fix](https://github.com/ractivejs/ractive/issues/2461#issuecomment-201050948) and claiming it as my own. Don't fully understand how it works but it does!

#2461 

**Is breaking:**
no


evs-chris:
:smile: it just keeps the regular model mark from calling getValue and sets a dirty flag instead. Then once the deps ask for the value, if the model is still dirty, get value is called to do the actual computation. So what was a push relationship becomes a pull relationship - or lazy.

It looks like there may be some extraneous value setting going on in the patch.

src/view/resolvers/ReferenceExpressionProxy.js
```diff
@@ -142,11 +142,27 @@ export default class ReferenceExpressionProxy extends Model {
 	}
 
 	get () {
-		return this.model ? this.model.get() : undefined;
+		const value = this.model ? this.model.get() : undefined;
+		if ( this.dirty ) {
+			this.dirty = false;
+
+			if ( !isEqual( value, this.value ) ) {
+				this.value = value;
```
evs-chris: I think this line can go

Rich-Harris:
Gah. Think we might need to go back to the drawing board on this one – several node tests are failing:

```
1) ractive.toHTML() Single keypath expression:
     TypeError: Cannot read property 'fragments' of undefined
      at Object.runloop.addFragment (tmp/ractive.js:765:21)
      at Fragment.bubble (tmp/ractive.js:13457:16)
      at Interpolator.bubble (tmp/ractive.js:4981:26)
      at Interpolator.handleChange (tmp/ractive.js:6488:10)
      at handleChange (tmp/ractive.js:1166:7)
      at Array.forEach (native)
      at ReferenceExpressionProxy.get (tmp/ractive.js:6235:17)
      at Interpolator.getString (tmp/ractive.js:6537:54)
      at Interpolator.toString (tmp/ractive.js:6569:23)
      at toEscapedString (tmp/ractive.js:1206:14)
      at Array.map (native)
      at Fragment.toString (tmp/ractive.js:13693:23)
      at Ractive$toHTML (tmp/ractive.js:1064:25)
      at tmp/test/node-tests/toHTML.js:35:43
      at Array.forEach (native)
      at Context.<anonymous> (tmp/test/node-tests/toHTML.js:18:20)
```

The problem is that rendering the HTML triggers a `get`, which now has the side-effect of triggering a bunch of updates outside the runloop. I'll poke at it a bit more but obviously we can't merge this as-is

evs-chris:
Hmmm... `scheduleTask` already guards against being called outside of a runloop, and adding guards to the other methods seems to pass the tests. I think I'm with you on going back to the drawing board though. There are some console messages I've never seen before from phantom, but it says all of the tests passed.

I may have mangled the runloop in some ways that are well beyond what it was intended for in order to address issues that popped up with rendering causing keypaths to pop into existence and thus causing fragments to be registered and ignored on the current batch. If that's still copacetic, then guarding may be an acceptable solution.

evs-chris:
Fixed :smile:

PR #2483, opened 2016-04-02
Update AppVeyor config

MartinKolarik:
Removes node `0.10` and `1.0` from AppVeyor.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,8 +1,7 @@
 # Test against this version of Node.js
 environment:
   matrix:
-  # node.js
-  - nodejs_version: "0.12"
+  - nodejs_version: "5"
 
 # Install scripts. (runs after repo cloning)
 install:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,10 +2,7 @@
 environment:
   matrix:
   # node.js
-  - nodejs_version: "0.10"
   - nodejs_version: "0.12"
-  # io.js
-  - nodejs_version: "1.0"
 
 # Install scripts. (runs after repo cloning)
 install:


fskreuz:
Is there a reason to use older versions of Node? Can we just go with 4 and 5?

MartinKolarik:
We could, but as long as we want to support 0.10 / 0.12, we should probably run tests on that. I tried adding 4 and 5 as well, but each build on AppVeyor takes forever to complete, so I decided to go with just 0.12.

Rich-Harris:
Ideal situation would be to build it using Node 5 but then test it using different versions we want to support. Is there a straightforward way to do that? Any reason it wouldn't be possible to use nvm inside Appveyor?

MartinKolarik:
@Rich-Harris if we want to run tests on multiple node versions, I'd do that on Travis. It supports concurrent builds and is much more reliable. We only need to change the setup so that only one node version runs the deployment script.

MartinKolarik:
Yay, four green lights!

Rich-Harris:
> Yay, four green lights!

:cocktail: 

> if we want to run tests on multiple node versions, I'd do that on Travis

That makes sense to me – should we update this config so that it always uses stable node, if it's only to ensure that developers can build the library on Windows? If we did that, and required node>=5 to build the lib, we might even be able to avoid transpiling the test .js files with Babel, which would be nice

MartinKolarik:
>should we update this config so that it always uses stable node, if it's only to ensure that developers can build the library on Windows

Makes sense, I changed it to v5, and am going to setup Travis.

>If we did that, and required node>=5 to build the lib, we might even be able to avoid transpiling the test .js files with Babel, which would be nice

:+1: I think that's a perfectly fine requirement

Rich-Harris:
> we might even be able to avoid transpiling the test .js files with Babel

actually, that's nonsense... we still need to support older browsers! D'oh

fskreuz:
> actually, that's nonsense

The node tests can use ES6 to some degree if run in the latest Node. The browser tests would still need transpilation though.

MartinKolarik:
>actually, that's nonsense... we still need to support older browsers! D'oh

lol, you are right :laughing:

PR #2484, opened 2016-04-02
Test on node v5, v4, v0.12, and v0.10

MartinKolarik:


Rich-Harris:
Nice :+1: Don't we want to only *build* on 5 though? (I realise that's kinda awkward with the convoluted build process we have at present)

MartinKolarik:
>Don't we want to only build on 5 though

I think not. It's easier to see what versions fail this way, and all 4 builds can run in parallel. Running the build only once and then changing node versions to run tests would require running it sequentially in one Travis job.

Rich-Harris:
Fair enough. Was just thinking from the point of view of freeing us of the need to support 0.10 at build time – but we can worry about that another time :+1:

PR #2487, opened 2016-04-03
Build perf

MartinKolarik:
Using `rollup-plugin-babel` means we can't take advantage of gobble's great caching, and all files are processed by babel every time. By using babel directly (for tests and dev Ractive build), the rebuild time is reduced dramatically.

Cuts down rebuild time to ~3s.

**Is breaking:**
No.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,4 +1,0 @@
-{
-	"presets": [ "es2015-rollup" ],
-	"compact": false
-}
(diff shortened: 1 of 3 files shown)


Rich-Harris:
Ultimately we really need to get incremental rebuilds working in Rollup (https://github.com/rollup/rollup/issues/191) since that way we can get fast rebuilds without the downsides of transpiling each file individually (i.e. duplicated Babel helpers). At the moment we're jumping through some fairly complex hoops, and we could make things a lot simpler if more of the work was transferred to Rollup.

In the meantime though, I think this is the right approach, we'll go insane if rebuilds take upwards of 10 seconds post-#2455 :+1:

PR #2488, opened 2016-04-03
Tests & bug fixes

MartinKolarik:
Trying to get tests to pass in all browsers; current status:
 - [x] Chrome - all passing
 - [x] Opera - all passing
 - [x] Edge - all passing
 - [x] Firefox - ~~2 failing~~ - all passing
 - [x] IE 11 - ~~10 failing~~ - all passing
 - [x] IE 10 - ~~136 failing~~ - 106 failing / all passing in emulator or when run separately
 - [x] IE 9 - ~~138 failing~~ - 103 failing / all passing in emulator or when run separately

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -11,6 +11,7 @@
 	"-W041": false,
 	"-W030": false,
 	"esnext": true,
+	"browser": true,
 	"globals": {
 		"process": true,
 		"__dirname": true,
--- a/gobblefile.js
+++ b/gobblefile.js
@@ -129,7 +129,7 @@ test = (function () {
 
 			return sander.readFile( inputdir, 'browser-tests', f ).then( function( data ) {
 				data = data.toString( 'utf8' );
-				if ( data.indexOf( 'initModule(' ) !== -1 ) {
+				if ( f.indexOf( 'polyfills.js' ) !== -1 || data.indexOf( 'initModule(' ) !== -1 ) {
 					modules.push( f );
 					return sander.link( inputdir, 'browser-tests', f ).to( outputdir, 'browser-tests', f );
 				} else {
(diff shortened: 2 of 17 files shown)


MartinKolarik:
Looks like the IE emulator mode is just as shitty as the browser itself. Everything was passing in IE 10 & 9, then I fired up a real IE in VM... another 100+ tests failing there...

evs-chris:
I say we take off and nuke it from orbit... it's the only way to be sure...

I am a bit surprised that emulation mode sucks that hard, but you're awesome for taking it that far :clap:. I dropped my last IE VM years ago.

With IE8 past end of life, and the rest of them following quickly, does anyone know when it's reasonable to rip out that chunk of legacy code and throw a party?

MartinKolarik:
>With IE8 past end of life, and the rest of them following quickly, does anyone know when it's reasonable to rip out that chunk of legacy code and throw a party?

As far as I'm concerned, we could do that right now (at least for IE 8). But really, if we ever release v1.0.0, that would be a good time.

fskreuz:
jQuery 3.0 will still do IE8 :smile: https://blog.jquery.com/2014/10/29/jquery-3-0-the-next-generations/comment-page-1/

MartinKolarik:
>jQuery 3.0 will still do IE8

And Angular 2, Ember 2, and React 0.15 don't/won't :smile:

MartinKolarik:
It seems most of the tests that were passing in an emulator but failing in real IE actually pass when run separately :grin:

Going to merge this, and maybe revisit another time...

fskreuz:
I've seen this behavior a few times in other browsers as well. Could possibly be a QUnit issue, or the way we write the tests.

PR #2496, opened 2016-04-05
Let array methods branch too - RFC

evs-chris:
Ractive will currently branch for you if you `set` an undefined keypath e.g. `r.set( 'foo.0.bar', 'test' )` with an entirely empty data dictionary results in data that looks like `{ foo: [ { bar: 'test' } ] }`. However, if you `push`, `splice`, `unshift`, etc on an undefined keypath, an error is thrown.

This adds a check for `undefined` keypaths and swaps in an empty array instead of throwing an error. This allows regular array methods to be used without having to initialize every keypath that may be used with an array method.

Makes array methods a little more method-call event friendly.

**Is breaking:**
Sort of, but I doubt anyone depends on an exception being thrown when using array methods on undefined keypaths.


evs-chris:
Any objections?

Issue #2494, opened 2016-04-05
<select> with no options matching the bound value should not select the first item


MartinKolarik:
**Description of the bug:**
When a bound `<select>` value is set to something that doesn't match any options, it is ignored, and first `<option>` is selected instead.

Related: http://stackoverflow.com/questions/36409229/using-ractive-js-how-do-you-unselect-a-selected-select-box

**Versions affected:**
both 0.7.3 (`<select>` and model go out of sync) and edge (model value replaced by first `<option>`)

**Reproduction:**
http://jsfiddle.net/7bx6pc8h/1/

**Code:**
```html
{{selected}}<br>
  
<select value="{{selected}}">
  <option>1</option>
  <option>2</option>
</select>
```

```js
var ractive = new Ractive({
  el: 'body',
  template: '#template',
  data: { selected: 'x' }
});

setTimeout(() => {
  ractive.set('selected', 'y');
  console.log(ractive.get('selected')); // logs 1
}, 3000);
```

**Note:**
The current behavior is correct in a situation where there had been a selected `<option>`, but it was removed from the DOM. In that case, we [should select the first option](https://github.com/ractivejs/ractive/issues/675#issuecomment-42466300). However, I think this should not be the case when `ractive.set()` is called explicitly.

evs-chris:
I think this is more or less forced by browsers now. Firefox as of mid-last-year (at least that's the first encounter I had with it) and Chrome as of 49 don't seem to let you have a select with options and no selected value. There's a strong chance I've missed something though :smile:

martypdx:
Try `select.selectedIndex = -1`

MartinKolarik:
@evs-chris I posted [this](https://jsfiddle.net/xubo91oh/) as a solution on SO and it works in all browsers.

PR #2497, opened 2016-04-05
Don't force the first option to be selected in a select with no matching value - #2494

evs-chris:
Removes the forced selection of the first option for bound selects that don't have an option with a matching value.

#2494

**Is breaking:**
No

### DIFF
--- a/src/view/items/element/attribute/getUpdateDelegate.js
+++ b/src/view/items/element/attribute/getUpdateDelegate.js
@@ -104,13 +104,12 @@ function updateSelectValue () {
 
 			if ( optionValue == value ) { // double equals as we may be comparing numbers with strings
 				option.selected = true;
-				break;
+				return;
 			}
 		}
-	}
 
-	// if we're still here, it means the new value didn't match any of the options...
-	// TODO figure out what to do in this situation
+		this.node.selectedIndex = -1;
+	}
 }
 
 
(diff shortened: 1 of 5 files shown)


MartinKolarik:
:+1: for covering the option removal case. Was just about to add the test myself.

MartinKolarik:
Now the question is, why does this break another test in phantom.

evs-chris:
:smile: was just about to ask that... I've noticed that different nodes with the same phantom behave differently.

Interestingly, it breaks in IE11 too, but it is fine in Firefox and Chrome. Off to suffer through the IE devtools...

MartinKolarik:
>I've noticed that different nodes with the same phantom behave differently.

It's actually different phantoms on the same node. Coveralls uses phantom 2.x.x, and the coverage report is only generated on node 5, that's why it sometimes fails when others don't.

>Off to suffer through the IE devtools...

Arguably better that debugging this in phantom :smile:

evs-chris:
> It's actually different phantoms on the same node.

Ah! That makes a bit more sense. I somehow got it stuck in my head that the latest phantom was used everywhere.

I found the issue for IE and Edge, and it's weird that it didn't affect anything else. If github decides to see the commit I just pushed anytime soon, hopefully they'll pass CI this round :smile:

MartinKolarik:
phantom still dies (you can `npm run coverage` to test locally)

evs-chris:
It looks like github was stuck. Everything's passing locally and it turned green after that last kick, so I'm going to roll with it.

heavyk:
if you `git commit --ammend && git push --force` -- it'll also restart the ci

Issue #2503, opened 2016-04-07
Ractive.js v0.8.0-edge observe *.* problem


Jeff17Robbins:
**Description of the bug:**
`observe('* *.*', ...)` works in Ractive.js v0.7.3 but doesn't in Ractive.js v0.8.0-edge

**Versions affected:**
Ractive.js v0.8.0-edge

**Platforms affected:**

**Reproduction:**
Ractive.js v0.7.3 (works)
http://jsfiddle.net/bhpamt0s/

![r 7 3 wildcards](https://cloud.githubusercontent.com/assets/2102919/14359838/6a3a6ec4-fcc1-11e5-88aa-f5256f245a64.png)

Ractive.js v0.8.0-edge (doesn't)
http://jsfiddle.net/xsrmz3bx/

![r 8 0 wildcard](https://cloud.githubusercontent.com/assets/2102919/14359847/79130938-fcc1-11e5-82b7-c796e30af83f.png)

**Code:**
HTML
```html
<div id="container"></div>
<script id="template" type="text/ractive">
    <div>{{counter}}</div>
    <ul>
    {{#list}}<li>{{.}}</li>{{/list}}
    </ul>
</script>
```

JavaScript
```javascript
ractive = new Ractive({
    el: 'container',
    template: '#template',
    data: {counter: 0, list:['a', 'b', 'c']}
});

var obs = ractive.observe('* *.*', function(newValue, oldValue, key) {
    console.log(key, newValue);
});

ractive.add('counter');
```

evs-chris:
Fixed. Thanks for the report!

PR #2505, opened 2016-04-09
Don't throw when a pattern observer encounters a value type - #2503

evs-chris:
Instead of throwing when a pattern observer tries to check a value type (not object, array, or function) for children, this returns false and has the caller check to see if any children were actually returned.

#2503

**Is breaking:**
Nope

### DIFF
--- a/src/model/Model.js
+++ b/src/model/Model.js
@@ -219,7 +219,8 @@ export default class Model {
 			if ( key === '*' ) {
 				matches = [];
 				existingMatches.forEach( model => {
-					matches.push.apply( matches, model.getValueChildren( model.get() ) );
+					const children = model.getValueChildren( model.get() );
+					if ( children ) matches.push.apply( matches, children );
 				});
 			} else {
 				matches = existingMatches.map( model => model.joinKey( key ) );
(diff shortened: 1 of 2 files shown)


src/model/Model.js
```diff
@@ -219,7 +219,8 @@ export default class Model {
 			if ( key === '*' ) {
 				matches = [];
 				existingMatches.forEach( model => {
-					matches.push.apply( matches, model.getValueChildren( model.get() ) );
+					const children = model.getValueChildren( model.get() );
+					if ( children ) matches.push.apply( matches, children );
```
fskreuz:
```js
const children = model.getValueChildren( model.get() );
matches.push(...children);
```

src/model/Model.js
```diff
@@ -307,8 +307,7 @@ export default class Model {
 		}
 
 		else if ( value != null ) {
-			// TODO: this will return incorrect keypath if model is mapped
-			throw new Error( `Cannot get values of ${this.getKeypath()}.* as ${this.getKeypath()} is not an array, object or function` );
+			return false;
```
fskreuz: The `|| []` above can be removed if `getValueChildren` is always guaranteed to return an array. In this case, an empty one.

evs-chris: :+1: I was about to do that before just before I submitted the PR, but decided not to because reasons :smile:

Issue #2066, opened 2015-08-12
Re-add Ractive.noConflict


Madgvox:
As per https://github.com/ractivejs/ractive/issues/1547#issuecomment-130145501, the Ractive.noConflict function has ~mysteriously disappeared~

JonDum:
Is noConflict even needed? @Ghazgkull Would you mind chiming in why you need it?

Here's the current module definition in edge:

```
(function (global, factory) {
  typeof exports === 'object' && typeof module !== 'undefined' ? module.exports = factory() :
  typeof define === 'function' && define.amd ? define(factory) :
  global.Ractive = factory()
}(this, function () { 
```
Only gets added to global if there's *no* detected module system at all.

Do developers that don't use a module system but care about global pollution exist?

Ghazgkull:
I use a module system (Browserify), I care about global pollution, and I need Rangy.noConflict because when you load Rangy asynchronously it stomps on the window.rangy property.

In my case, I have a bit of Javascript (transpiled with Browserify) that gets loaded onto a 3rd party website. The first thing I do is turn around and load my dependencies (jQuery, Ractive, Rangy) by inserting new script tags into the DOM. And when Rangy loads, it sets the global window.rangy property. jQuery does the same thing, but using noConflict it's easy to reset the global var as soon as the script loads.

For now I'm taking a poor-man's approach where I squirrel away the value of window.rangy before I insert the script tag to load my own version and then restore it manually. But it would be nicer to just use noConflict.

JonDum:
That doesn't convince me of the need for noConflict.

You could easily include your dependancies into your browserify bundle instead of hackily injecting script tags and you'd have the benefit of removing additional HTTP requests which will more than likely be an improvement to cold load times anyways.

Ghazgkull:
Whether to bundle required libraries into an application or to reference them externally (e.g. from a CDN) is a decision that has to be made for each application. Sometimes it makes sense to bundle things together and sometimes it absolutely doesn't.

In my case, warm load times are much more important than cold load times. And my application Javascript changes much (much!) more frequently than my required libraries. So I get much faster load times by keeping my libraries out of my app bundle and instead letting the browser hit a CDN for a 304 that returns a few bytes of data.

There's nothing hacky about injecting script tags into the DOM. It's the normal way to load additional Javascript and CSS files.

martypdx:
[...]

PR #2506, opened 2016-04-10
Reinstate Ractive.noConflict()

evs-chris:
Adds a step to the build process to add a `Ractive.noConflict()` method that sets the `Ractive` global back to whatever it was when this particular Ractive file loaded and returns the newly removed value e.g. the current Ractive. This only affects those using no module loader.

This is slightly brittle, as it depends on a specific string coming out in the build output, but there is a test canary in the coal mine. 

#2066

**Is breaking:**
Nope

### DIFF
--- a/gobblefile.js
+++ b/gobblefile.js
@@ -56,6 +56,11 @@ function bloodyIE8 ( src ) {
 	return src;
 }
 
+function noConflict ( src ) {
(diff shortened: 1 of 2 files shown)


fskreuz:
I'm curious why this is in the build instead of the actual source.

evs-chris:
Because the build sets up the UMD boilerplate, which has all of the globals and whatnot in it. So there's nowhere in the source to put it because it doesn't exist until it is built.

Issue #2500, opened 2016-04-06
Ractive.js v0.8.0-edge update problem with dotted keypath and adaptor


Jeff17Robbins:
**Description of the bug:**
Ractive.js v0.8.0-edge `update()` using a dotted keypath into an adaptor is broken.  Works in Ractive.js v0.7.3.

**Versions affected:**
	Ractive.js v0.8.0-edge
	Wed Apr 06 2016 06:02:37 GMT+0000 (UTC) - commit 1507b3d61caa8225205a509b5bad8c93495b9693

**Platforms affected:**

**Reproduction:**
Ractive.js v0.7.3 works:
http://jsfiddle.net/yyuhmvx5/
`a=5`
`b=5`

Ractive.js v0.8.0-edge doesn't work:
http://jsfiddle.net/noxLjqzv/
`a=3`
`b=5`

**Code:**
HTML
```html
<div id="container"></div>
<script id="template" type="text/ractive">
    {{#adapted:key}}
    <div>{{key}} = {{.}}</div>
    {{/adapted}}
</script>
```
JavaScript
```javascript
(function() {
    var Adaptor = Ractive.adaptors.Test = {
        filter: filter,
        wrap: wrap
    };

    function filter(child, keypath, parent) {
        if ((child === undefined) || !(child.adapted))
            return false;
        if (parent &&
            parent._ractiveWraps &&
            parent._ractiveWraps[keypath])
            return false;
        return true;
    }

    function wrap(parent, child, keypath, prefixer) {
        markAsWrapped();
        return {
            get: get,
            teardown: teardown
        };

        function get() {
            var ret = {};
            for (var key in child.adapted)
                ret[key] = child.adapted[key] + 1;
            return ret;
        }

        function teardown() {
            delete parent._ractiveWraps[keypath];
        }

        function markAsWrapped() {
            if (!parent._ractiveWraps) parent._ractiveWraps = {};
            parent._ractiveWraps[keypath] = child;
        }
    }
}());

ractive = new Ractive({
    el: 'container',
    template: '#template',
    adapt: ['Test'],
});

var wrapped = {
    adapted: {
        a: 1,
        b: 4
    }
};

ractive.set('adapted', wrapped);

setTimeout(function() {
    wrapped.adapted.a = 2;
    ractive.update('adapted');
    wrapped.adapted.a = 4;
    ractive.update('adapted.a');
}, 2000);
```

Jeff17Robbins:
[...]

Issue #2090, opened 2015-08-31
Ractive 0.8.0-edge onchange bug


Jeff17Robbins:
A stripped-down Ractive instance, with no container and no template, doesn't fire its onchange listener.

```
<!doctype html>
<html lang='en'>

<head>
    <meta charset='UTF-8'>

    <title>Ractive 0.8.0-edge onchange test</title>

    <script type='text/javascript' src='http://cdn.ractivejs.org/edge/ractive.js'></script>

    <script type='text/javascript'>
    var r = new Ractive({
        data: {
            test: 1,
           }
    });

    r.on('change', function(updates) {
        console.log(updates);
    });

    r.set('test', 2);
    </script>
</head>

<body>
</body>

</html>
```

The problem appears to be in `flushChanges()`.  There is a call to `changeHook.fire(ractive, ractive.viewmodel.changes)`, but it is driven by a `batch.fragments` loop that is skipped because `batch.fragments` is zero length.  

Also, the `changeHook.fire()` section is marked with `TODO deprecate`.

Yet if I manually call `changeHook.fire(r, r.viewmodel.changes);` while stepping into `flushChanges()` in Chrome, my `r.on('change')` function gets called.  

Suggesting that in refactoring the code, this simplistic use-case perhaps got forgotten?

```
  function flushChanges() {
  	var i, thing, changeHash;

  	batch.immediateObservers.forEach(dispatch);

  	// Now that changes have been fully propagated, we can update the DOM
  	// and complete other tasks
  	i = batch.fragments.length;
  	while (i--) {
  		thing = batch.fragments[i];

  		// TODO deprecate this. It's annoying and serves no useful function
  		var ractive = thing.ractive;
  		changeHook.fire(ractive, ractive.viewmodel.changes);
  		ractive.viewmodel.changes = {};

  		thing.update();
  	}
  	batch.fragments.length = 0;

  	batch.transitionManager.start();

  	batch.deferredObservers.forEach(dispatch);

  	var tasks = batch.tasks;
  	batch.tasks = [];

  	for (i = 0; i < tasks.length; i += 1) {
  		tasks[i]();
  	}
```

Jeff17Robbins:
@evs-chris I feel like an ingrate (thank you for working on this!) but I found a problem perhaps more about `update()` than about the `change` event.  I opened it as a new issue #2518.  Calling `update()` triggers the `change` event, but the value for the updated keypath is missing.

PR #2508, opened 2016-04-11
Adaptor updates - #2500 and maybe more

evs-chris:
This:

* consolidates all of the adaptor handling in model into the `adapt` method
* adds a check in `ractive.update` for an adapted parent and delegates the update to the parent if it is adapted
* adds reset methods to all of the provided adaptors that don't cause them to teardown if set with their current value

I think that covers the rest of the weird cases of adaptors, but I'm not sure as I haven't used them extensively. This may also fix some of the magic failures in edge, as it adds a working `reset`. At least I think it is a working `reset`... can anyone comment on the implementation?

Have I missed or completely misunderstood anything?

#2500 

**Is breaking:**
I don't think so

**Reviewers:**
Anyone familiar with how adaptors should behave

### DIFF
--- a/src/Ractive/static/adaptors/array/index.js
+++ b/src/Ractive/static/adaptors/array/index.js
@@ -50,6 +50,11 @@ class ArrayWrapper {
 		return this.value;
 	}
 
+	reset ( value ) {
+		if ( this.value === value ) return true;
+		return false;
+	}
+
 	teardown () {
 		var array, storage, wrappers, instances, index;
 
--- a/src/Ractive/static/adaptors/array/index.js
+++ b/src/Ractive/static/adaptors/array/index.js
@@ -51,8 +51,7 @@ class ArrayWrapper {
 	}
 
 	reset ( value ) {
-		if ( this.value === value ) return true;
-		return false;
+		return this.value === value;
 	}
 
 	teardown () {
--- a/package.json
+++ b/package.json
@@ -90,7 +90,6 @@
     "karma-phantomjs-launcher": "~1.0",
     "karma-qunit": "~0.1",
     "mocha": "^2.1.0",
-    "phantomjs": "^1.9.20",
     "phantomjs-prebuilt": "~2.1",
     "promises-aplus-tests": "^2.1.0",
     "qunitjs": "~1.23",
--- a/src/Ractive/prototype/shared/makeArrayMethod.js
+++ b/src/Ractive/prototype/shared/makeArrayMethod.js
@@ -18,6 +18,7 @@ export default function ( methodName ) {
 		const result = arrayProto[ methodName ].apply( array, args );
 
 		const promise = runloop.start( this, true ).then( () => result );
+		promise.result = result;
 
 		if ( newIndices ) {
 			model.shuffle( newIndices );
--- a/src/Ractive/prototype/update.js
+++ b/src/Ractive/prototype/update.js
@@ -11,6 +11,8 @@ export default function Ractive$update ( keypath ) {
 		this.viewmodel.joinAll( keypath ) :
 		this.viewmodel;
 
+	if ( model.parent && model.parent.wrapper ) return this.update( model.parent.getKeypath( this ) );
+
 	const promise = runloop.start( this, true );
 
 	model.mark();
--- a/src/Ractive/static/adaptors/magic.js
+++ b/src/Ractive/static/adaptors/magic.js
@@ -97,8 +97,8 @@ class MagicWrapper {
 		return this.value;
 	}
 
-	reset () {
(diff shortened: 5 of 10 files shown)


src/Ractive/static/adaptors/array/index.js
```diff
@@ -50,6 +50,11 @@ class ArrayWrapper {
 		return this.value;
 	}
 
+	reset ( value ) {
+		if ( this.value === value ) return true;
+		return false;
+	}
```
fskreuz:
```js
reset ( value ){
  return this.value === value;
}
```

evs-chris: :+1: I had a stashing derp whilst testing something just before commit.

Jeff17Robbins:
We have been using https://github.com/rstacruz/ractive-ractive in Ractive.js v0.7.3, and would like to do so in Ractive.js v0.8.0.  

Aside from the update issues I've been trying to reproduce (and @evs-chris  has been fixing -- thank you!), there's also issue #2090.  

The Ractive "change" event doesn't fire for template-less Ractive instances.  (In fact, it appears that ractive.viewmodel.changes accumulates the changes for a template-less ractive instance, but `flushChanges` doesn't handle the template-less or lacking-a-fragment case, and so the viewmodel.changes simply accumulate, never to be fired via `changeHook.fire(ractive, ractive.viewmodel.changes);`.)

So...if the https://github.com/rstacruz/ractive-ractive adaptor is intended to work in Ractive.js v0.8.0, the Ractive lifecycle event `change` needs to be fixed to work on template-less Ractive instances.  

This comment likely doesn't really belong on this PR, but I don't know how to flag the whole adaptor topic, and would like to help with making adaptors work.

aharpervc:
I agree with @Jeff17Robbins that ractive-ractive is a huge boon and would be a giant problem if there was no way to achieve a similar result in 0.8.0 (which might as well be called v1 at this point :skull: )

martypdx:
@evs-chris I think there are three things that adaptors get used for, it might be worth separating them out:

1. Mediating between a data source and the ractive instance model. Most classic use and what people generally think of as adaptors. Your spot on that we need to have dedicated array modification methods. One of the other needed additions to delineate requests made for child properties for the purpose of access a further child property versus for a value to be used in the template. For me, this came up with working with Firebase as getting a childen reference has a much different effect than getting a child value (the former is a reference, the later fetches data from the server).
2. Mediating between ractive model and the template. For the most part, this should be handled by computed properties, but I've seen it crop up as an adaptor when devs try to apply to a testable conditions, for example show any boolean value as Y or N. If it's between the model and template, IMO it shouldn't be an adaptor
3. Extending the ractive API. The data needs a way to tell ractive about changes, the inverse of #1. Usually this would be in conjunction with first point, but interestingly both the array and magic adaptors fall in this category, but not the first.

Jeff17Robbins:
@martypdx thank you for the three use cases.  I thought it might be worth sharing what we have been doing with the [Ractive-Ractive adaptor](https://github.com/rstacruz/ractive-ractive).  

We have been using it to create a template-less Ractive instance that acts as a "store" to several other Ractive instances that act as "views".  We use it one-way only, meaning that we only need data flowing one-way from the store to the views.

While there is nothing mandating that we use a Ractive instance as our store, it seems handy because we then also get computeds in our store, for free, so to speak.  Since many views can share one store, it seems clean to us.  

If there is another way to manage a single external model, and somehow map it into models of our views, we could use that other way. The basic requirements are unsurprising:

1. A single location to deposit incoming JSON data.  
2. A way to then update **efficiently** several different Ractive instances with that data.

Our views use mustaches like `{{store.somevalue}}` or `{{store.someobj.someprop}}` in their templates.  The initial part of the path 'store' is meant to pick out the store Ractive instance.

In some ways, our use is like your use-case 1, except we want a staging area (aka "store") to first deposit incoming data so that we can then have many Ractive instances ("views") sharing this data.  But in another sense, it seems like your use-case 3, in that when our store gets new data, we need it to tell ractive (the ractive instances acting as "views") about the changes.  If we can continue to use the Ractive-Ractive adaptor, that would be easiest for us.  But if there is a better (or equivalent) way to model this in [Ractive.js v0.8.0](http://cdn.ractivejs.org/edge/ractive.js), we are totally open to a different approach.

Any suggestions?

evs-chris:
I think I've managed to mangle the various adaptor scenarios into working, but it could certainly use some review. Adaptors are now a bit more expensive during shuffles than they used to be because the model now tears adaptors on shuffled paths down so that they don't get out of whack with further updates. So the test for #2005 is now passing in this PR too.

This solution is probably not ideal, but I'm pretty sure a more reasonable solution involves breaking the adaptor API.

sabob:
Looks good.

PR #2515, opened 2016-04-15
Switch from Babel to Bublé

Rich-Harris:

Some of you know that last week I wrote an ES2015 compiler called [Bublé](http://buble.surge.sh/). This PR switches our build process over to use it instead of Babel.

Build performance. On this machine, `npm start` took over **15** seconds for the initial build (44 seconds with npm2, because Babel is rather slow to start up if you install it with npm2); it now takes around **6** seconds with Bublé. A complete `npm run build` drops from **43** to **23** seconds.

The resulting build is also slightly leaner, shaving 4kb off the full minified build (24kb off the unminified build). Bublé doesn't rewrite any code that it doesn't need to, so the end result also reflects the original source code more closely.

**Is breaking:**

No


evs-chris:
:+1: :tada: Another fun stat that has dropped significantly: Travis lost a full 2 minutes of build time. Though old node apparently isn't a fan of Bublé at this point.

Rich-Harris:
> Though old node apparently isn't a fan of Bublé at this point.

Ah, whoops! Looks like there's an `array.find(...)` in there (and possibly some other stuff) which isn't supported. Will fix that after lunch

fskreuz:
:+1: I'm all in on this one since Ractive doesn't use the funny things from ES6+

fskreuz:
It just occurred to me, isn't Buble targetting ES5? What about the legacy build? (unless we'll start to drop those).

MartinKolarik:
Wanted to try this, but getting `error fetch failed https://registry.npmjs.org/buble/-/buble-0.4.19.tgz` on `npm install`

Rich-Harris:
@MartinKolarik Weird – seems like npm lost the file. Have bumped version, try now – seems to be working

@fskreuz Bublé is doing the same job Babel was, i.e. converting arrow functions and classes and the like (and rewriting `promise.catch(...)` to `promise['catch'](...)`. It's orthogonal to the question of legacy builds

evs-chris:
It may actually be a good bit better for legacy builds because babel's helpers aren't particularly es3ish friendly... and they get peppered everywhere.

MartinKolarik:
@Rich-Harris works now, and seems to be about 50% faster, so :+1: from me.

PR #2516, opened 2016-04-15
Faster builds

Rich-Harris:
Follow-up to #2515 – because Bublé doesn't have 'helpers', there's no benefit to deferring compilation in order to deduplicate them. This means we can convert ES2015 -> ES5 once, and bundle from there, saving us some time. It *does* mean that non-legacy builds are getting the `.catch` -> `['catch']` transform, but that probably adds about 9 bytes so ¯\ _ (ツ) _ /¯

I've also made the build file a bit DRYer.


evs-chris:
Cold build down to 6s for me 😁

PR #2519, opened 2016-04-21
Detach non-transitioning nodes immediately - #2027 - RFC

evs-chris:
This adds a check at the beginning of the transition cycle for nodes that are not up- or downstream of a transitioning node and goes ahead and detaches them immediately.

I'm not entirely sure what this means for nested batches and transitions, but a quick run through looks like it won't release zalgo. I have a feeling ~~that to be 100% accurate this would have to also check parent (and child) manager nodes and transitions, but I also have the feeling that~~ there's no way to cover every corner as parent managers won't be able to check all of their children that may spring into existence.

Thoughts?

#2027

**Is breaking:**
I don't think so.

### DIFF
--- a/src/global/TransitionManager.js
+++ b/src/global/TransitionManager.js
@@ -100,28 +100,19 @@ function check ( tm ) {
 // check through the detach queue to see if a node is up or downstream from a
 // transition and if not, go ahead and detach it
 function detachImmediate ( manager ) {
+	if ( manager.totalChildren ) return;
+
 	const queue = manager.detachQueue;
 	const outros = manager.outros;
 
-	let i = queue.length;
+	let i = queue.length, j = 0, node, trans;
 	start: while ( i-- ) {
-		const node = queue[i].node;
-		let j = outros.length;
+		node = queue[i].node;
+		j = outros.length;
 		while ( j-- ) {
-			// check to see if the node is a parent of the transition
-			const trans = outros[j].node;
-			let parent = trans;
-			while ( parent ) {
-				if ( parent === node ) continue start;
-				parent = parent.parentNode;
-			}
-
-			// check to see if the transition is a parent of the node
-			parent = node;
-			while ( parent ) {
-				if ( parent === trans ) continue start;
-				parent = parent.parentNode;
-			}
+			trans = outros[j].node;
+			// check to see if the node is, contains, or is contained by the transitioning node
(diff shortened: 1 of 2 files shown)


JonDum:
Yessss! #2027 is really annoying. Caused all sorts of weird flickering and layout issues.

src/global/TransitionManager.js
```diff
@@ -95,3 +96,36 @@ function check ( tm ) {
...
+				if ( parent === trans ) continue start;
+				parent = parent.parentNode;
+			}
+		}
+
+		// no match, we can drop it
+		queue[i].detach();
+		queue.splice( i, 1 );
+	}
+}
```
martypdx: How about:

```js
let i = queue.length, j = 0, node = null, parent = null;
while ( i-- ) {
    node = queue[i].node;
    j = outros.length;
    while ( j-- ) {
        // check to see if the node and transition are same or related:
        trans = outros[j].node;
        if ( trans === node || trans.contains( node ) || node.contains( trans ) ) break;

  		// no match, we can drop it
  		queue[i].detach();
  		queue.splice( i, 1 );
 	 }
}

evs-chris: Oh, yeah, sure, if you want to be efficient and all that. :smile: Didn't know there was a `Node.contains` and have no idea why my googling didn't manage to turn it up. :+1:

evs-chris:
Turns out that checking up and down the transition manager hierarchy is necessary so that components don't get dropped before they're ready. So managers that have children when they are started will not kick off immediate removal for now. I _think_ that's a realtively uncommon case. The fix is _probably_ to collect all the nodes and transition up and down the hierarchy when checking for immediate removal, but it's only a one-way association right now and I'm not sure it's worth the refactor for that particular corner case. Of course, I have a long and storied history of being wrong. :smile:

evs-chris:
I somehow completely missed the children collection on the transition manager. I think it was the counters that threw me. At any rate, this now checks the whole hierarchy of transitions when considering whether or not a node can be detached immediately.

Does that look ok?

martypdx:
Are we expecting fairly small numbers of outros at any given time that we'd have to dig through?

evs-chris:
I would think so, at least not more than a hundred or so. I'd guess more than a few tens would make most browsers stutter pretty badly anyways.

Issue #2427, opened 2016-03-07
Bug: input field and data gets updated despite it shouldn't


dagnelies:
Hi,

As usual, here is the fiddle:

http://jsfiddle.net/xc2o8y08/31/

The other input field (and data) sometimes get updated despite it shouldn't. Note that the behavior is rather odd. If the text of the input fields is the same, they'll get updated (and the data too), but only some of them, and only those "after" it.

This concerns both latest and edge.

dagnelies:
There is some kind of loopback effect here. Because the data changed, the sorting is triggered again, then it seems the data is set again, despite it shouldn't.

evs-chris:
I've tracked this back to a locking issue that will hopefully get resolved along with all of the observer issues. What's happening is the element is being updated and locked while the computation is also taking place, so the correct value gets shuffled down to its new location while its replacement can't update the locked element.

evs-chris:
I was trying to come up with an alternate fix for this, and I think I found a reasonable one that fails part of a test for #771. The fix is to change locking from a boolean to an equality check against the value set in the binding - so in binding `this.attribute.lock = value` and in update delegate `if (this.lock !== value)`.

The [failing test](https://github.com/ractivejs/ractive/blob/dev/test/browser-tests/twoway.js#L422) is expecting the value not to update when the observer sets it because it is locked. Only the first assertion fails because the value gets set to `BAR` when the lock is expecting it to be `bar` and thus allows it to change. The rest of the test seems to be fine.

*I* think that this test case is a bit bizarre because it expects the model and view to be out of sync on purpose. It seems like modifying an input with an observer should be handled by a lazy binding, which is now possible at the element level. Locking is still required because without it, the value gets reset with every char on non-lazy bindings, which makes the cursor jump to the  beginning.

So, I would propose that we drop support for observers getting the model and view out of sync in favor of lazy bindings and locking with the value instead of a non-specific true/false. Thoughts?

martypdx:
> modifying an input with an observer should be handled by a lazy binding

What does that mean? Can you explain "handling by a lazy binding"?

evs-chris:
I'll see if I can ramble through it :smile:

The current behavior, which the test relies on, is:

1. a non-lazy two-way binding causes an update to the model (say, the user typed 'f')
2. the observer for that value then fires and also updates the model (toUpperCase - so 'F' is now in the model)
3. the time to update the attribute rolls around, and it is locked, so the attribute keeps its original value ('f')

[...]

Issue #771, opened 2014-05-30
Validation (by way of observer) doesn't work post-blur


Rich-Harris:
See http://jsfiddle.net/rich_harris/7KSby/, by way of #746

PR #2525, opened 2016-04-23
(Partially) fix bindings that cause themselves to be updated - #2427

evs-chris:
Attributes with bindings currently get locked while the binding is updating to avoid input jitter (cursor jumping home) that would be caused as the value updated itself. Unfortunately, if the attribute has a legitimate reason to change while the binding updates, like the binding causing a computed to update and change the underlying value for the attribute, then the wrong value gets stuck in the attribute because it is locked.

The locking is also in place to allow observers to do validation on a value as it is being updated while also not breaking the input by overwriting its value should the validation observer make changes. I think this behavior should be removed because a lazy binding skirt that particular issue without doing weird things like having the input value and its underlying model have two different values. This seems to have originated in #771, and I think individual binding laziness should make that mostly unnecessary now. That would be breaking though.

In order to avoid breaking changes, this attempts to mostly handle both cases by checking to see if an observer may have done an update if the attribute is locked. If there is no chance that an observer has caused an update (the current batch has had no children) and the model value is not equal to the input value, then change is allowed through. This will fall over if there happens to be an observer that makes a change while a computed also causes a change, but ¯\_(ツ)_/¯

#2427 excepting the corner-case

**Is breaking:**
No

### DIFF
--- a/src/global/runloop.js
+++ b/src/global/runloop.js
@@ -15,6 +15,7 @@ const runloop = {
 			promise = new Promise( f => ( fulfilPromise = f ) );
 		}
 
+		// TODO: this is a temporary hack that needs to go away with a probably breaking change
 		if ( batch ) batch.children++;
 
 		batch = {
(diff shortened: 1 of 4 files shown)


fskreuz:
Code looks good to me.

Issue #2535, opened 2016-05-04
Lazy attribute on inputs cannot be dynamic


ISNIT0:
**Reproduction:**
0.7.3:
http://jsbin.com/yirapomeru/7/edit?html,js,output
Here the attribute is ignored if it's a ractive value

0.8:
http://jsbin.com/yirapomeru/5/edit?html,js,output
Here, the value of the attribute when the input is respected, but changes do not take effect.

**Code:**
`<input type='text' value='{{foo}}' lazy='{{lazy}}' />`

evs-chris:
Fixed in edge :smile:

PR #2536, opened 2016-05-04
ISNIT0-2535 Lazy Attribute now supports dynamism

ISNIT0:
I have added a middleman which intercepts input and keyup events on all elements, it then checks for laziness every time an event is fired. This may not be the best way, but It's the most sensible I can find with my limited knowledge of the codebase.

Tests included!

Issue #2535

### DIFF
--- a/src/view/items/Element.js
+++ b/src/view/items/Element.js
@@ -52,7 +52,7 @@ export default class Element extends Item {
 		if ( this.template.a ) {
 			Object.keys( this.template.a ).forEach( name => {
 				// TODO process this at parse time
-				if ( name === 'twoway' || name === 'lazy' ) return;
+				if ( name === 'twoway' ) return;
 
 				const attribute = new Attribute({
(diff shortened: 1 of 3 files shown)


ISNIT0:
I don't think that failing build is my fault?

evs-chris:
That is a reasonable approach. This is also one of the things covered by #2369 in a slightly different way.

Did you run a build locally (`npm run build`)? Phantom is pretty sensitive about async input tests, but it looks like there are two that are actually broken. I'm not sure about the failing observer test (but I hate that test... it's a tricky one to keep green :smile:), but for the twoway one, `0` is considered `true` because `<input lazy />` results in `lazy === 0`.

ISNIT0:
@evs-chris Just saw your comment, I originally didn't realise the defaults, but just updated the PR :)

evs-chris:
This is now available in edge :smile:

Issue #2545, opened 2016-05-20
Suggestion: track-by in <select />


guilhermeaiolfi:
Hi,

I was wondering how everybody does what seems to be a pretty common use case when using the select element in ractive: 
### Use Case

You have your model loaded from the server and want to show it AND edit it, something that looks like:

``` javascript
{
  name: 'Record #1',
  owner: {
     id: 1
     name: 'User Name'
  }
}
```

You bring the name of the user because you may be listing tons of records and you don't want to go back to the server and get the user name one by one, so it came with it.

But you want to edit this object occasionally and so you create a form to have:

``` html
<select value="{{owner.id}}">
   {{#each users}}
       <option value="{{id}}">{{name}}</option>
   {{/each}}
</select>
```

By this time you'll change the user ID but not its name. To change its name you'll need to go over the `users` array and look for the one that has the new `owner.id`. You can do that with a computed property to automate a little bit, but if you have a deeply nested model you'll have to manually go over your data and sync everything.
### Suggestion

So, I was thinking about it and thought that maybe if we were able to tell the select to compare by some of the object properties instead of the whole object we could set the object directly into the model, like:

``` html
<select value="{{owner}}" track-by="id">
   {{#each users}}
       <option value="{{this}}">{{name}}</option>
   {{/each}}
</select>
```

This would be doing some things:
- it wouldn't set the object into the option, it would use the `track-by` property, get the `id` of `{{this}}` and set it in the `<option>` element AND keep a reference for `{{this}}` somewhere else.
- when `{{owner}}` changes, it will go over the options and look by the option that has `option.id` to compare. 
- Same thing goes for the other way: when the selected option changes it gets the `{{this}}` object of the selected option and sets it in the `{{owner}}`.

It doesn't need to change the current behavior. The change would be for when there is a `track-by` in the select element.

What do you think?

MartinKolarik:
Just a note that if we end up implementing this, it should probably work with `radio` and `checkbox` inputs as well.

fskreuz:
Sounds like a good idea, reminds me of Angular 1.x and all the goofy expressions it has for selects and repeats. Just not convinced with `track-by` tho. I have a gut feeling there's a better way to express this than another unofficial property.

Side note: We should start putting these properties like `ractive-*`. I'm losing track of what Ractive has with all the new changes (esp. transitions).

martypdx:
@guilhermeaiolfi I don't understand your use case. Can you provide example of what you mean by:

[...]

PR #2547, opened 2016-05-22
<select> comparator attribute

guilhermeaiolfi:
**NOT READY TO MERGE**
It's just to see if that's the right way to go. It needs to work for radios and checkboxes too.

This tries to fix the problem when binding objects in your <select> and the two objects that it compares are different using a simple '==' but for the developer perspective they are the same. So the developer is able to define `<select value-comparator=X">`, X being a string or a function reference to compare the two objects.

#2545

**Is breaking:**
NO

### DIFF
--- a/src/view/items/element/binding/CheckboxNameBinding.js
+++ b/src/view/items/element/binding/CheckboxNameBinding.js
@@ -9,7 +9,8 @@ const push = [].push;
 function getValue() {
 	const all = this.bindings.filter(b => b.node && b.node.checked).map(b => b.element.getAttribute( 'value' ));
 	let res = [];
-	all.forEach(v => { if ( !arrayContains( res, v ) ) res.push( v ); });
+	const binding = this.bindings[0];
+	all.forEach(v => { if ( !binding.arrayContains( res, v ) ) res.push( v ); });
 	return res;
 }
 
@@ -35,7 +36,7 @@ export default class CheckboxNameBinding extends Binding {
 			const existingValue = this.model.get();
 			const bindingValue = this.element.getAttribute( 'value' );
 
-			if ( !arrayContains( existingValue, bindingValue ) ) {
+			if ( !this.arrayContains( existingValue, bindingValue ) ) {
 				push.call( existingValue, bindingValue ); // to avoid triggering runloop with array adaptor
 			}
 		}
@@ -73,10 +74,10 @@ export default class CheckboxNameBinding extends Binding {
 		this.isChecked = this.element.node.checked;
 		this.group.value = this.model.get();
 		const value = this.element.getAttribute( 'value' );
-		if ( this.isChecked && !arrayContains( this.group.value, value ) ) {
+		if ( this.isChecked && !this.arrayContains( this.group.value, value ) ) {
 			this.group.value.push( value );
-		} else if ( !this.isChecked && arrayContains( this.group.value, value ) ) {
-			removeFromArray( this.group.value, value );
+		} else if ( !this.isChecked && this.arrayContains( this.group.value, value ) ) {
+			this.removeFromArray( this.group.value, value );
 		}
 		super.handleChange();
 	}
@@ -90,11 +91,10 @@ export default class CheckboxNameBinding extends Binding {
 		const bindingValue = this.element.getAttribute( 'value' );
 
 		if ( isArray( existingValue ) ) {
-			this.isChecked = arrayContains( existingValue, bindingValue );
+			this.isChecked = this.arrayContains( existingValue, bindingValue );
 		} else {
-			this.isChecked = existingValue == bindingValue;
+			this.isChecked = this.element.compare( existingValue, bindingValue );
 		}
-
 		node.name = '{{' + this.model.getKeypath() + '}}';
 		node.checked = this.isChecked;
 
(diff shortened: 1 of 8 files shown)


guilhermeaiolfi:
Since this if my first contribution, I would like extra eyes in it. I think I manage to don't break anything else. If it gets thumbs up so far I'll right more tests for it.

fskreuz:
Don't worry. There's always eyes in here, esp. lurkers (me included 😄 ). It's actually recommended [to send in half-done PRs](https://github.com/ractivejs/ractive/blob/dev/.github/CONTRIBUTING.md#pull-requests) instead of worrying too much that it won't be accepted.

> If in doubt, submit the PR. A PR that needs tweaking is infinitely more valuable than a request that wasn't made because you were worrying about meeting these requirements.

src/view/items/element/specials/Input.js
```diff
@@ -5,4 +5,16 @@ export default class Input extends Element {
 		super.render( target, occupants );
 		this.node.defaultValue = this.node.value;
 	}
+	compare ( value, attrValue ) {
```
evs-chris: There's a bit of whitespace mismatch from the style guide here. It always takes me a bit to switch into Ractive style mode :smile:

evs-chris:
Looking good! Throw in some tests on radio and check name bindings and maybe a few more complex select scenarios specifically around what happens if the comparator matches but the items aren't `===` and I'd say it's a good candidate for merge in 0.8. Nice first PR :smile:

test/browser-tests/select.js
```diff
@@ -4,6 +4,52 @@ import { initModule } from './test-config';
...
+	});
+
+	test( 'Use as a function to compare', t => {
+		const ractive = new Ractive({
+			el: fixture,
+			comparatorFn: function(option, value) {
+				return option.id == value.id;
+			},
+			template: `
+				<select id="select" value-comparator="{{@ractive.comparatorFn}}" value="{{selected}}">
```
evs-chris: Oh yeah, `@ractive` in now `@this` in edge.

src/view/items/element/binding/CheckboxNameBinding.js
```diff
@@ -127,4 +127,24 @@ export default class CheckboxNameBinding extends Binding {
 		node.removeEventListener( 'change', handleDomEvent, false );
 		node.removeEventListener( 'click', handleDomEvent, false );
 	}
+
+	arrayContains ( selectValue, optionValue ) {
+		let i = selectValue.length;
+		const compare = this.element.compare;
```
fskreuz: Not sure where `compare` is used here.

src/view/items/element/binding/CheckboxNameBinding.js
```diff
@@ -127,4 +127,24 @@ export default class CheckboxNameBinding extends Binding {
...
+		while ( i-- ) {
+			if ( this.element.compare( optionValue, selectValue[i] ) ) return true;
+		}
+		return false;
+	}
+
+	removeFromArray( array, item ) {
+		if (!array) return;
+		let i = array.length;
+		const compare = this.element.compare;
```
fskreuz: Not sure where `compare` is used here.

src/view/items/element/specials/Input.js
```diff
@@ -5,4 +5,16 @@ export default class Input extends Element {
...
+		if (comparator) {
+			if (typeof comparator === 'function') {
+				return comparator(value, attrValue);
+			}
+			if (value && attrValue) {
+				return value[comparator] == attrValue[comparator];
+			}
+		}
+		return value == attrValue;
+	}
```
fskreuz: Any particular reason why the `==`?

Also, just throwing this in, in case you'd reconsider.

```js
return typeof comparator === 'function' ? comparator(value, attrValue)
                   : value && attrValue ? value[comparator] == attrValue[comparator]
                                        : value == attrValue
```

src/view/items/element/specials/Select.js
```diff
@@ -65,8 +58,15 @@ export default class Select extends Element {
...
+				let shouldSelect = false;
 				const optionValue = o._ractive ? o._ractive.value : o.value;
-				const shouldSelect = isMultiple ? valueContains( selectValue, optionValue ) : selectValue == optionValue;
+				const isMultiple = this.getAttribute( 'multiple' );
+
+				if (isMultiple) {
+					shouldSelect = this.valueContains ( selectValue, optionValue );
+				} else {
+					shouldSelect = this.compare(optionValue, selectValue);
+				}
```
fskreuz:
```js
let shouldSelect = isMultiple ? this.valueContains( selectValue, optionValue )
                              : this.compare(optionValue, selectValue);
```

Also, the `let shouldSelect = false;` up top might be unnecessary. It's initial value will always be overridden.

src/view/items/element/specials/Select.js
```diff
@@ -88,7 +88,24 @@ export default class Select extends Element {
...
+		if (comparator) {
+			if (typeof comparator === 'function') {
+				return comparator(selectValue, optionValue);
+			}
+			if (selectValue && optionValue) {
+				return selectValue[comparator] == optionValue[comparator];
+			}
+		}
+		return selectValue == optionValue;
+	}
```
fskreuz: Same as the comment in `Input.js`'s comparator.

src/view/items/element/binding/CheckboxNameBinding.js
```diff
@@ -9,7 +9,8 @@ const push = [].push;
 function getValue() {
 	const all = this.bindings.filter(b => b.node && b.node.checked).map(b => b.element.getAttribute( 'value' ));
 	let res = [];
-	all.forEach(v => { if ( !arrayContains( res, v ) ) res.push( v ); });
+	const binding = this.bindings[0];
+	all.forEach(v => { if ( !binding.arrayContains( res, v ) ) res.push( v ); });
```
fskreuz: This is really a good use case of `Set`. Too bad we can't use it.

guilhermeaiolfi:
That test that fail seems more related to the event expressions PR, right?

I don't know why... everything is green running locally.

src/view/items/element/specials/Input.js
JonDum: That is certainly more succinct, but much harder to interpret at a glance.

PR #2558, opened 2016-05-31
Fix #2551

### LINKED ISSUE #2551: Progressive enhancement fails to remove redundant class names
**Description of the bug:**

The "enhancement" phase fails to remove class names that shouldn't be in the final dom.

```html
<div class="someCls someCls--variant">foo</div>
```

Based on the example I have provided, the html above should have turned into the one below after ractive had done its job:

```html
<div class="someCls">foo</div>
```

I suspect there's something wrong in function [`updateClassName`](https://github.com/ractivejs/ractive/blob/96f593b59316029485d39293115425f2ae305294/src/view/items/element/attribute/getUpdateDelegate.js#L222).

Here is the transpiled code (Debugging the example I've given, the comments at the end of the lines are the runtime values):

```js
function updateClassName () {
    var value = readClass( safeToStringValue( this.getValue() ) ); // value = ["someCls"]
    var attr = readClass( this.node.className ); // attr = ["someCls", "someCls--variant"]
    var prev = this.previous || []; // prev = []

    var i = 0;
    while ( i < value.length ) {
        if ( !~attr.indexOf( value[i] ) ) attr.push( value[i] );
        i++;
    }

    // remove now-missing classes
    i = prev.length;
    while ( i-- ) {
        if ( !~value.indexOf( prev[i] ) ) {
            var idx = attr.indexOf( prev[i] );
            if ( ~idx ) attr.splice( idx, 1 );
        }
    }

    this.node.className = attr.join( ' ' );

    this.previous = value;
}
```

Here, we're making sure all the elements in `value` gets pushed into `attr`, but we're not checking for the opposite case: "Remove elements from `attr` that are not present in `value`." Am I missing something?

***

And one last thing:

```js
this.node.className = attr.join( ' ' );
```

We're always updating the value of `className` attribute even if it is already the same (fires attribute modification events in Chrome Dev Tools). Shouldn't we do something like this instead?:

```js
var newValue = attr.join(' ');

if (newValue !== this.node.className) {
    this.node.className = newValue;
}
```

**Versions affected:**

v0.8.0-edge

**Platforms affected:**

All

**Reproduction:**

http://jsfiddle.net/monoblaine/jtvbdrop/2/

**Code:**

```html
<div id="foo"><div class="someCls someCls--variant">foo</div></div>

<script type="text/template" id="template"><div class="someCls">foo</div></script>
```

```js
setTimeout(function () {
    new Ractive({
        el: 'foo',
        template: '#template',
        enhance: true
    });
}, 200);
```

monoblaine:
This pull request fixes #2551. As I said [here](https://github.com/ractivejs/ractive/issues/2551#issuecomment-222802281), using a copy of `attr` instead of an empty array ensures all redundant classes are removed.

#2551

**Is breaking:**
It shouldn't break anything.

### DIFF
--- a/src/view/items/element/attribute/getUpdateDelegate.js
+++ b/src/view/items/element/attribute/getUpdateDelegate.js
@@ -221,7 +221,7 @@ function updateInlineStyle () {
 function updateClassName () {
 	const value = readClass( safeToStringValue( this.getValue() ) );
 	const attr = readClass( this.node.className );
-	const prev = this.previous || [];
+	const prev = this.previous || attr.slice( 0 );
 
 	let i = 0;
 	while ( i < value.length ) {
--- a/src/view/items/element/attribute/getUpdateDelegate.js
+++ b/src/view/items/element/attribute/getUpdateDelegate.js
@@ -238,7 +238,11 @@ function updateClassName () {
 		}
 	}
 
-	this.node.className = attr.join( ' ' );
+	const className = attr.join( ' ' );
+
+	if ( className !== this.node.className ) {
+		this.node.className = className;
+	}
 
 	this.previous = value;
 }
--- a/test/browser-tests/render/enhance.js
+++ b/test/browser-tests/render/enhance.js
@@ -128,6 +128,20 @@ export default function() {
 		t.ok( !button.disabled );
 	});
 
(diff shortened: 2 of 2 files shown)


evs-chris:
:+1:

Issue #2513, opened 2016-04-14
Breaking change of adaptor behaviour in edge


skeptic35:
I noticed a change in adaptor behaviour that makes it impossible for me to switch to edge.
Provided I have a Backbone model as ```model``` property of my ractive ```data``` and I use the Backbone adaptor. Then I do ```this.get("model")```.
In 0.7.3 that would return the original, unwrapped, unadapted Backbone model, complete with it's own ```get``` and ```set``` methods and what have you.
In edge I only get the model's attributes.
This has rather catastrophic effects on my application, especially in cases where the model is passed on to components, which then receive a POJO instead of a Backbone model.

Question is: Is that a problem of the adaptor or a problem of Ractive or - heaven forbid - an intentional breaking change?

fskreuz:
We're aware of this one. See discussion about adaptors https://github.com/ractivejs/ractive/issues/2512 :smile:

skeptic35:
Sorry to be bothering you again (and that after almost two months have passed...) but I don't quite see how the adaptor discussion is an answer to my question or how this issue is a duplicate to ... well I don't know what it's supposed to be a duplicate of.
Might very well just be me being to dumb but anyway...
Thing is: I'd really love to use Ractive edge and I still can't, because the Backbone adaptor is breaking my Backbone models. And I fail to see another ticket adressing that issue and I also fail to see a solution for it.

skeptic35:
Just downloaded the latest edge version and I have the very same problem again: Wrapped models that are passed down to components are "destroyed", i. e. the component receives a POJO instead of a Backbone Model.

evs-chris:
@skeptic35 oops, I missed a test case, so it slipped back in. It should be good now.

skeptic35:
Yup - good now :)

PR #2561, opened 2016-06-08
if a value is wrapped, return the wrapped value from ractive.get - 2513

evs-chris:
This adds a check for a wrapper in `Ractive.prototype.get`, and if it finds one, it returns the wrapped value whilst still respecting capture. This feels a little dirty, but it can't be done in `Model.get` because that would break the rest of the world. Should I add a new `get` method meant for external use to `Model` instead?

#2513 

**Is breaking:**
Quite the opposite.

**Reviewers:**
@martypdx, please

### DIFF
--- a/src/Ractive/prototype/get.js
+++ b/src/Ractive/prototype/get.js
@@ -1,5 +1,6 @@
 import { splitKeypath } from '../../shared/keypaths';
 import resolveReference from '../../view/resolvers/resolveReference';
+import { capture } from '../../global/capture';
 
 export default function Ractive$get ( keypath ) {
 	if ( !keypath ) return this.viewmodel.get( true );
@@ -22,5 +23,9 @@ export default function Ractive$get ( keypath ) {
 	}
 
 	model = this.viewmodel.joinAll( keys );
-	return model.get( true );
+	// if the model is wrapped, the wrapped value should be returned
+	if ( model.wrapper ) {
+		capture( model );
+		return model.wrapper.value;
+	} else return model.get( true );
 }
--- a/src/Ractive/prototype/get.js
+++ b/src/Ractive/prototype/get.js
@@ -1,6 +1,5 @@
 import { splitKeypath } from '../../shared/keypaths';
 import resolveReference from '../../view/resolvers/resolveReference';
-import { capture } from '../../global/capture';
 
 export default function Ractive$get ( keypath ) {
 	if ( !keypath ) return this.viewmodel.get( true );
@@ -23,5 +22,5 @@ export default function Ractive$get ( keypath ) {
 	}
 
 	model = this.viewmodel.joinAll( keys );
-	return model.get( true, true );
+	return model.get( true );
 }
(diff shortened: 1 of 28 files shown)


skeptic35:
@evs-chris Firstly, you are now officially my hero :)
Secondly
>  Should I add a new get method meant for external use to Model instead?

If I understood that correctly, that wouldn't help in situations where I want to pass the Model on to a component like so:
```<my-component model=[[model]]/>```

**Comment by evs-chris** [comment]
@skeptic35, in this case, Model is actually an internal ractive class that represents a piece of data. Just an implementation detail and not anything that would affect API :smile:

**Comment by skeptic35** [comment]
Ok - I'll take your word on that ;)
Applied the patch to my Ractive source and it seems to work quite well. Now I can turn to the other errors I get ;)

**Comment by skeptic35** [comment]
Sorry, one more thing: Passing the model on to a function within a template. E. g.:
```{{doSomething(model)}}```
In 0.7.3 ```doSomething``` received the Backbone-Model, in 0.8 it receives the wrapped model (i. e. only the attributes). Also in 0.7.3 I could actually do something like this in my template:
```{{model.get("some_attribute")}}```
That also doesn't work anymore in 0.8
Same problem as the original one?

**Comment by evs-chris** [comment]
Ah, so expressions should get unwrapped values too. I haven't really used adaptors much, and it seems our test suite is a little lacking there. Out of curiosity though, why not `{{model.some_attribute}}`?

**Comment by martypdx** [comment]
@skeptic35, @evs-chris I think prior to 0.7 they used wrapped values, but I'm not 100% sure. It's an area where there's ambiguity in what is the "right thing" to do.

Consider these three uses: (see https://jsfiddle.net/martypdx/tbumr4st/):

```html
{{ model }} 
{{ model.toUpperCase() }}  //if adapted model was a string, you might want to do this
{{ method(model) }} //but maybe you want to pass the unadapted source to a method
```

evs-chris:
I'm about to update this to have `Ractive.get` and expressions call `Model.get` and receive an unwrapped value. Non-expressions and chain-related `get`s will still get the adapted value. Does that sound appropriate?

martypdx:
@evs-chris I just about to push almost identical changes :)

src/Ractive/prototype/get.js
```diff
@@ -1,5 +1,6 @@
 import { splitKeypath } from '../../shared/keypaths';
 import resolveReference from '../../view/resolvers/resolveReference';
+import { capture } from '../../global/capture';
```
martypdx: this can be removed now

martypdx:
@evs-chris it doesn't appear in the current code there's a time we don't do `true, true`. Did I miss a usage? If that's true, we could just have a single param that means both things...

evs-chris:
That's a fair point... if it's being captured, it's external/user-facing and should thus be unwrapped. Fixed!

skeptic35:
@evs-chris 
> Out of curiosity though, why not ```{{model.some_attribute}}?```

```{{model.get("some_attribute")}}``` was simply a bad example. Actual point is to be able to call methods on the model. But I agree with @martypdx: There is amibguity as to what is the right thing to do. I could well live without being able to call methods of the wrapped model in the template, As long as ```{{doSomething(model)``` receives the unwrapped model that is ;)

Issue #2550, opened 2016-05-24
Computed array's dependent fragments not updating (edge, ractive-datatable)


JonDum:
**Description of the bug:**

Ractive is not updating a computed array's dependent DOM fragments even though the computed is getting updated in the viewmodel. Check out the edge demo of ractive-datatable [here](http://jondum.github.io/ractive-datatable/demo/edge.html). Generate some data, then change pages. Nothing happens except for when reaching the last page. You can verify that the `rows` computation is getting recalculated correctly by placing a breakpoint or running `main.findAllComponents()[0].get('rows')`. However, the DOM is not being updated to reflect the new values of the `rows` computation.

**Versions affected:**

0.8-edge

**Platforms affected:**

Chrome 50

**Reproduction:**

http://jondum.github.io/ractive-datatable/demo/edge.html

**Code:**
```html
<datatable data='{{data}}'></datatable>
```
```html
{{#rows:r}}
    <tr {{#if selectionMode == 'row' && _selection.indexOf(index) !== -1 }}class='selected'{{/if}} on-
        ...
    </tr>
{{/}}
```
```js
rows: function() {
    var self = this;

    var page = self.get('page') - 1;
    var _data = self.get('_data');
    var perpage = self.get('perpage');
    var total = self.get('total');

    // the original data, unfiltered
    var data = self.get('data');

    return _data.slice(page * perpage, Math.min(page * perpage + perpage, total));
},
```

evs-chris:
Interestingly, if you go to a page that changes the length of the displayed result set, then it rerenders as expected.

JonDum:
Maybe something to do with equality checking the objects? Since it has the same length it goes to check each item in the array but it's not passing the equality check? The `rows` array looks like this:

```js
[ {
    item: {
        ...
    },
    index: 30
}, ...]
```

I verified that after changing a page `rows[0] !== rows[0]` though. Hmm.

evs-chris:
I've tracked this down to an issue with the content of the column partial not being flagged as dirty because it doesn't exactly depend on any of the values that are changing. I'm thinking it's a bug with how the expression proxies notify and/or are notified about changes.

evs-chris:
Just wanted to note that the events aren't working because the pre-compiled template is not compatible with expression events.

JonDum:
Tested it out on edge and works with raw template string 👍 .

So I guess this should only be a problem now until 0.8 drops to NPM (if you use the prebuilt version).

[...]

PR #2564, opened 2016-06-09
Make sure reference expressions capture - #2550

evs-chris:
In certain interesting scenarios, a reference expression proxy can become a model in another computation, notably with the reference expression is used as context. Turns out reference expressions where not paying attention to capture flags, which makes for sadness when dependencies in the context update. This just adds the capture flag on `ReferenceExpressionProxy.get` to pass along.

#2550

**Is breaking:**
Nerp.

**Reviewers**
@JonDum, if you have a bit of time, could you try this out? I'mma merge it once it goes green, but I want to be sure it handles your more complex case too.


JonDum:
Good job! LGTM

Issue #2560, opened 2016-06-04
Break attributes xmlns:xlink, xlink:href in browser


kachurun:
**Description of the bug:**
Hello, i'm used ractive for display list of SVG icons with `use` tag, like this:

```
<svg width="40" height="40" style="vertical-align:middle; fill:{{item.ico.color}};" viewBox="0 0 52 52">
        <use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="{{item.ico.url}}#icon"></use>
</svg>
```  

then svg4everybody polyfill converts `use` tag to SVG content, but now i've got error with it because  Ractivejs compile my template to this:
```
<svg width="40" height="40" style="vertical-align:middle; fill:#242424;" viewBox="0 0 52 52">
        <use xlink="http://www.w3.org/1999/xlink" href="/_s/images/svg/line/1.svg#icon"></use>
</svg>
```
its turned attributes:
`xmlns:xlink`-> `xlink`  
`xlink:href` -> `href`  

p.s. I have no problem with compile template with node.js (or browser with use .toHTML() method)

**Versions affected:**
0.8-edge  

**Platforms affected:**
Any browser, not node.js or .toHTML() method  

**Reproduction:**
http://jsfiddle.net/fs07b7zL/

evs-chris:
I'm not seeing quite this issue on Chrome 52. Locally, I get an erroneous xlink attribute, but I'm not seeing the href without the xlink namespace on it. I have a fix for the extra attribute, but I'm not sure why the fiddle should behave differently than a local test. Changing the name of the namespace to foo e.g. `xmlns:foo="..."` and using `foo:href` seems to work around the issue in the fiddle.

I have very little experience with namespaces, but I believe ractive will actually supply xlink for you, so you don't have to include the xmlns:xlink bit. But again, I have no idea why it's not behaving in the fiddle.

evs-chris:
I _think_ this is ok, or at least I've successfully used xlinked use tags in an inline svg in ractive. Please re-open if this is still an issue.

PR #2568, opened 2016-06-10
Don't export attributes that are namespace definitions

evs-chris:
Currently if you drop an xmlns:whatever in an element, the whatever portion will also show up as a regular attribute. This checks for the xmlns namespace and skips popping those out into the dom.

I've no idea why behavior would be different in a fiddle than run locally. Maybe something to do with jsfiddle declarations?

sort #2560

**Is breaking:**
Shouldn't be.

**Reviewers:**
Anyone familiar with xmlns?


evs-chris:
It seems phantom is not a fan. Does anyone more familiar with the subject matter know if this is one of those things that different browsers do completely differently? The test passes in Chrome.

PR #2573, opened 2016-06-14
Check check for binding updates before the observer phase to unlock the attr - #2427

evs-chris:
This is a weird one that is caused by behavior that should probably be dropped since laziness is available to help with input validation. At any rate, this checks the see if a binding's model changes to something not set immediately after it is set and before any observers get a chance to fire. If the value is not the same, then the attribute is unlocked so that the binding doesn't get stuck on the old value.

One side effect is that as soon as you press a key that triggers a re-sort on the underlying list, your current box is shuffled off and the new one (actually the same one with a new value, but you get the idea) is sitting there with focus. I think that's better behavior than getting one value stuck in the old box while the new one appears correctly in the new destination. Again, `lazy` can resolve that entirely. 

A further potential side-effect would be that binding to a computation that morphs inputs would get cursor resets while typing. And again, I believe `lazy` is the solution.

Thoughts?

#2427

**Is breaking:**
Maybe? It wasn't really well-defined behavior to begin with...

**Reviewers:**
@martypdx @fskreuz please

### DIFF
--- a/src/view/items/element/binding/Binding.js
+++ b/src/view/items/element/binding/Binding.js
@@ -74,7 +74,11 @@ export default class Binding {
 		runloop.start( this.root );
 		this.attribute.locked = true;
 		this.model.set( this.getValue() );
-		runloop.scheduleTask( () => this.attribute.locked = false );
+
+		// if the value changes before observers fire, unlock to be updatable cause something weird and potentially freezy is up
+		if ( this.model.get() !== this.getValue() ) this.attribute.locked = false;
+		else runloop.scheduleTask( () => this.attribute.locked = false );
+
 		runloop.end();
 	}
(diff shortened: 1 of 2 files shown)


heavyk:
noob question: why don't you save the result of `this.getValue()` into a variable instead of doing it twice? can that value change?

evs-chris:
In the midst of mangling things to try to get the various tests involved to pass, the versions with cached values just didn't survive :smile:. `getValue` is basically just a front for `node.value`, so it's not exactly expensive or likely to change. ¯\_(ツ)_/¯

heavyk:
even though I don't understand the locking logic very well (and all of its implications), I'd still say LGTM :)

PR #2575, opened 2016-06-14
Mark mappings along with root viewmodels - fixes #2574

### LINKED ISSUE #2574: Calling update on component does not update mapping
When calling update on a component where the data has changed on a nested property, that changed data is not updated and thus the view does not update.

Affects 0.7.3

See jsFiddle to reproduce: [http://jsfiddle.net/jpas/xcumonru/15/](url)
Click run, then click redraw row and look at the dev console. You will see that both key-paths (which should point to the same object) in fact point to different objects.

Thanks.

evs-chris:
This adds a step to root model mark that iterates the mappings and marks them as well.

#2574

**Is breaking:**
Don't think so.

**Reviewers:**
Please.

### DIFF
--- a/src/model/RootModel.js
+++ b/src/model/RootModel.js
@@ -131,6 +131,17 @@ export default class RootModel extends Model {
 		return remapped;
 	}
 
+	mark () {
+		const keys = Object.keys( this.mappings );
+
+		let i = keys.length;
+		while ( i-- ) {
+			this.mappings[ keys[i] ].mark();
+		}
+
+		super.mark();
+	}
+
 	resetMappings () {
 		for ( let k in this.mappings ) {
 			this.mappings[k].unregister( this );
--- a/src/model/RootModel.js
+++ b/src/model/RootModel.js
@@ -132,13 +132,7 @@ export default class RootModel extends Model {
 	}
 
 	mark () {
-		const keys = Object.keys( this.mappings );
-
-		let i = keys.length;
-		while ( i-- ) {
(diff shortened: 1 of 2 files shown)


martypdx:
Does that mark call happen on any upstream change from actual children? My concern would be that every change in the component data now causes cascading changes up to all mapped data.

Perhaps it should only be applied in the case of `ractive.update()` with no keypath specified.

evs-chris:
`mark` only goes downstream at present. Would it be safer to go ahead and just move this to the update method in case that changes in the future?

martypdx:
Ah yes, it goes upstream [here](https://github.com/ractivejs/ractive/blob/dev/src/model/Model.js#L419-L427). I think it's fine, just wanted to check that we weren't spamming the mappings hierarchy. :)

src/model/RootModel.js
```diff
@@ -131,6 +131,17 @@ export default class RootModel extends Model {
 		return remapped;
 	}
 
+	mark () {
+		const keys = Object.keys( this.mappings );
+
+		let i = keys.length;
+		while ( i-- ) {
+			this.mappings[ keys[i] ].mark();
+		}
```
fskreuz: If order doesn't matter
```js
Object.keys( this.mappings ).forEach( key => this.mappings[ key ].mark() );
```

evs-chris: Fair enough :smile:. For some reason I default to performance mode, which is definitely premature optimization here.

Issue #2463, opened 2016-03-23
Whacky and broken transitions. Elements not being removed from dom, added too soon.


JonDum:
Hover over and out of each box quickly and watch the magic happen. :dancer: 

http://jsfiddle.net/d3y75cup/

```html
  <div class='root' on-hover='hoverA'>
      Hover over me (nothing fancy)
      
      <div class='items'>
        {{#itemsA}}
            <div class='item' intro-outro='fly'>
                {{.}}
            </div>
        {{/}}
       </div>
  </div>
  
  <div class='root' on-hover='hoverB'>
      Hover over me (fancy)
      
      <div class='items'>
        {{#itemsB:i}}
            <div class='item'
                 intro='fly:{delay: {{ i*100 }}, x: 0, y: {{i*50}} }'
                 outro='fly:{delay: {{ i*100 }} x: 0, y: -50}'>
                {{.}}
            </div>
        {{/}}
       </div>
  </div>
```
```js
var ractive = new Ractive({
  el: 'main',
  template: '#template',
  computed: {
  	itemsA() {
    	var open = this.get('hoveringA');
        if(open)
            return this.get('things');
    },
  	itemsB() {
    	var open = this.get('hoveringB');
        if(open)
            return this.get('things');
    }
  },
  data: {
      things: ["one", "two", "three", "four", "five"],
  },
  oninit() {
  
  	this.on('hoverA', event => {
       this.set("hoveringA", event.hover)
    });
    
  	this.on('hoverB', event => {
       this.set("hoveringB", event.hover)
    });
  }
});
```

evs-chris:
...oh my... :laughing: 

On the plus side, I can only seem to get one extra set of elements stuck in the DOM. Looks like perhaps the cancelled transitions aren't getting their nodes detached in all circumstances.

martypdx:
I found some transition bugs that I fixed as part of #2167, basically it wasn't working correctly when same property would be transitioned prior to previous transition being completed (sounds like this case).

I'd like to get the PR merged anyway, so maybe that would be good first step.

Rich-Harris:
Now that #2167 has been merged I can't get DOM nodes to stick around past their use-by date... does that qualify as 'fixed' in this case? :grin:

evs-chris:
I'm gonna go with "fixed as long as the browser doesn't drop a transition even anywhere?" :smile:

JonDum:
@evs-chris I still see the same behavior :(

[...]

Issue #2435, opened 2016-03-12
Outro animations in components don't apply on the correct element


ADantes:
Rendered directly: http://jsfiddle.net/0dn4qqw1/1/
This does exactly what I want.  Click on a row and you can smoothly see it collapse.

Rendered as a component in another view: http://jsfiddle.net/0dn4qqw1/2/
This does not.  You can see Ractive reorder the DOM, placing the row it is about to delete at the bottom.  Only after that does it remove the row. Since that row is at the bottom of the list now, the animation is visually disconnected from the action the user performed and it overall looks pretty jumpy.

evs-chris:
Fixed in edge, though I'm not entirely sure when :smile:

Issue #2415, opened 2016-02-28
Component intro transitions regression on Edge


ADantes:
Intro transitions are correctly animated on initial load.  However, if a component is rendered later, its intro transitions are not animated.

0.7.3: http://jsfiddle.net/29fLzj6g/1/
Edge: http://jsfiddle.net/29fLzj6g/
* Click "Tab One" and observe the differences between the two versions

evs-chris:
Just wanted to note that I'm looking into it and, potentially, a way to resolve the issues with transitions holding up non-transitioning elements.

evs-chris:
Fixed :smile:

Issue #1015, opened 2014-07-20
Transitions never complete if nodes are removed from DOM prematurely


Rich-Harris:
This is a real edge case, but one I've encountered while rebuilding learn.ractivejs.org. When nodes that are undergoing an outro transition are removed from the DOM by some external force (e.g. an ancestor node has `node.innerHTML = ''` done to it), the CSS `transitionend` event is never fired. As a result, the outro is never considered to have completed, and other nodes fail to detach from the DOM.

Example: http://jsfiddle.net/rich_harris/j3KJ7/

Madgvox:
I'm closing this since I feel like this will get solved when ~~browsers have decent mutation apis~~ hell freezes over.

PR #2578, opened 2016-06-15
Transition regression fixes

evs-chris:
This _should_ fix three outstanding transition issues, one of which was fixed and then broken again in a slightly different way by the recent conditional merge. I say _should_ because transition-related issues are, shall we say, challenging to test.

1. #2463 - this is the one that was fixed and re-broken. This one was caused by some binding happening during transition construction, a time at which parts of the virtual dom are not in place to do resolution. Part II of this issue is dom sticking around past its due date. I did manage to kinda reproduce two or three times it (a tiny percentage) by shaking the crap out of the mouse over the hover trigger. It's not particularly consistent and I couldn't reproduce it with a test rapidly triggering transitions. I think this may be more a bug with `animateStyle` not always resolving its promise, but I'm not positive. I think a longer term solution to would be to set an adjustable timeout on the transition and force completion after that timeout. I think the transition completion dom event may be most of the issue.

2. #2415 - this is one is caused by a component rendering into a document fragment and having its transitions kicked off before it's actually attached to the dom. To work around that, this makes the transition managers in the batch stack wait until the root batch is done before kicking off transitions. Since non-transitioning nodes need to be removed before deferred observers, this adds another transition manager stage `ready` which is fired where `start` was before. I couldn't come up with a decent/reliable test for this.

3. #2435 - I _think_ this one is caused by removed fragments being dropped immediately from the top of the list. Again, I couldn't come up with a decent/reliable test for this. This sort of reverts some performance optimization in `RepeatedFragment.updatePostShuffle` so that exiting fragments can be more easily re-attached as near their old position as possible and transition out from there instead of popping up to the top of the section. There may be something component-related to this issue too, but I'd like to see if this addresses the issue before I add yet more to the runloop. As for the de-optimization, I have another avenue that should more than make up for the 5-10% slowdown removing the optimization introduced (looking like a 70% improvement, but there are still many failing tests).

I love that ractive has pretty great transition support, but man are they a pita to debug.

Definitely #2415, at least most of #2463, and hopefully #2435 

**Is breaking:**
Don't think so.

**Reviewers:**
Yes, please.

### DIFF
--- a/src/Ractive/prototype/transition.js
+++ b/src/Ractive/prototype/transition.js
@@ -30,8 +30,12 @@ export default function Ractive$transition ( name, node, params ) {
 	params = params || {};
 	const owner = node._ractive.proxy;
 	const transition = new Transition({ owner, parentFragment: owner.parentFragment, name, params });
+	transition.bind();
+
 	const promise = runloop.start( this, true );
 	runloop.registerTransition( transition );
 	runloop.end();
+
+	promise.then( () => transition.unbind() );
 	return promise;
 }
--- a/src/global/TransitionManager.js
+++ b/src/global/TransitionManager.js
@@ -46,6 +46,10 @@ export default class TransitionManager {
 		this.children.forEach( _detachNodes );
 	}
 
+	ready () {
+		detachImmediate( this );
+	}
+
 	remove ( transition ) {
 		var list = transition.isIntro ? this.intros : this.outros;
 		removeFromArray( list, transition );
@@ -53,7 +57,7 @@ export default class TransitionManager {
 	}
 
 	start () {
-		detachImmediate( this );
+		this.children.forEach( c => c.start() );
 		this.intros.concat( this.outros ).forEach( t => t.start() );
 		this.ready = true;
 		check( this );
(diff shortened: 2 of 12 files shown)


sabob:
> may be more a bug with animateStyle not always resolving its promise

Not sure if related to this issue but there is a known animation issue where promises don't always resolve. This was discussed here:
https://github.com/ractivejs/ractive/issues/1015

No practical solution was found though.

evs-chris:
@sabob interesting. I think a defaulted but configurable timeout would at least patch around the issue. Still, it's pretty hard to nail down.

I think I've got the remaining component transition issues handled. Basically, components weren't waiting on transitions to complete before unrendering, so this now adds them to the detach queue. I had to adjust the unrender process a bit to acommodate, but I think I got it squared away and still _mostly_ respecting hook order. If there are components with transitions, they are not going to fire hooks in exactly the right order since they will wait on their transitions before firing.

evs-chris:
It seems that that last commit was not actually accomplishing anything, so I've stashed it locally and rolled it back here. I'm going to merge this and see how it does on the various issues, since it also fixes some bugs introduced recently by big merges in edge too.

Issue #2556, opened 2016-05-31
Returning "false" in an event handler does not stop the event if "@this.fire" is used


monoblaine:
**Description of the bug:**

Returning "false" in an event handler does not stop the event if `@this.fire` is used.

The old school method works:

```html
<a on-click="foo"></a>
```

But the new one does not work:

```html
<a on-click="@this.fire('foo')"></a>
```

I tried to debug it, sort of found the reason, but I don't know how to fix it. Here is what I've found ([Fiddle here](http://jsfiddle.net/monoblaine/hf68w350/1/)):

1. When we click on the link, in the DOM event handler, `directive.fire` is called [here](https://github.com/ractivejs/ractive/blob/6ff7104a6bbc47f7a918a2e890f36a51534f5f4e/src/view/items/element/ElementEvents.js#L25).
2. In method [`fire`](https://github.com/ractivejs/ractive/blob/eb6c7d5670878bf483e50383556451a636060ff6/src/view/items/shared/EventDirective.js#L108) of `EventDirective`, [`if ( this.fn )`](https://github.com/ractivejs/ractive/blob/eb6c7d5670878bf483e50383556451a636060ff6/src/view/items/shared/EventDirective.js#L120) is true because we're not using the old school method.
3. [Here](https://github.com/ractivejs/ractive/blob/eb6c7d5670878bf483e50383556451a636060ff6/src/view/items/shared/EventDirective.js#L150), our event handler is kinda called:
    
    ```js
    var result = this.fn.apply( ractive, values ).pop();
    ```
4. Then this (auto-generated):
    
    ```js
    (function(
    /**/) {
    return function (_0){return([_0.fire("foo")]);};
    })
    ```
5. Then [this](https://github.com/ractivejs/ractive/blob/6ff7104a6bbc47f7a918a2e890f36a51534f5f4e/src/Ractive/prototype/fire.js#L4):
    
    ```js
    fireEvent( this, eventName, { args: args });
    ```
6. In [`fireEvent`](https://github.com/ractivejs/ractive/blob/6ff7104a6bbc47f7a918a2e890f36a51534f5f4e/src/events/fireEvent.js#L6), `if ( !options.event )` is true so a weird object named "event" is added to "options" object.
7. [`fireEventAs` is called](https://github.com/ractivejs/ractive/blob/6ff7104a6bbc47f7a918a2e890f36a51534f5f4e/src/events/fireEvent.js#L21).
[...]

PR #2580, opened 2016-06-15
Return bubble status from fire for better default fire expression behavior - #2556

evs-chris:
This makes `fire` return its bubble/cancel status so that `<button on-click="@this.fire('someEvent')">` behaves a bit more intuitively. The current workaround is to just add a false expression to the end e.g. `@this.fire('someEvent'), false`. So after this, if you want the event to bubble regardless of the result of the fire, you'd just have to add a true expression e.g. `@this.fire('someEvent'), true`.

#2556 

**Is breaking:**
No

**Reviewers:**
@martypdx


martypdx:
LGTM!

PR #2582, opened 2016-06-16
Make reference expression proxies lazy - #2461

evs-chris:
This updates `ReferenceExpressionProxy` such that instead of triggering an immediate get on all of its models, it waits until the value is actually needed by setting a dirty flag and deferring the actual pull to its own `get`.

#2461

**Is breaking:**
No


martypdx:
nice!

PR #2584, opened 2016-06-16
Pop to the first actual context when resolving rootpath - fixes #2026

### LINKED ISSUE #2026: recursive components keypaths empty
Hi,

I took an [example](http://jsfiddle.net/rich_harris/LDk5N/) that @Rich-Harris put together for [recursive components](https://groups.google.com/forum/#!topic/ractive-js/jGtCt3QDhqU) and began to adapt.

[Here's mine](http://jsfiddle.net/felixhummel/LDk5N/15/)

Changes I made:
- Replace `a` with `button`
- Add `on-click` handler and `*.debug` listener

Steps Taken:
- click "README.txt"

Expected:
- `keypath` is `"root.files.2.files.0.filename"`

Actual:
- `keypath` is `""`

Notes:
`ractive.get('root.files.2.files.0.filename')` returns `README.txt` as expected.
The same goes for all other buttons. Keypath is always "".

evs-chris:
This adds a step to `@rootpath` resolution that pops to the first non-empty (keypath-wise) context to get the keypath. Here's the example:

```handlebars
{{#with foo.bar.baz.bat}}<cmp>{{@rootpath}}</cmp>{{/with}}
```

Without this, the result is an empty string, and with it, the result is `foo.bar.baz.bat`. So, with node info helpers on the event object, you can get the root path to the event context with `event.resolve('@rootpath')`.

#2026

**Is breaking:**
Shouldn't be.

### DIFF
--- a/src/view/resolvers/resolveReference.js
+++ b/src/view/resolvers/resolveReference.js
@@ -19,6 +19,11 @@ export default function resolveReference ( fragment, ref ) {
 		return context.getKeypathModel( fragment.ractive );
 	}
 	if ( ref.indexOf( '@rootpath' ) === 0 ) {
+		// check to see if this is an empty component root
+		while ( context.isRoot && context.ractive.component ) {
+			context = context.ractive.component.parentFragment.findContext();
+		}
+
 		const match = keypathExpr.exec( ref );
 		if ( match && match[1] ) {
 			const model = resolveReference( fragment, match[1] );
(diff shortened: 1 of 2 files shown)


heavyk:
awesome!! I think this will allow me to do something in a decorator I previously couldn't: resolve the keypath correctly from inside a loop instead of having to pass the value directly to the decorator. excellent!!

evs-chris:
Sounds like `info.resolve('..')` could help in that case too. In an iteration, that would give you the path to the list being iterated. Any template refs can now be resolved in code with that `resolve` method, so `../../foo` would work as you would expect too.

Issue #1551, opened 2014-12-06
Request for global function to stop/pause animations


sabob:
I'd like to propose a new animation feature to stop/pause the animation, similar to jQuery stop[1](http://api.jquery.com/stop/).

jQuery.stop can be called with a "jumpToEnd" argument which stops the animation but also sets the element's final animation properties.

My use case is that I have a menu that switches views. These views are Ractive instances with a fade in/out animation. 

If the user clicks quickly enough on the menus, the effect is that new views are rendered (fading in) while previous views are still fading out.

Currently with Ractive one can simulate jQuery.fx.off[2](http://api.jquery.com/jquery.fx.off/), by setting a Ractive'a transitionsEnabled to false, although jQuery.fx.off is a global function which makes it easy to control all animations on a page.

Looking at the source code of Ractive's animation it supports CSS transitions and fall back to JS transitions if CSS transitions is not supported.

sabob:
Just tried my hand at adding at patching my local Ractive with a Ractive.prototype.stopTransitions function, but digging into the code I realized there isn't currently a way to access running transitions (in order to abort them).

The TransitionManager isn't for public consumption correct?

Kind regards

Bob

sabob:
Any chance this issue can be reopened? I still have a use case to stop animations to force Ractive into sync mode.

PR #2585, opened 2016-06-17
Wrap animateStyle css transitions in a timeout in case transitionend doesn't fire - #2463

evs-chris:
This adds a timeout of `duration + (delay || 0) + 10` to css transitions in `animateStyle` to force the transition to complete within 10ms of its target. `transitionend` events appear to not be particularly reliable in scenarios where you trigger transitions like a madman.

After aiming the fiddle from #2463 at a build of this branch and about 5 minutes of vigorous mouse wiggling, I'm reasonably confident that this will take care of transition-event-related orphan nodes. I'll probably go ahead and merge this later this afternoon (once it goes green) so that others can get their wiggle on as well to make sure I'm not experiencing the placebo effect.

Side note: I have an idea for a subscription-based software shake weight competitor...

Hopefully #2463

**Is breaking:**
No

### DIFF
--- a/src/view/items/element/transitions/createTransitions.js
+++ b/src/view/items/element/transitions/createTransitions.js
@@ -53,6 +53,7 @@ if ( !isClient ) {
 		setTimeout( () => {
 			let jsTransitionsComplete;
 			let cssTransitionsComplete;
+			let cssTimeout;
 
 			function checkComplete () {
 				if ( jsTransitionsComplete && cssTransitionsComplete ) {
@@ -79,6 +80,8 @@ if ( !isClient ) {
 			style[ TRANSITION_DURATION ] = ( options.duration / 1000 ) + 's';
 
 			function transitionEndHandler ( event ) {
+				clearTimeout( cssTimeout );
+
 				const index = changedProperties.indexOf( camelCase( unprefix( event.propertyName ) ) );
 
 				if ( index !== -1 ) {
@@ -90,6 +93,10 @@ if ( !isClient ) {
 					return;
 				}
 
+				cssTransitionsDone();
+			}
+
+			function cssTransitionsDone () {
 				style[ TRANSITION_PROPERTY ] = previous.property;
 				style[ TRANSITION_TIMING_FUNCTION ] = previous.duration;
 				style[ TRANSITION_DURATION ] = previous.timing;
@@ -101,6 +108,12 @@ if ( !isClient ) {
 			}
 
 			t.node.addEventListener( TRANSITIONEND, transitionEndHandler, false );
+
+			// safety net in case transitionend never fires
+			cssTimeout = setTimeout( () => {
+				changedProperties = [];
+				cssTransitionsDone();
+			}, options.duration + ( options.delay || 0 ) + 10 );
 
 			setTimeout( () => {
 				let i = changedProperties.length;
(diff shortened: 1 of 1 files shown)


evs-chris:
I did another set of wiggle tests before and after this, and the before left stuck elements where the after didn't. @JonDum, of you have a few minutes, can you see if you can get it to break again?

JonDum:
Looks good now! 93682182ec38 seemed to do the trick.

lol

Still looks hilarious, but that's what I would expect. The important part is that it's not leaving any orphans! 👍

evs-chris:
Thanks! 🎉

sabob:
There is a slight issue with this setTimeout when render/teardown ractive instances.

I use ractive as the view in a single-page app, with a menu  for switching between ractive views. When switching I unrender/teardown the previous view and render the new one. This is done using a fade effect.

Here is the fiddle:
http://jsfiddle.net/sabob/q2j22hL5/2/

Quickly click twice on the "Change view" button and notice the counter appears, then disapears as the timeout kicks in. If I remove the "cssTransitionsDone()" from the timeout, it works as expected. 

I think the timeout should also be cleared when the view is unrendered?

evs-chris:
@sabob I've been playing with this a bit, and I'm not certain exactly how things are interacting poorly here. The elements stick around, but the interpolator is not. I have some changes that add better transition cancellation to the unrender, but I'm having a hard time with a test to verify that it solves the problem.

evs-chris:
Turns out I was overthinking it...

@sabob, it seems to be much less frequent, but there is definitely something in the render process that doesn't like to have instances unrendered while they're still transitioning in. Was it also happening some before?

sabob:
@evs-chris still happens after complete queue code. Doesn't happen in 0.7.3 or without the cssTransitionsDone line in the timeout.

In case you cannot produce the error here is a gif showing the issue. Clicking the button twice updates the counter to 2, then the counter disappears.

animate

sabob:
In a future release it would be great if there is a way to stop/start the animations, I've created a issue for this awhile back:

#1551

evs-chris:
Found it. Basically, a new instance was being torn down before it was rendered, and rendering the torn down (and thus unbindable) interim instance was causing the new, correct instance to be torn down. I've added a check for render on a torn down instance. It will now throw out a warning in debug mode and roll on without breaking stuff.

sabob:
Thanks Chris, seems to work now. 

From your explanation that ractive tears down before render, why does that occur? Is it because animations turns things asynchronous?

evs-chris:
Pretty much, yes. It's the combination of transition asynchrony and triggering an additional round of callbacks before the previous round has completed.

sabob:
Perhaps in a future release things could be more robust if render cancels current active intro/outro animations on the ractive instance. This will turn things back into synchronous mode.

On the other hand maybe I'm naive and there are use cases for the current behavior...

evs-chris:
Unfortunately, once it's gone asynchronous there's no coming back, and I think transitions are pretty much required to be async.

In this particular case, you could side-step the issue by keeping the latest promise around to make sure you don't step on another transition by triggering a render during an outro. At that point, the user would experience some lag when rapid-fire clicking, but there's no opportunity to damage rendering.

```js
// next to oldRactive declaration
var promise;
//...

if (oldRactive == null) {
    promise = (promise || Promise.resolve()).then(() => newRactive.render('main'));
} else {
    promise = promise.then(() => oldRactive.teardown().then(() => newRactive.render('main')));
}

oldRactive = newRactive;
```

Though with the teardown and render guards, I think it's pretty safe to do what the fiddle is doing. Unless you're depending on component render hooks to side-effect somewhere. Then your only option is to keep the promises around somewhere to ensure that all instances actually get rendered.

PR #2588, opened 2016-06-19
Make magic mode not break 'this' for the underlying object's getters and setters

zlamma:
 - preserve `this` by using `.call(this, ...)`

When your model has getters/setters that use `this`, applying magic mode to the model will break them.

**Is breaking:**
None 

**Reviewers:**
Any


evs-chris:
Looks good, but can you add a test? Also, there's an extra space before your open paren on line 42 😄

zlamma:
Fixed the space now & forced.
I think I can add a test. Sadly it's a bit slow for me as I'm on a fresh Windows.

zlamma:
Added test, which revealed it still didn't work due to use of arrow functions, fixed + forced.

Ready to go, me thinks.

PR #2591, opened 2016-06-21
First pass at slightly smarter (and much faster) shuffling - RFC

evs-chris:
Shuffling is one of the last things that is somewhat slow in edge, primarily because when an array shuffles, everything downstream from that array re-resolves all of its references, which is a pretty slow process. This attempts to make that process slightly smarter by assuming that there are roughly three states for a shuffling rebind: not affected by the shuffle, shuffled, and unknown. The unknown state is kind of a special case that is also used by unmapping and remapping in components to get updates propagated correctly.

When a shuffle kicks off, it registers with the runloop. Then when a rebind fires, if it already has a model, it checks to with the model to see if there is a known new model. If there is, it uses it. If it is known that the model is not affected by a shuffle, then it stops there. If it is unknown, then the usual re-resolve process takes place.

Since keypaths are used a ton in this, it also adds keypath caching to avoid a pretty good bit of keypath re-calc. Substitution is still done if the path is mapped, but it doesn't have to start by walking up to the root model.

I have all of the tests passing, but I'm not positive that the test suite covers 100% of the corner cases involved in shuffling. I had to add actual rebind handlers in a few places because there were/are a few items that just unbind/bind on rebind.

There are still a few models and model-like things that do not participate in the tryRebind process, but there shouldn't be any reason why everything wouldn't be covered. Notably, I haven't looked at reference expressions and children of those and computations at all yet. At this point, I think everything in the jsweb perf suite is covered, and I'm seeing this take roughly 35% of the time that edge does for removing the first row (out of 1000) and 20% of the time for the last. Splicing the last row is still not quite as fast as 0.7.3, but everything else is considerably faster.

**Fixes**
#2366 - to the extent that it can... edge is already considerably faster than 0.7.3 in most of the perf suite.

**Is breaking:**
Doesn't look like it, but the test suite may be lacking.

**Reviewers:**
Yes please! If you have a little time to look over the implementation or try out this PR, by all means, please do, and feel free to leave some feedback!

### DIFF
--- a/perf/tests/jsweb.js
+++ b/perf/tests/jsweb.js
@@ -14,6 +14,7 @@ var tests = [
 						template:
 `<table>
 	<tr><th>name</th><th>index</th><th>remove</th></tr>
+	{{#if show}}
 	{{#rows:i}}
 	<tr class="{{#selected === .id}}selected{{/}}">
 		<td>{{.id}}</td>
@@ -22,8 +23,9 @@ var tests = [
 		<td><button on-click="remove:{{i}}">remove</button></td>
 	</tr>
 	{{/rows}}
+	{{/if}}
 </table>`,
-						data: { rows: [] }
+						data: { rows: [], show: true }
 					});
 
 					var id = 0;
@@ -53,12 +55,21 @@ var tests = [
 			},
 
 			{
-				name: 'add 100 rows',
+				name: 'add 100 rows (accumulates)',
 				test() {
 					/* global ractive, gen */
-					ractive.push( 'rows', gen( 100 ) );
-				},
-				maxCount: 5
+					const rows = gen( 100 );
+					rows.unshift( 'rows' );
+					ractive.push.apply( ractive, rows );
+				}
+			},
+
+			{
+				name: 'reset to 1000 rows',
+				test() {
+					/* global ractive, gen */
+					ractive.set( 'rows', gen() );
+				}
 			},
 
 			{
@@ -71,6 +82,14 @@ var tests = [
 			},
 
 			{
+				name: 'reset to 1000 rows',
+				test() {
+					/* global ractive, gen */
+					ractive.set( 'rows', gen() );
+				}
+			},
+
+			{
 				name: 'remove first row',
 				test() {
 					/* global ractive */
@@ -80,12 +99,64 @@ var tests = [
 			},
 
 			{
+				name: 'reset to 1000 rows',
+				test() {
+					/* global ractive, gen */
+					ractive.set( 'rows', gen() );
+				}
+			},
+
+			{
 				name: 'remove last row',
 				test() {
 					/* global ractive */
 					ractive.pop( 'rows' );
 				},
 				maxCount: 100
+			},
+
+			{
+				name: 'reset to 1000 rows',
+				test() {
+					/* global ractive, gen */
+					ractive.set( 'rows', gen() );
+				}
+			},
+
+			{
+				name: 'hide rows',
+				test() {
+					/* global ractive */
+					ractive.set( 'show', false );
+				},
+				maxCount: 1
+			},
+
+			{
+				name: 'show rows',
+				test() {
+					/* global ractive */
+					ractive.set( 'show', true );
+				},
+				maxCount: 1
+			},
+
+			{
+				name: 'hide rows again',
+				test() {
+					/* global ractive */
+					ractive.set( 'show', false );
+				},
+				maxCount: 1
+			},
+
+			{
+				name: 'show rows again',
+				test() {
+					/* global ractive */
+					ractive.set( 'show', true );
+				},
+				maxCount: 1
 			},
 
 			{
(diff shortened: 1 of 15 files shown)


JonDum:
Very sexy. I'm gonna try this out with `comp-bind` in the morning 😈

sabob:
Haven't tried this out yet, but over the weekend I noticed the perf test in #2366 was a bit slower that it was couple of months ago:

"create 10000 rows" went up from 500ms to 700ms.
"Delete row" went up from 180ms to 350ms.

Will try out this PR and see the difference.

evs-chris:
I've played around a bit more with this and learned a few interesting things:

1. A Chromebook Flip is roughly as fast as a Nexus 6P (not very fast)
2. ... which are only about 50% slower than Edge on a kickass Windows development machine at DOM manipulation 😕. For reference, by kickass, I mean runs visual studio 2015 without stuttering _anywhere_ - I know, right?
3. My linux dev box appears to be 20-50% slower than the machine doing the benchmarks mentioned in #2366, based on their current table and depending on the test step.
4. This PR (and edge in general) beats the pants off of 0.7.3 in just about every test (not splice last row).
5. Even just keypath caching seems to contribute a fair amount of performance bump for lots of vdom.
6. A ton of DOM is painful to deal with no matter where you are. Pagination is your friend.
7. The Flip actually gets a kill script popup during the test with 0.7.3.

Here are the relevant numbers on the kickass machine:

|test|edge|pr|0.7.3|
|---:|---:|---:|---:|
|create 1000|177.2|166.7|301.9|
|remove first|66.0|23.9|104.8|
|remove last|47.8|13.1|5.6|
|create 10,000|1846.2|1147.1|3660.9|

... and the linux dev machine:

|test|edge|pr|0.7.3|
|---:|---:|---:|---:|
|create 1000|443.8|393.4|882.6|
|remove first|146.0|73.0|229.7|
|remove last|98.6|37.7|11.8|
|create 10,000|2601.9|2535.3|14112.6|

... and the Chromebook Flip:

|test|edge|pr|0.7.3|
|---:|---:|---:|---:|
|create 1000|1892.8|1588.9|1813.3|
|remove first|315.8|171.8|676.1|
|remove last|214.6|77.8|29.5|
|create 10,000|6872.9|6411.6|20807.1|

All tests were done on Chrome 51, except the Flip, which is on 53 (not that it matters).

evs-chris:
I should probably have noted that I updated the tests so that the removals are only working with 1000 rows instead of starting with whatever was left after running however many 'add 100 rows' iterations could finish. My original 35% and 20% figures were based on working with ~1700 rows, so this seems to scale a lot better than edge.

Also, the numbers are avg ms from the output of the jsweb perf test.

JonDum:
Those numbers are insane improvements!

Strengthens my resolve through the dark times of the almost year-long hodgepodge ordeal that the 0.8 refactor has been.

aharpervc:
@JonDum v0.7.3 was released 1 year 2 months 2 days ago :cry:

mefjuu:
@aharpervc Yes, RactiveJS release cycle is the only thing confusing me when I'm picking the library for a project...

evs-chris:
The good news is that Ractive shouldn't have any more massive refractors like 0.8 so maintenance and further development will be much easier. Also, 0.8 is about to go into alpha, so it should be out soon™.

aharpervc:
If the 0.8 api is now stable, perhaps it is appropriate to call it 1.0

skeptic35:
Don't know if I'm right, but I suspect this change to be responsible for breaking ... well, basically just about everything in my application. I was running an 0.8 edge version, built on 13th of June 2016. Everything was looking good, ran beautifully fast, and I was really looking forward to the final 0.8 release. Then today I decided to pull the latest edge version, started up my app and almost nothing worked anymore. And I don't even get any errors. For some reason some of the ```complete``` events don't get fired anymore and some of my observers don't get called anymore. If at least I got some errors, but no, not a single one. 
What led me to suspect that this particular change is responsible, is the fact that at some point I actually did receive an error: "TypeError: m.tryRebind is not a function(…)" in line 8855 of ractive.js
It was the ```tryRebind``` part that led me here. 
I'm really sorry, that I can't provide any helpful input to solve the problem or even narrow it down. As I said: that was the only error I got, possibly initiated by a ```.splice()``` call, some 50 or 60 steps down into the trace log.

It's terribly frustrating! Finally 0.8. seemed to be working just fine and to be getting close to release and now this. What I really don't get is the following: Why introduce this change into 0.8 at all? It doesn't seem related to any bug and the performance of edge already was very much superior to 0.7. So why not just fix the open issues and get 0.8 on the road? It's been long enough already since 0.7.3. No need to slow down the process by introducing massive changes like this one, which always bear the risk to break things.

Sorry for that rant, I'm usually a big fan of Ractive and I really appreciate all the great work you guys do but seeing my app fall apart with the latest edge version was simply the single most frustrating thing that
Ractive ever did to me.

michaljerabek:
The bug with `oncomplete` is caused by https://github.com/ractivejs/ractive/issues/2609.

evs-chris:
I'm reverting this to the old rebinding behavior. I'm not positive that this actually caused the issues or if another bug fix in the interim did, but either way, I've been looking at a number of other somewhat related issues and have come to the conclusion that rebinding needs to happen from the model side, as it appears to in #1740, rather than from the view, as it does now.

Performance may be better with edge than with 0.7, but splicing around lists of any significant length is way slower than it should be. I consider it a bug when the page freezes for a second or two when removing an item from the middle of a list of 1500 items.

_I'll_ consider edge frozen, except bug fixes only now, and everything else will wait until 0.9, unless another maintainer or BDFL wants it in.

skeptic35:
Obviously the fact that my complete handlers were not called was caused by commit 8a17551. But the TypeError for```m.tryRebind``` remains, even if I use a pached version of Ractive including this pull request.
I do think that it is the right decision to postpone this till after 0.8. I'll be glad to help debugging it then, if I can. After all it's not like I'm not in favor of performance improvements. And the last thing I want is to cause frustration on your side :)

Issue #2512, opened 2016-04-13
Adaptors on edge - Discussion


evs-chris:
**Description of the issue:**
I've been looking at fixing some of the adaptor-related bugs on edge, and I've come to the conclusion that there's not really a way to get magic adaptors and array adaptors working together in a clean way. They currently mostly work by getting a reference to their model to side-step the public API, which isn't really ideal. They also don't deal well with being shuffled (see #2005) and can't be used to control models more than one key deep, which is problematic for more advanced adaptors like ractive-ractive.

I think I may be able to eventually mangle things into working, but it feels like I'm fighting the API pretty hard at this point. Of course, I may be completely missing something, and the fix may actually be super simple :smile:.

I know I've mentioned before that I think the array and magic adaptors should be pulled into separate plugins, but I'll say it here to to keep everything together. That would also pave the way for an observer-based magic adaptor to be swapped in for those not requiring legacy browser support.

Link #2288, #2500

**Versions affected:**
edge/0.8

**Potential fix/brain dump:**
The model currently has a handle to its adaptor, but it can't really do much with it beyond setup, reset, and teardown. I was thinking that an adaptor handle object might be a decent solution to communication between the model and adaptor:

* on adaptation, an object is constructed with methods to interact directly with the model and passed off to the adaptor with
  * a set method - to update the model, particularly including nested keypaths
  * a keypath method - to get the current keypath
  * a way to get the ractive instance
  * some sort of shuffle/merge method to signify that stuff changed but only in a re-arranging way - may need a new index helper too
  * maybe a begin/end to correspond to a runloop turn - automatically tracked such that it any methods will automatically do their own loop or participate in the current one for the adaptor if present
  * a method to indicate how deep this adaptor should manage keypaths - probably just 3 flavors would work - 0, 1, and all the way - like ractive-ractive-type adaptors would need
  * keypath helper methods (split, escape, etc) for convenience
  * a method to tell the model "I'm done, replace me with this new value", though set may address this - I'm thinking about promise adaptors in particular, but other situations may call for this sort of thing
* on shuffle, the adaptor doesn't need to change but it does need to know that it is no longer pointing at the same model. The handle could just have its pointer updated and the actual adaptor code need not care.
  * optionally call some sort of keypathChanged/shuffled adaptor callback?
* on set, the usual check for reset and the need to teardown still tracks. I think an automatic strict equality check is probably in order for the reset.
* teardown is roughly the same
[...]

PR #2595, opened 2016-06-24
Attach/Detach/Map external instances - kinda like ractive-ractive but with view too - RFC

evs-chris:
This is my attempt to solve one of the last big issues I have with components: there's not a good API for dealing with them outside of the template. My primary irritation being that there's a bit more friction making larger apps with lots of views than I'd like. There are three concerns I've tried to address here:

1. There's no (good/safe) way to share data between to instances without having one be a component child of the other, and even then it can be a bit touchy.
2. There's not an obvious way to dynamically render an externally controlled (Ractive) component into a Ractive-controlled piece of DOM without getting hacky.
3. There is some layout support, but it's not really conducive to dealing with complex component structures as-is.

Some of those are partially addressed by conditional directives and event expressions, but in a very small way.

Here's what this does, or will do when it's finished, to address those:

* Add a ~~`mapPath`/`unmapPath`~~ `addMapping`/`removeMapping` API to allow programmatic control over what mappings a component gets. Mapping directive attributes are great, but sometimes you need more control or dynamism than can be achieved with those.

* Add `attachChild` and `detachChild` API to allow programmatic insertion and removal of external instances/components in a vdom-aware manner. There are actually two sides to this, one of which addresses points 2 and 3 above.

  * Starting with the simpler one, you can attach a child instance, and it will take on a new parent instance, meaning that events will bubble and, assuming no isolation, you get implicit mappings to the root of the parent when the child attaches and rebinds.

* For the more complex side, there is a new vdom item that is somewhat like the existing Component called an Anchor. An anchor has a name and may house one or more components depending on how it is specified. Since it's similar to a partial, the preliminary syntax is `{{>>name [multi]}}`, where `multi` is optional and specifies that the anchor should be multi-tenant. `attachChild` accepts an options object that may specify a target name, and if it does, the instance is re-rendered into the matching anchor. If no anchor is found, the instance is unrendered and will render if/when a matching anchor appears. If an anchor is unrendered, then its content instances are also unrendered.
A single tenant anchor will render the last instance in, so if you attach a new child and specify an anchor that is occupied, the current tenant will be unrendered and the new tenant rendered. When the new tenant is removed, the the previous tenant will be re-rendered.
A multi-tenant anchor will simply append any instances that target it. The instance also render and unrender with the anchor, as you'd expect, so they work nicely in sections.
I've considered a few more advanced behaviors for anchors, like multiple anchors with the same name allowing children to fill them in order, but I wanted to get some feedback before I get too far off the rails. As a for instance, specifying queue ordering (fifo/filo) for handling single tenant anchors.

I think the layout/anchor portion would be a huge win for more complex apps and navigation, though I've been wrong before :smile:. I also think the mapping API would be tremendously useful for larger apps with many components/views. I think simply having non-anchored/external instances would be a pretty decent help for arranging components so that you could at least share base state easily without having to shove all of your dom into one overarching Ractive template.

To expand a bit on multi-tenant anchors, I see those as useful for components that are static or absolutely positioned. There's probably more that needs to be done there to make that fully useful, like let the instance know that it is number x of y tenants.

**Status**

* [x] Basic attach/detach support
* [x] find working
* [x] findAll working - including live queries
* [x] findComponent working
* [x] findAllComponents working, including live queries
* [x] mapPath
* [x] unmapPath
* [x] Basic anchor support
* [x] multi-tenant anchor support
* [x] safe transitions
* [ ] API fully fleshed and discussed
* [ ] Functionality feedback
* [x] Tests!
* [x] More tests!

**Is breaking:**
I don't think so, but it's based on the try-rebind branch that can handle rebinding when mappings appear and disappear a bit more gracefully than edge.

**Reviewers:**
Yes, please! Any feedback is welcome, including hacking. I've seen a number of requests that match some portion of what I'm trying to address here.

### DIFF
--- a/src/Ractive/construct.js
+++ b/src/Ractive/construct.js
@@ -145,6 +145,13 @@ function initialiseProperties ( ractive ) {
 	// links
 	ractive._links = {};
 
+	// external children
+	ractive._children = [];
+	ractive._anchors = [];
+
+	// manual mappings
+	ractive._mappings = [];
+
 	if(!ractive.component){
 		ractive.root = ractive;
 		ractive.parent = ractive.container = null; // TODO container still applicable?
--- a/src/Ractive/construct.js
+++ b/src/Ractive/construct.js
@@ -150,7 +150,7 @@ function initialiseProperties ( ractive ) {
 	ractive._anchors = [];
 
 	// manual mappings
-	ractive._mappings = [];
+	ractive._mappings = {};
 
 	if(!ractive.component){
 		ractive.root = ractive;
--- a/src/Ractive/initialise.js
+++ b/src/Ractive/initialise.js
@@ -24,26 +24,9 @@ export default function initialise ( ractive, userOptions, options ) {
 	configHook.fire( ractive );
 	initHook.begin( ractive );
 
-	let fragment;
-
 	// Render virtual DOM
-	if ( ractive.template ) {
-		let cssIds;
-
-		if ( options.cssIds || ractive.cssId ) {
-			cssIds = options.cssIds ? options.cssIds.slice() : [];
-
-			if ( ractive.cssId ) {
-				cssIds.push( ractive.cssId );
-			}
-		}
-
-		ractive.fragment = fragment = new Fragment({
-			owner: ractive,
-			template: ractive.template,
-			cssIds
-		}).bind( ractive.viewmodel );
-	}
+	let fragment = ractive.fragment = createFragment( ractive, options );
+	if ( fragment ) fragment.bind( ractive.viewmodel );
 
 	initHook.end( ractive );
 
@@ -63,5 +46,25 @@ export default function initialise ( ractive, userOptions, options ) {
 				});
 			}
 		}
+	}
+}
+
+export function createFragment ( ractive, options = {} ) {
+	if ( ractive.template ) {
+		let cssIds;
+
+		if ( options.cssIds || ractive.cssId ) {
+			cssIds = options.cssIds ? options.cssIds.slice() : [];
+
+			if ( ractive.cssId ) {
+				cssIds.push( ractive.cssId );
+			}
+		}
+
+		return new Fragment({
+			owner: ractive,
+			template: ractive.template,
+			cssIds
+		});
 	}
 }
--- a/src/Ractive/prototype.js
+++ b/src/Ractive/prototype.js
@@ -1,4 +1,5 @@
 import add from './prototype/add';
+import addMapping from './prototype/addMapping';
 import animate from './prototype/animate';
 import attachChild from './prototype/attachChild';
 import detach from './prototype/detach';
@@ -14,7 +15,6 @@ import get from './prototype/get';
 import getNodeInfo from './prototype/getNodeInfo';
 import insert from './prototype/insert';
 import link from './prototype/link';
-import mapPath from './prototype/mapPath';
 import merge from './prototype/merge';
 import observe from './prototype/observe';
 import observeList from './prototype/observeList';
@@ -24,6 +24,7 @@ import on from './prototype/on';
 import once from './prototype/once';
 import pop from './prototype/pop';
 import push from './prototype/push';
+import removeMapping from './prototype/removeMapping';
 import render from './prototype/render';
 import reset from './prototype/reset';
 import resetPartial from './prototype/resetPartial';
@@ -40,7 +41,6 @@ import toCSS from './prototype/toCSS';
 import toHTML from './prototype/toHTML';
 import transition from './prototype/transition';
 import unlink from './prototype/unlink';
-import unmapPath from './prototype/unmapPath';
 import unrender from './prototype/unrender';
 import unshift from './prototype/unshift';
 import update from './prototype/update';
@@ -48,6 +48,7 @@ import updateModel from './prototype/updateModel';
 
 export default {
 	add,
+	addMapping,
 	animate,
 	attachChild,
 	detach,
@@ -63,7 +64,6 @@ export default {
 	getNodeInfo,
 	insert,
 	link,
-	mapPath,
 	merge,
 	observe,
 	observeList,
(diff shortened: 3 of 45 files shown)


evs-chris:
Something didn't look right about `mapPath`/`unmapPath`, so I renamed them `addMapping`/`removeMapping`, which at least gels with what they do and other verbiage in the docs a little better.

I also have the beginnings of the requisite test suite expansion, so you can kinda get a feel for what this does from the tests.

evs-chris:
I got live queries working and found and fixed a few potential live query issues in edge. It looks like if a component is un-rendered and re-rendered, it will get added to any live queries again in edge.

Anchored children come out in the query in fragment order, the same as regular components. Non-anchored children sort to the end of the list.

I still want to get a few more query tests in place to cover behavior of nested components in un-anchored children, but beyond that, I think the code is complete here.

aharpervc:
Does this mean the following will work?

```
var Comp = Ractive.extend({});

var c1 = new Comp();
c1.set("a", 1);

var c2 = new Comp();
c2.set("c1", c1);

// this VVVVV
assert(c2.get("c1.a") == 1);
```

evs-chris:
@aharpervc yes, but like this:
```js
const r = new Ractive({}); // any ractive will do, including components
const c = new Ractive({ // also any ractive
  data: { foo: 'bar' }
});
r.addMapping( 'c', '', { ractive: c } );
r.get('c.foo'); // 'bar'
r.set('c.foo', 'test');
c.get('foo'); // 'test'
```

That's the mapping portion. `removeMapping` can be used to remove the `c` mapping in `r`.

For the curious, here's a quick `attachChild` example:

```js
const r = new Ractive({
  el: someElement,
  template: 'example template with an {{>>anchor}} and some stuff on the other side',
  data: { foo: 'bar' }
});
const sep = new Ractive({
  el: someOtherElement,
  template: 'this is my template and will always live in someOtherElement {{foo}}',
  // when attached foo will implicitly map to the new parent foo in r
});
const join = new Ractive({
  el: yetAnotherElement, // this is actually optional because
  // this one will be anchored, and thus, become part of the parent template like a component
  template: 'could also have have implicit mappings, or could be isolated and use addMapping'
});

r.attachChild(sep, { name: 'foozle' }); // name is optional, but can be handy for findComponent
// sep is now a child of r, but manages its own rendering
// events bubble, and other component stuff still work

r.attachChild(join, { target: 'anchor' }); // since join is already rendered, it will be unrendered
// and re-rendered in the anchor of r
// this one is now effectively a full component of r, including render/unrender
// so if the anchor is in a section that unrenders, join will also be unrendered

r.detachChild(sep); // sep stays rendered, but has all of its implicit mappings removed
// it also loses its parent, so events cease bubbling, etc

r.detachChild(join); // join is now unrendered, since it was part of r
// it can be manually rendered elsewhere, or even attached to another instance
// or a different anchor in the same instance
```

Transitions still happen as you would expect when an instance is rendered or un-rendered when being attached, detached, or dropping out with its anchor.

As mentioned in the opening comment, single tenant anchors unrender their current tenant if a new one is attached. They will then re-render the previous one when the new one is detached. Multi-tenant anchors just append the new instance.

As currently written, _any_ attached children contribute to `find`, `findAll`, `findComponent`, and `findAllComponents`, including for live queries. I was thinking that perhaps making un-anchored children either opt-in or having an option to opt-out would be useful.

aharpervc:
Oh, that addMapping bit is a bummer.

One of the nice things about the Ractive-ractive adapter is that you can set it once in your beforeInit, and any subsequent `set` call works for any key name for any future ractive instance. What you're saying with this API is that addMapping needs to get called *for every key whose value is a ractive instance*. Well that is a huge hassle. Imagine a function signature with a param that can either be a pojo or a ractive instance, and you call `this.set("x", p)`. That'll work today with Ractive-ractive and break tomorrow with this branch.

Or, another example: what about when one of my components has in data an array of ractive components. How would splicing that property work with regards to add/remove mapping?

What do you think about the possibility auto detecting when a `set` value param is a ractive instance and automatically mapping it?

evs-chris:
Mappings only work as top-level keys, so I don't know that you could fully replace ractive-ractive with this support. This is more aimed at manually managed components and instance than using instances as data directly, as the array of instances would be. ractive-ractive should still work after this change, though it may have some issues with edge at the moment. An adaptor API update that can handle things like ractive-ractive more gracefully and move the array and magic adaptors into their own independent plugins is an outstanding issue (#2512).

For the `set`, it wouldn't be difficult to detection for setting an instance as a top-level key and change the set into a mapping automatically. I'm not sure you'd want that built-in rather than a separate monkey patch extension, though, because there're definitely a few traps for the unwary in that direction.

aharpervc:
> Mappings only work as top-level keys

Does that mean `r.addMapping('comp_container.comp', '', { ractive: c } );` is invalid, because `comp_container.comp`, a data property key path I just made up but hypothetically could be valid, isn't a "top level key"?

On the topic of auto mapping, what if `map: Boolean|undefined` could be passed to `set`? Just spitballing. Public api's to manually manage mappings is less than ideal.

evs-chris:
Yes, mappings must be top-level e.g. `<component foo="{{bar}}" baz="{{bat.bop}}" />`. This API is just the JS version of adding the attributes to the component tag and is mostly proposed to deal with dynamic components, which currently must be handled _entirely_ in template. In-template-only components mean that you have to know up-front about everything that the component will need. It also means that if the component gets unrendered, it will be torn down, losing any local state, and recreated when it gets rerendered, so you can't easily solve the dynamism problem by using dynamic templates in many scenarios.

It sounds like what you're after is closer to a cross-instance version of links, which sounds doable, but not quite in scope for the original aim for this PR. I think there is some potential for merging links and mappings to cut out some redundancy in the code, and it would address a few open enhancement requests. I'll look into that and see what shakes out.

evs-chris:
Superceded by #2674

Issue #2604, opened 2016-07-04
BUG: Arrays + components + observers create duplicates


michaljerabek:
**Description of the bug:**
Rearranging "an array of components" creates duplicates if an observer changes data in the component.

**Versions affected:**
0.8-edge

**Platforms affected:**
Any browser

**Reproduction:**
http://jsfiddle.net/saut2tx2/3/

EDIT: The description is not entirely accurate. It simply creates a new object.

evs-chris:
This is now as fixed as it can be for 0.8. You will need to use deferred observers for scenarios like this where you want the observer to stop firing when the component is removed e.g. `this.observe('...', () => {}, { defer: true });`.

evs-chris:
This is back in a weird way. Basically, since modifyArrays is off by default now, the first splice (removing the element) doesn't notify ractive of the change, so the second splice creates a duplicate when it _does_ notify ractive. I'm looking into the best way to handle this, but I think it's probably to allow array self merges as proposed in #2648.

PR #2618, opened 2016-07-19
Make sure cancelled observers don't fire - #2604

evs-chris:
This adds a check in observers that stops them from firing after `cancel` has been called even if they were already scheduled to run in the current loop. This means that a deferred observer will no longer fire in components that have been removed from the view. So observers that set data on an array that is their context and/or mapping don't create extra array elements as they are spliced out.

#2604

**Is breaking:**
Maybe? I'm not really sure what the prescribed behavior is in this case.

**Reviewers:**
@martypdx @Rich-Harris @MartinKolarik @fskreuz - again, need feedback on whether this is the right way to handle this or if the check should be handled by the dev.

### DIFF
--- a/src/Ractive/prototype/observe.js
+++ b/src/Ractive/prototype/observe.js
@@ -118,6 +118,7 @@ class Observer {
 	}
 
 	cancel () {
+		this.cancelled = true;
 		if ( this.model ) {
 			this.model.unregister( this );
 		} else {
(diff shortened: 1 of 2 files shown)


MartinKolarik:
I think this is the way to go, but it seems it doesn't fix the example provided in #2604. Also, it looks like you missed pattern observers?

evs-chris:
Thanks! Yes, forgot to mention that this still requires the observer to be deferred because otherwise the observers are run before the component is removed. That seemed like the least confounding fix for immediacy.

My long term solution involves reworking the way shuffles work to be handled through the model rather than through deps alongside the view. Observers have some interesting corner cases with components that I hope to address in 0.9.

I wrote a test for pattern observers, and it looks like since they are handled a bit differently that they weren't affected by this particular scenario... unless the test is inadequate.

MartinKolarik:
Indeed, I missed the second test. :+1: then.

Issue #2629, opened 2016-08-10
Recomputing a computed values causes observers to re-fire, even if the value hasn't changed


Rich-Harris:
**Description of the bug:**

Recomputing a computed values causes observers to re-fire, even if the value hasn't changed

**Versions affected:**

Edge

**Reproduction:**

http://jsfiddle.net/ub7540xq/

**Code:**

```js
var ractive = new Ractive({
  computed: {
  	int () {
    	return Math.round( this.get( 'number' ) );
    }
  },
  data: {
    number: 1,
    observerCalledTimes: 0
  }
});

ractive.observe( 'int', function ( int ) {
  console.log( int );
  this.add( 'observerCalledTimes', 1 );
});

ractive.set( 'number', 1.1 );
ractive.set( 'number', 1.2 );
ractive.set( 'number', 1.3 );
ractive.set( 'number', 1.4 );

console.log( ractive.get( 'observerCalledTimes' ) );
// expected 1, got 5
```

evs-chris:
I thought the fix for this would be pretty simple: just check `!isEqual( newValue, oldValue )` before scheduling the observer, but it turns out that that runs afoul of the #771 test that seems to hate me so. Out of curiosity does anyone actually use observers for validation i.e. mutating a value from an observer that is watching it as the user types? I would think per-element laziness would preclude the need for that particular facility, but I've been wrong so very many times.

giovannipiller:
> mutating a value from an observer that is watching it as the user types?
@evs-chris: like observing `a` and modifying it while user types on a `<input ... value="{{a}}">`?

If that's so, I'd say per-element laziness is an ok approach.

michaljerabek:
> Out of curiosity does anyone actually use observers for validation i.e. mutating a value from an observer that is watching it as the user types?

Yes, I use observers for this. (And it causes many issues, because of those bugs, I've reported.)

PR #2633, opened 2016-08-15
Check observer value before firing in case a computed is invloved - #2629

evs-chris:
This adds an `isEqual` check on observers so that observers on computations don't fire multiple times for an unchanged value just because the computation was invalidated.

That by itself is not compatible with two-way bindings and mutating observers, so this also adds a check in the `handleChange` of `Binding` that compares the last value it notified for with the one that it would notify for now. If they are the same, it just skips the notification. That avoids firing the same value multiple times and confusing the observer based on the state of the model/observer and the binding/attribute being in different states while the user still has input control i.e. the control is locked as far as Ractive is concerned so that it doesn't cause the cursor to jump home on every keystroke.

I _think_ that has the issue covered. It's kinda hard to say with some of the nuance involved in these scenarios, but all of the tests turn green. I still say that mutating observers on bindings are a whole heap of trouble :laughing: 

#2629 

**Is breaking:**
Don't think so.

### DIFF
--- a/src/Ractive/prototype/observe.js
+++ b/src/Ractive/prototype/observe.js
@@ -136,7 +136,10 @@ class Observer {
 
 	handleChange () {
 		if ( !this.dirty ) {
-			this.newValue = this.model.get();
+			const newValue = this.model.get();
+			if ( isEqual( newValue, this.oldValue ) ) return;
+
+			this.newValue = newValue;
 
 			if ( this.strict && this.newValue === this.oldValue ) return;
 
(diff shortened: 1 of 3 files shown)


aharpervc:
What's the behavior if `update` is called manually? Like in your test, `r.update('int')`?

evs-chris:
@aharpervc nothing happens because there still isn't actually a change to observe. If you actually modify some value externally that would affect the computation, then the observer may fire again depending on whether or not the value coming into the observer actually changed.

I should note that this does not affect objects, because `isEqual` with objects will always be false. There just isn't a good way to determine if an object (or array) has changed somewhere in a deeply nested path, so Ractive's stance has always been to leave those sorts of cases to the library user to sort out. This only really affects primitives.

PR #2637, opened 2016-08-22
Attempt to address remaining blockers for 0.8 - RFC/Review

evs-chris:
I want to apologize in advance for the big funky PR and the length of the description here. There are a number of chewy issues outstanding in edge that I consider to be blockers and this is my attempt to address them. There are a few somewhat overlapping things combined here:

## Rebinding
This moves from view-based rebinding to model-based rebinding, meaning that instead of the shuffle telling the view that everything in the view needs to rebind, the model that is shuffling tells all of its deps that its value is moving to a new model. It also supplies the target model so that the deps can make sure that it matches so that deps on the absolute path `list.0.bar` don't accidentally get shuffled to `list.1.bar` on an unshift. This is a good deal faster than having everything in the view re-resolve and has the added bonus of catching things that aren't view-related, like observers, mappings, and resolvers.

There are a handful of outstanding shuffle bugs, particularly on observers, that are resolved by this type of rebinding along with the next change...

## Mappings
This also takes the mapping and link constructs and merges them using a new type of model. The `LinkModel` is simply a model that points to another model, which may be a plain data model or another link itself. When a link model is mutated, it triggers the mutation on its source model and lets it filter back to the link, so there's still a single source of truth at the origin. That creates something like a parallel namespace for the link that manages its own deps, resolvers, etc. The separate namespace makes changing the root source of the link much simpler, since now it doesn't have to try to comb through combined deps and figure out which ones should be shuffled off to the new source and which ones should remain on the model.

**Side Note:** Since there was a ton of overlap between `Model` and `LinkModel`, I rebased them both on a new `ModelBase`. I'm not crazy about the name, but it DRYs up a good bit of code.

Any regular model may now be turned into a link by having a `LinkModel` installed on it at `_link`. The `ModelBase.link` method takes care of creating a link on a model, transferring (rebinding) the model's deps to the new link, and making sure everything updates. Linking on an already linked model simply updates the source of the link.

The issues with shuffling under a mapping e.g. `{{#each list}}<component foo="{{.}}" />{{/each}}` and observers are much easier to address, and I think they're all covered here. Conditional mappings also have their corner cases addressed when combined with the rebinding changes.

Since the method to generate aggregate root data on `instance.get()` changed to accommodate links, I also added an options object for `get` that allows you to control whether or not you want the aggregate or the source data directly. The root model defaults to providing the virtual data on get, and child models default to providing the source data so as not to be a breaking change. You can now call `instance.get({ virtual: false })` to get the actual root object and `instance.get('some.path', { virtual: true })` to get `some.path` including any children that may be links.

Since mappings and links are the same thing, this would open up the possibility to have nested mappings for components. It would also make cross instance links equivalent to mappings. I haven't addressed either of those here, but it looks like they'd be easy to achieve.

## Expressions
Expressions in edge are composed of a proper `Computation` and an `ExpressionProxy` that manages the computation. The problem with that is that the computation is registered with the root of the model based on its signature and signature can, and with expressions often do ,overlap. They also create a giant map to deal with in the root model if there happen to be a few expressions in a large iterative section. I was originally going to try to reuse the computations, but then I realized that expressions with object and array literals in them can't be safely reused.

This takes the `ExpressionProxy` and detaches it from a computation to let it stand as its own full model somewhat mirroring what `Computation` does. There is probably more that could be shared between them, but I'll address that later if this works out. Now realistically, there was no way to interact with an expression model outside of the template before, but there is definitely not a way to do so here where the expression is not registered anywhere in the model tree (other than as a dep). So to address that, I extended the context methods to resolve to actual models and perform their operations instead of resolving to keypaths and just calling the appropriate instance method. That adds a bit of code to the context methods, but it also opened up an opportunity to share more code as generalized helpers among the instance methods, notably handling `*` keypath resolution and maps of paths to values.

So, you can now cause an update in an expression as long as you have an anchoring event or node info object from an element where the expression is part of the context. To be clear, if you want to call any instance methods from an expression context, you will need to use the context method on the event object to do so, as the keypath can no longer be resolved e.g. `{{#foo()}}<a on-click="event.toggle('.bar')">click</a>{{/}}` as opposed to `{{#foo()}}<a on-click="@this.toggle(@keypath + '.bar')">click</a>{{/}}`. _I_ think that's an overall improvement anyways, and it has the added benefit of not updating the wrong section if the expression keypath happened to overlap somewhere else in the template. I also added `reverse` and `sort` to the context methods, as they were missed the first time around.

`if`/`elseif`/`else` in edge use a clever negation and stacking of their expressions to split each branch into its own section. The downside to that is that side-effecting or expensive computations with be run an additional time for each successive branch. This PR drops the expression duplication and adds a flag on elseif and else sections that marks them as subordinate. When a subordinate section binds, it looks up its nearest sibling and links it with itself. Then when any of the siblings updates, it notifies its subordinate siblings to update as well, and each of them checks to see that none of the prior siblings is truthy before checking their own condition. This effectively emulates the usual branching method of most languages. This is technically a breaking template change, but the new code can still manage an old template, so I don't think a template bump is necessary since there's already one in edge.

## Epilogue
If you have some time to look over the code, I'd really appreciate some review. If you have some more time to build it and try it out with a project, that would be great too! I'm thinking about publishing some scoped Ractive experiments on npm, and this would probably be a good place to start considering that if it passes review I'd still like to get some mileage on it before merging.

Thanks!

## TODOs
* [x] Implement and test rebinding for decorators
* [x] Implement and test rebinding for transitions

### DIFF
--- a/src/Ractive/construct.js
+++ b/src/Ractive/construct.js
@@ -142,9 +142,6 @@ function initialiseProperties ( ractive ) {
 	// observers
 	ractive._observers = [];
 
-	// links
-	ractive._links = {};
-
 	if(!ractive.component){
 		ractive.root = ractive;
 		ractive.parent = ractive.container = null; // TODO container still applicable?
--- a/src/Ractive/prototype/animate.js
+++ b/src/Ractive/prototype/animate.js
@@ -30,6 +30,30 @@ function getOptions ( options, instance ) {
 	};
 }
 
+export function animate ( ractive, model, to, options ) {
+	options = getOptions( options, ractive );
+	const from = model.get();
+
+	// don't bother animating values that stay the same
+	if ( isEqual( from, to ) ) {
+		options.complete( options.to );
+		return noAnimation; // TODO should this have .then and .catch methods?
+	}
+
+	const interpolator = interpolate( from, to, ractive, options.interpolator );
+
+	// if we can't interpolate the value, set it immediately
+	if ( !interpolator ) {
+		runloop.start();
+		model.set( to );
+		runloop.end();
+
+		return noAnimation;
+	}
+
+	return model.animate( from, to, options, interpolator );
+}
+
 export default function Ractive$animate ( keypath, to, options ) {
 	if ( typeof keypath === 'object' ) {
 		const keys = Object.keys( keypath );
@@ -42,27 +66,6 @@ ${keys.map( key => `ractive.animate('${key}', ${keypath[ key ]}, {...});` ).join
 ` );
 	}
 
-	options = getOptions( options, this );
 
-	const model = this.viewmodel.joinAll( splitKeypath( keypath ) );
-	const from = model.get();
-
-	// don't bother animating values that stay the same
-	if ( isEqual( from, to ) ) {
-		options.complete( options.to );
-		return noAnimation; // TODO should this have .then and .catch methods?
-	}
-
-	const interpolator = interpolate( from, to, this, options.interpolator );
-
-	// if we can't interpolate the value, set it immediately
-	if ( !interpolator ) {
-		runloop.start();
-		model.set( to );
-		runloop.end();
-
-		return noAnimation;
-	}
-
-	return model.animate( from, to, options, interpolator );
+	return animate( this, this.viewmodel.joinAll( splitKeypath( keypath ) ), to, options );
 }
--- a/src/Ractive/prototype/get.js
+++ b/src/Ractive/prototype/get.js
@@ -1,8 +1,8 @@
 import { splitKeypath } from '../../shared/keypaths';
 import resolveReference from '../../view/resolvers/resolveReference';
 
-export default function Ractive$get ( keypath ) {
-	if ( !keypath ) return this.viewmodel.get( true );
+export default function Ractive$get ( keypath, opts ) {
+	if ( typeof keypath !== 'string' ) return this.viewmodel.get( true, keypath );
 
 	const keys = splitKeypath( keypath );
 	const key = keys[0];
@@ -22,5 +22,5 @@ export default function Ractive$get ( keypath ) {
 	}
 
 	model = this.viewmodel.joinAll( keys );
-	return model.get( true );
+	return model.get( true, opts );
 }
--- a/src/Ractive/prototype/link.js
+++ b/src/Ractive/prototype/link.js
@@ -17,7 +17,7 @@ export default function link( there, here ) {
 		model = model.joinAll( sourcePath.slice( 1 ) );
 	}
 
-	this.viewmodel.joinAll( splitKeypath( here ) ).link( model || this.viewmodel.joinAll( sourcePath ) );
+	this.viewmodel.joinAll( splitKeypath( here ) ).link( model || this.viewmodel.joinAll( sourcePath ), there );
 
 	runloop.end();
 
--- a/src/Ractive/prototype/link.js
+++ b/src/Ractive/prototype/link.js
@@ -8,77 +8,18 @@ export default function link( there, here ) {
 	}
 
 	const promise = runloop.start();
-
 	let model;
-	let ln = this._links[ here ];
-
-	if ( ln ) {
-		if ( ln.source.model.str !== there || ln.dest.model.str !== here ) {
-			ln.unlink();
-			delete this._links[ here ];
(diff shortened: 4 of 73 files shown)


evs-chris:
No, that's definitely not intended. In fact, I don't even see where it's happening, as there's nowhere that sets `el` other than render, which isn't called on teardown. I'll see if I can track it down :smile:

kouts:
If understand correctly @keypath cannot longer be used inside e.g an {{#each}} section.
Is this correct? Isn't this a breaking change?

evs-chris:
@kouts you can still use `@keypath` in any context as long as it's not an expression, so `{{#each lists[whichList]}}{{#some.child}}<button on-click="@this.set(@keypath + '.foo', 'whatever')">click me</button>{{/}}{{/each}}` is still fine. Having `{{#each someExpression()}}<button on-click="@this.set(@keypath + '.foo', 'whatever')">click me</button>{{/each}}` will no longer work with this changes as it exists, but there were already template constructs wherein that would fail because of the overlap, making only the last instance of the expression update.

So yes, this is a breaking change in that regard. Is that a frequently used construct? I have it some as a workaround, but I'm migrating to the much less painful equivalent that is already available in edge and better here: `{{#each someExprsession()}}<button on-click="event.set('.foo', 'whatever')">click</button>{{/each}}`. I can rig a deprecation cycle if this is going to land and it would help.

__Vote__ here for your preference on deprecation: thumbs up yes add in a deprecation cycle for using `@keypath` in an expression context, thumbs down no.

kouts:
Oh, ok it's a thumbs up on deprecation from me then.
Thanks for the detailed explanation @evs-chris.

martypdx:
Read through novel and it all made sense to me. Scanned through code, looks good too. Outside of the large number of added tests, I think you're close to a net 0 in added code 😀

evs-chris:
TL;DR faster rebinding, more complete (decorators and transitions are handled), magic arrays are painful

**Decorators and transitions** now have proper rebind handlers, so their params don't get confused if they happen to be shuffled. There are now also tests to cover more weird cases.

**Rebinding is now considerably faster**, though I suppose it's more the post-rebind update that's faster. While profiling I discovered that every dep in a shuffled tree was getting updated after the shuffle,   which can be quite painful for large iteration counts and complex templates with expressions. All that updating is really unnecessary with a shuffle, because the underlying values don't change - they just move around a bit. To remedy that, I split the post-shuffle update into two stages: early and mark. The early stage runs immediately after the rebinding for the tree is complete. The mark stage runs immediately after the mark towards the end of the shuffle method.

Mustaches and expressions will now wait until the mark stage to register with their new dep, so that they do not get called on to update during the mark. They do so by checking the new safe flag on rebinding, which is true for shuffling models, but not for keys, indexes, and keypaths or during remapping. The end result is that expressions and mustaches no longer are dirtied and updated during a shuffle unless they actually need to. If you avoid index or key (non-shufflable) based deps, shuffling is super fast, and it's even sped up considerably if you do have those deps. Expressions also recompute much less.

**Magic arrays are painful** as an aggregate wrapper consisting of separate magic and array adaptors, really weird things go on with them. When the array side is doing a splice operation, each index has its setter called, which in turn triggers a set through the magic side. Naturally this can lead to some really weird behavior when shuffling. I've temporarily address this by adding a way to lock the magic side from the array side so that the setters aren't called during shuffles, thus avoiding the weirdness with updates. To be clear the weirdness would be the view getting completely out of sync as it updates against an already spliced array and is then shuffled with the new splice info afterwards. I hope to get these fully handled in an adaptor overhaul PR at some point.

evs-chris:
@JonDum I haven't been able to reproduce the issue with the element not being set on teardown. If you've got a few minutes to spare could you put together a demo that I can step through?

evs-chris:
Also forgot to mention that there is now deprecation in place for calling methods with `@keypath` derivations within an expression context. It'll throw out a warning in debug mode each time it happens, and the deprecation and support will be removed in 0.9. It will also fall over in the same way as current edge if there happens to be a duplicate expression keypath.

JonDum:
> I haven't been able to reproduce the issue with the element not being set on teardown. If you've got a few minutes to spare could you put together a demo that I can step through?

Can't repro anymore after latest merge of dev into this branch. Whatever it was seems to be fixed now. 👍 🐋

evs-chris:
Right on! If there are no objections, I shall merge this tomorrow so I can deal with any bugs that shake out this week.

PR #2645, opened 2016-08-30
BREAKING: default modifyArrays to false - RFC

evs-chris:
This defaults modifyArrays to false. The goal is to eventually split magic and array adaptors into plugins, and this is one of the first steps to do so - that has the extra bonus of boosting performance.

The longshot is that if you want to use the non-Ractive flavors of array methods, you would now have to opt in. I believe the Ractive versions have been pushed as the preferred way to handle arrays for a while. What does everyone think? Change it, or leave it as it is for now?

**Is breaking:**
Yes

### DIFF
--- a/src/Ractive/config/defaults.js
+++ b/src/Ractive/config/defaults.js
@@ -21,7 +21,7 @@ export default {
 	data:                   {},
 	computed:               {},
 	magic:                  false,
-	modifyArrays:           true,
+	modifyArrays:           false,
 	adapt:                  [],
 	isolated:               false,
 	twoway:                 true,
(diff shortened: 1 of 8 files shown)


kouts:
I think it's ok to change it.

fskreuz:
I think Ractive methods, setters and getters should be the way to go. This way, Ractive can do intermediate operations where required. Direct setting of values, whether it's array methods or magic mode, always come with gotchas that always makes one go back and read the docs.

JonDum:
Yes PLEASE. It has always been a source of confusion on my team.

PR #2648, opened 2016-08-31
Allow arrays to be merged with themselves - RFC

evs-chris:
Arrays are not currently allowed to merge with themselves, which means you can't modify them in place and then have a nice non-destructive shuffle for the DOM. Since the model representing the array can keep track of its length, and its children keep a reference to their value, the original array can be reconstructed pretty easily. This allows merging arrays with themselves by doing that.

There have been a few questions recently about how to move items in an array without tearing it down in one place and recreating it in another. I think this may be a reasonable way to achieve that.

**Is breaking:**
Nope

### DIFF
--- a/src/Ractive/prototype/merge.js
+++ b/src/Ractive/prototype/merge.js
@@ -21,9 +21,7 @@ export default function Ractive$merge ( keypath, array, options ) {
 	const promise = runloop.start( this, true );
 	const value = model.get();
 
-	if ( array === value ) {
-		throw new Error( 'You cannot merge an array with itself' ); // TODO link to docs
-	} else if ( !isArray( value ) || !isArray( array ) ) {
+	if ( !isArray( value ) || !isArray( array ) ) {
 		throw new Error( 'You cannot merge an array with a non-array' );
 	}
 
--- a/src/model/Model.js
+++ b/src/model/Model.js
@@ -50,6 +50,7 @@ export default class Model {
 
 			if ( parent.value ) {
 				this.value = parent.value[ this.key ];
+				if ( isArray( this.value ) ) this.length = this.value.length;
 				this.adapt();
 			}
 		}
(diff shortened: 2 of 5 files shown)


JonDum:
In practice, what does this mean? Does this only affect `.prototype.merge()` calls? Or does this affect shuffling when you `get()` an array, modify it, then `set()` it back?

evs-chris:
As it stands, this only affects `merge`. There was a note sometime way back that merge should probably turn into a flag on set, and I'm inclined to agree with that with the caveat that I think it should be the default behavior. There's no reason `set` couldn't check to see if the target was an array that happened to be `===` to the new value and perform a merge instead. I'll punt on that til 0.9 though. I only merged this because it has a fix for the weird duplication issue in #2604.

Honestly, up to this point I've never used merge, but having to play around with it, I will probably replace all of my `splice`ing around elements to move them up and down lists to be `on-click="@this.merge( ../splice(@index - 1, 0, ../splice(@index, 1)[0]) && ../)"` or thereabouts.

I also considered adding a merge option that let you just specify indexes to move, remove, and add, but I never came up with anything that wasn't weird.

JonDum:
> `on-click="@this.merge( ../splice(@index - 1, 0, ../splice(@index, 1)[0]) && ../)"`

That's pretty scary looking... 😨 

I think the expectation is that Ractive does virtual-dom-diffing/minimal-dom-rerendering  "magically" by default. I think it's a win if we get that in all methods rather in just `merge` since I would say most people (myself included) are somewhat afraid/unsure of when to use `merge`.

evs-chris:
> That's pretty scary looking... 😨

😆 Fair enough. It behaves a bit better than what I normally use when I'm lazy and just want to move stuff in-template `on-click="event.splice(../, @index - 1, 0, event.splice(../, @index, 1).result[0])"`.

The difference between `set` and `merge` is that `set` will just trigger updates and make the lengths match however is necessary by adding or subtracting to/from the of the iteration. Not exactly dom diffing, but it's pretty darned efficient. `merge` will trigger transitions on the elements that are coming in and out, even if the net length of the list stays the same, kinda like doing all of the appropriate splices at once. I don't know if that's a good description of the difference. I _have_ wondered if a shuffling transition would be useful to give visual feedback when things move instead of just having them appear at their new index.

I think that proposing merge-by-default behavior for set on arrays will probably happen for 0.9. Maybe update too.

Issue #2668, opened 2016-09-26
0.8 Edge: Observers not firing correctly inside decorators


kouts:
Observers inside decorators are not firing correctly.
This [fiddle](http://jsfiddle.net/yvhhp2n6/) uses the latest edge, you can press the buttons and see that some values are not formatted correctly (you can press the buttons more than once to make it more obvious).
This [fiddle](http://jsfiddle.net/mqmzny8s/2/) uses the v0.8.0-edge-c0fd077dcbf14872da9f3a84dbc07a6b599e2118 commit which updates correctly.

evs-chris:
The fiddle mostly works now, but note that formatting values while typing in a non-lazy bound field is not recommended.

kouts:
thanks @evs-chris , I will try it using the "lazy" attribute.

PR #2669, opened 2016-09-26
Fix another observers modifying the observed value issue - #2668

evs-chris:
This sets observers to use their current model value, if available, as the value for future change comparison instead of the model value going into the observer dispatch. The net effect is the observers that change their observed value keep the fresh value cached, whereas current edge gets out of sync because observers can't cause themselves to fire again.

That all sounds right to me, and the test suite is ok with it. Have I missed a corner case somewhere?

#2668

**Is breaking:**
No

**Reviewers:**
@martypdx @kouts


evs-chris:
Turns out a dirty build env caused my tests to turn green when they shouldn't have. I force pushed a fix for the test derp.

(also, thanks @fskreuz)

PR #2674, opened 2016-10-03
API for embedding and managing external instances (component API) - 0.9 - RFC/Review

evs-chris:

One of the last big things that I miss in Ractive is the ability to manage components from outside of the template. This is the port of my attempt to make that possible (#2595) to current edge. After reviewing the open issues, I don't think I'm alone in looking for this, as there are at least 10 open issues dealing with managing components.

### What this does

This adds a few new instance methods, extends the existing linking system to be able to cross instance boundaries, and introduces a new vdom element, anchor, that can play host to foreign instances.

* **`attachChild( instance, options )`** - take the Ractive instance, that may be a component, and attach it to this instance. If a `target` option is specified, then the instance will be registered with that name. When an anchor with that name becomes available, or immediately if there already is one, the instance will be rendered into the appropriate place in the anchor. If no target is supplied, the instance will just inherit its new parent's context and be left to its own devices for rendering. A `name` can also be specified for `findComponent` purposes. If no name is supplied where a target is, then the target will be used as the name.
* **`detachChild( instance )`** - if the given instance is attached to this instance, detach it. If it is targeting an anchor and is currently rendered, then the instance will be unrendered.

`link` gets a new options argument that allows you to specify a `ractive` instance to use when resolving the target for the link. This replaces the `addMapping` and `removeMapping` from #2595.

### Anchors

Anchors can be defined in templates using a new proposed syntax `{{>>anchorName}}`, where an anchor ~~may optionally be defined as multi-tenant as `{{>>anchor multi}}` and~~ may optionally define mappings as `{{>>anchor name="{{some.path}}" foo="{{bar}}"}}`, hopefully as you would expect after having dealt with components.

~~Currently, if more than one instance is attached to the same anchor, the last one in will occupy it if the anchor is not `multi`. If the last instance is detached, then the new last one will be rendered in its place. If the anchor is `multi`, then all instances are rendered in attachment order. I have considered adding a block syntax to allow some template to be wrapped around each instance in the anchor, where block anchors would be inherently multi-tenant and the yield is the placeholder for the instance: `{{#anchor foo bar="{{baz.bat}}"}}some<wrappings>{{yield}}</wrappings>here{{/anchor}}`. I'd like to get some feedback before I attempt to address that, though.~~ Anchors will look for an attached child that has an index corresponding to their position in the template e.g. in `{{>>foo}}{{>>foo}}` the first anchor will try to grab the child targeting `foo` with index `0` and the second will try to grab index `1`. `prepend: true`, `insertAt: <some index>`, and `append: true` (the default) can be specified along with the `target` to control where in the list the child appears.

There is an instance array that is managed by Ractive as children are attached and detached that can be used to automatically create an anchor for each attached child - `{{#each @this.children.byName.foo}} ... {{>>foo}} ... {{/each}}`.

~~When an anchor is rendered, it checks with its instance to see if any children are waiting around for an anchor with the same name, and if there are, it adopts them and renders them as appropriate. When an anchor unrenders, it unrenders and releases any of its children back to its instance so that they may be picked up by another matching anchor. This allows anchors to exist in different places conditionally. I have also considered allowing multiple instances targeting the same name to be spread across multiple anchors with the same name if the anchors aren't `multi`. The last anchor would pick up the remaining instances. Again, I'd like some feedback before I attempt to address that.~~

Since the template isn't controlling the _instance_ lifecycle as it does with a component, an instance can be set up with whatever state and moved around in the template without having to be completely destroyed and recreated, which causes view state to get out of sync e.g. with a tab component resetting to the first tab when it is re-rendered and other similar cases involving private component state.

### Extra

This fixes a few subtle corner-case bugs with `find`, `findAll`, `findComponent`, and `findAllComponents` involving live queries. All of those methods also behave as you'd expect when a foreign instance is rendered into an anchor. Additionally, there is a `remote` option added to those methods that allows non-anchored instances to be considered for the returned results.

Events also behave as you would expect from a component. The event namespace is determined by the name supplied during attachment or the anchor name if no name was given.

## Questions:

From above (A):

1. Should there be a block-style anchor definition to handle additional template for each instance? **Update** there's an alternative now implemented.
2. If so, should the current singular anchor exist as well, or should it be replaced with an empty block? **Update** only the singular is required with current implementation.
2. Is there a better singular anchor format than `{{>>anchor}}`. **Update** I like `{{&anchor}}`, but it's already an alternative for triples `{{{ ... }}}`.
2. I think multi-tenant anchors are great, but I have some weird components that benefit from them. Should they exist? **Update** they don't now, but there is, what I consider, a better alternative implemented.
3. Should non-multi anchors act as they do where the last attached is rendered, should it be first rendered, should it be an error/warning, or should it be an option? **Update** they are one-for-one with no warning for missing a host anchor currently.
4. Should same-named anchors allow distribution of multiple instances targeting the same name? **Update** this is how the current implementation works.

Additional (B):

5. There isn't currently an API to automatically handle mappings when attaching/detaching an instance. Currently only the template defined mappings are automatic and any programmatic ones need to be handled manually with link/unlink. Should there be a `mappings` option with `attachChild`?
6. Have I missed anything obvious?

I had a few more, but I forgot them while typing out this wall of text... sorry.

I think this would take care of most of the issues and feature requests dealing with manually rendering/unrendering components or trying to modify them after they're rendered.

## Is breaking:

No, and it's targeting 0.9.

## Reviewers:

Yes, please


skeptic35:
Very interesting! I'd have a couple of use cases for this, so it generally gets a thumbs up from me. As for the questions, I think a couple of them (1, 2 and 6) depend on how the anchor would/should behave inside a loop.

```
{{ #some_array }}
  {{ >>some_anchor }}
{{ /some_array }}
```

In order for the above to work (and I'd definitely expect it to work) it would, as far as I understood it, be necessary to answer question 6 with "yes". Otherwise it would not be possible to use an anchor inside a loop and in that case we would definitely need some sort of block syntax. Personally I'd prefer the above loop over a block-style anchor (I never understood ```{{yield}}``` anyway ;-) ).

Question 3: I hope there is a better singular anchor format than ```{{>>anchor}}``` because that one is too easily confused with the partial syntax. I think my favourite would be ```{{&anchor}}``` 

Question 4: Multi-tenant anchors sound great, provided there would be some way to re-order it's instances, without having to detach an re-attach all of them.

Question 5: It's tough to decide what would be the more useful behaviour. I'd have use cases for both, which means it should probably be an option with "last attached is rendered" being the default.

That's my two cents so far. There probably is more to say once I've had time to give this more thought.

evs-chris:
@skeptic35 yep, I agree with most of that, so I've adjusted the PR to match. Thanks for perspective!

First off, unfortunately, `{{& ... }}` is already taken as the equivalent of `{{{ ... }}}`. I also considered some unicode characters, like the anchor :smile:, but figured that would make me a terrible person. I doubt that it is used much (don't even think it's in the docs), so this could take over that syntax as a breaking change. Anyone have any thoughts on that?

I like the idea of multi-tenancy, but it really needs to be able to have wrapping template injected with each tenant. The `{{#each ...}}` case should also work. To that end, multiple anchors can now exist with the same name, and they will pick up an attached child based on their position in the template and the child's position in the attached list. There is also a Ractive-managed children.byName object that updates as children are added and removed, so you can use `{{#each @this.children.byName.foo}}<div>{{>>foo}}</div>{{/each}}` to have an iterative section that automatically has enough anchors for the attached children, in this case, targeting `foo`.

Order of attachment now matters, since instances are distributed based on anchor location in the template, so you can specify `prepend: true`, `insertAt: <some number index>`, or `append: true` (the default) to specify where in the list of attached children for an anchor name that the new instance should appear.

There's no way to move an instance without detaching and re-attaching it right now. Of course Ractive has the same general problem with lists of things in regular data, as there's a splice for adding and removing stuff around the same index, but no way to say "swap index 5 with index 9, index 9 with index 1, and index 1 with index 5". I think an API for that would be a good thing and that the attached child version should match.

Jeff17Robbins:
This approach looks very appealing, and could replace needing to use a Ractive-Ractive adaptor.  If I have a "view" Ractive instance, and want to attach one ore more data-only "store" Ractive instances to it, is my understanding correct that I can simply do, e.g., this?

```
// note "full" instance with element, template, and data
var view = new Ractive({el: ..., template: ..., data: ...});
var storeA = new Ractive({data: ...}); // note data only
var storeB = new Ractive({data: ...}); // note data only

view.attachChild(storeA);
view.attachChild(storeB);
```

Or is the idea that I would use `link(...)` to attach the stores to the view?

If the first case (`attachChild(...)`), how do I refer to the data of a store?  Say `storeA` has `fieldA` as a property, how do I `get(...)` `fieldA`?

```
view.get( <== what keypath do I use here?
```

evs-chris:
`attachChild` aims to make sharing a common data store (parent) among many children possible, where parent paths can be mailed into the child like existing component mappings. The template in the children will resolve using the parent context as it would if were a regular component, assuming it's not isolated. Anchors take it a step further and let the parent template manage the render cycle too.

To achieve something like ractive-ractive, you would just need the enhanced `link` that lets you specify the target instance. Using that, you could link any keypath in any instance to a local keypath. There isn't in-depth cycle detection, so you'd have to be careful with really complex structures.

Jeff17Robbins:
So using my example (1 view, 2 stores), I would do something like this?

```
view.link('StoreA', storeA);
```

In other words, I'm assuming I can link the first arg (a keypath in `view`'s viewmodel) to the viewmodel of a completely different Ractive instance `storeA`.  Is that how it would work?

evs-chris:
It would be `parent.link('', 'StoreA', { ractive: storeA })`. The params on link try to follow `ln` syntax - src path, dest path, options. Then `parent.get('StoreA.some.path')` would be equivalent to `storeA.get('some.path')` and all of the internal tracking that ractive uses to keep things up to date works as you would expec.

Jeff17Robbins:
Thanks.  I see that I got it nearly perfectly wrong.   :-)

Is "parent", for example, my "view" Ractive instance?

evs-chris:
Yes, it would be your root instance, so `view.link('', 'StoreA', { ractive: storeA })`.

Jeff17Robbins:
Am I correct in assuming that this new feature will be a lot more efficient
than propagating every 'change' event from `storeA` to `view` via
`.update(...)`?

evs-chris:
Way more efficient than change events and updates. Only slightly less efficient than only having a single instance, because the link has to be maintained, but it's the exact same mechanism as component mappings in 0.8 if you want to check performance.

Jeff17Robbins:
If / when you have a built ractive.js for me to test with this feature,
happy to help (if that's helpful!)

We don't use Ractive components, because we are using webcomponents (and
the webcomponents-lite.js polyfill) to componentize things.  Which makes me
feel sometimes that we're going "cross grain" with Ractive's emphasis on
its own components.  But Ractive instances are most excellent, and we love
using them as the majority of our webcomponents' "inner workings".  Gaining
efficient linking across Ractive instances will be great!

evs-chris:
I have a branch on this that switches from `{{>> anchorName mapping="{{some.path}}"}}` to `{{#anchor anchorName mapping="{{some.path}}" /}}` and also allows `{{#anchor anchorName mapping="{{some.path}}"}}this is in the content partial{{#partial name}}this is the name partial{{/partial}}{{/anchor}}` so that inline content can be provided to the external instance. That would bring the external API to feature parity with inline components, but it doesn't quite seem to fit. Is there ever a case where you'd _need_ to provide inline partials to an external instance?

paulocoghi:
This makes things far easier. Excellent approach, @evs-chris !

Before your API, my only option was to compose a template string, adding necessary dynamic components to it, and use `resetTemplate( my_new_template_string )` in the root instance.

Since I let the developer inform which component areas he/she will use, I can't predict the template in the beginning. And the "resetTemplate approach" was mandatory to allow this flexibility.

Now, with this new API, I can attach/detach  new components to the root instance without resetting anything, in a more polite manner.

Perfect!

--

I will surely try to replace my old approach with this one in [altiva/altiva](https://github.com/Altiva/altiva) , and inform my experience!

evs-chris:
@paulocoghi being able to create a more generic routing framework based on Ractive was one of my goals with this :smile:.

Do you have any opinion on the anchor syntax and whether or not to include mappings in `attachChild` as an option? I'm starting to lean toward `{{#anchor anchorName ... /}}` as the initial syntax to allow the block syntax at some point in the future.

I'm also thinking about starting a 0.9 "next" branch so that I can publish some alpha builds and maybe get a little more feedback more easily.

kouts:
Great feature, does this mean we' ll be able to render - unrender components programmatically on-the-fly?
I could use it for popups instead of having the "hidden", it would be nice to have them inited and removed on demand.
Why not use the component syntax with an added attribute e.g <componentname is-anchor /> ?

evs-chris:
@kouts yes, you can render any old ractive instance into an anchor. If it's already rendered elsewhere, it will be unrendered first and then rendered inline at the anchor.

Re-using component syntax is an interesting proposition... I'll have to ponder it for a bit. Off-hand, I don't hate `<#anchorName ... />` or `<@anchorName ... />`. I don't know that an attribute stands out quite enough...

paulocoghi:
@evs-chris What you already have done solved 99% of the challenges.

---

**About mapping**

If I understand correctly: mapping would help to define a context to the component attached to an anchor, without the need to use link. If this is the case, it will be useful for a lot of people.

In my case I solved this in Altiva and it's not a must-have feature for us. I just don't did a pull request of this feature to Ractive because I think my work is not as advanced and well made as Ractive is.

---

**About link**

The link feature (again, if I understand correctly) would enable us to link specific data of a component to specific data on another component, right?

If this is the case, this will help us a lot!

**Example with `link` and `alias`**

One case is in our Ractive DB project (draft/alpha), that create a nosql O(1) / O(N) database using Ractive internal data, allowing indexing of desired fields, access in mustache and inside component logic, etc

Link will help our database create the relationships **inside Ractive**:

```html
<!-- An order is linked by Ractive DB with its user, through `user_id`: -->
{{#with db.order[ order_id ].user_id as user }}
        {{.name}} <!-- This resolves to db.user.name, awesome! -->
{{/with}}
```

---

Also, thank you so much for the beautiful work you have done!

(and sorry for my bad English)

evs-chris:
With that last set of commits, I have merged anchors into components, as it turns out that's basically what they are - a component for which you bring your own instance. The syntax for anchors has changed to `<#anchorName mapping="{{some.mapping}}" on-whatever="yep">there can be content{{#partial foo}}and there can be inline partials{{/partial}}</anchorName>`. The minimal example would be `<#anchorName />`, which is actually shorter than `{{>>anchorname}}`, so there's that :smile:. I like that `<#whatever />` looks like a component, but also kinda stands out as a target, similar to html anchors. I'm also open to suggestions on syntax.

I don't have any tests for events yet, because I kinda stumbled onto support for those when doing the merge. It made sense to have automatic event subscription on top of an anchor though, so I need to get some tests set up around that.

While this is a pretty heavy PR, merging anchors into components did drop the net increase in code to <600 lines - or a little over 7k on the minified build. Since this is a pretty big feature that not too many people have eyes on, I would like to land in 0.9 as experimental. That way I might be able to get a little bit of feedback from a wider audience before calling it stable. It has already changed a bunch just in this version of the PR.

@paulocoghi yes, cross instance linking basically makes data from one instance accessible from another while keeping all of the change propagation intact in both directions. You have to be a little careful with links though, since they may not move around like you'd think they would. They're basically an overlay on top of your data and not in your data directly, so if you have data `{ foo: { bar: 'baz' } }`, you link `foo.bar` to somewhere else, and then reset your data to `{ foo: { nope: true } }`, `foo.bar` will still point to the other end of the link because the link is still in place.

kouts:
@evs-chris Is it possible to "append" a component on an Ractive instance without using an anchor previously defined on the instance's template?
A jsfiddle would be useful as an example of this new features.

evs-chris:
@kouts there is. You can use a list that you push to to create a new anchor, or you can use `@this.children.byName.anchorName`, which is maintained specifically for that purpose. https://jsfiddle.net/2w20q1x5/1/

Until the docs get fleshed out, there are also a few examples in the tests at https://github.com/ractivejs/ractive/blob/dev/test/browser-tests/methods/attachChild.js

PR #2681, opened 2016-10-13
Update deps

fskreuz:

Update deps, because the deps shield is red.

**Reviewers:**

Everyone.


evs-chris:
I tried about 6 weeks ago and gave up after an hour or so of hair-pulling. I'd say try 'em one at a time to avoid the massive pile of fail. I think we may need to stick with an older gobble to avoid the fs watch terribleness with native modules.

.travis.yml
```diff
@@ -4,8 +4,6 @@ node_js:
   - '6'
   - '5'
   - '4'
-  - '0.12'
```
evs-chris: Every time I change one of the node tests, I forget that they aren't transpiled, so the CI fails on node < 4. So a bit :+1: on this :smile:

evs-chris:
Good to go?

fskreuz:
Yep. This will do for now.

Determined that updating Gobble and/or Rollup causes issues. Will
investigate later.

fskreuz:
The move was just in time for the news
https://medium.com/@nodejs/node-js-v6-transitions-to-lts-be7f18c17159 :D

PR #2685, opened 2016-10-13
Update and configure ESLint

fskreuz:

Configure ESLint for Ractive.

PR revised. This will only contain the ESLint configuration. Actual linting will sit in another PR.

**Is breaking:**

No.

**Reviewers:**

Everyone.


evs-chris:
Great! :+1:

I'd like to request not failing on lint or fixing all of the lint yet, as I have a few sweeping changes in progress that would not be fun to resolve conflicts on :smile:

fskreuz:
Yep. The linting on build was commented out. Just threw that in there to see how it fails both CI. I can remove it later. 😄

fskreuz:
@evs-chris Updated PR to only include configs. I'll put the linting in another PR.

evs-chris:
Many kind thanks!

May I suggest turning the linting back on but surrounding that section with 
```sh
set +e
# linting here
set -e
```
, so that we can see linting whilst building and go ahead sweep up lint where we're working? Then when things calm down a bit we can swiffer it good and make lint fail the build again.

evs-chris:
Ready to merge?

fskreuz:
@evs-chris 🚢

Issue #2431, opened 2016-03-10
on-click="resetPartial(...)" not working


dagnelies:
both latest/edge:

http://jsfiddle.net/xc2o8y08/42/

calling `resetPartial` in the code instead of the button works. Likewise, removing databound items from the partials also make it work in the button.

I guess I see a lot of things because I used ractive rather heavily in a production application.

martypdx:
Did a bit of research, it's the mustache `{{` in the on-click method expression that's failing at _parse_time.

evs-chris:
I think I can solve this and issues with nested quotes in directives if/once #2386 lands. Instead of grabbing everything in the attribute and parsing it again based on the attribute type and what the content looks like, it can do proper parsing of the attribute/directive values. It could technically be done now, but I don't think it would be pretty because of the legacy syntax for proxy events, decorators, and transitions.

evs-chris:
This is now fixed in edge and available in `0.9.0-build-1`.

PR #2687, opened 2016-10-14
BREAKING: Clean up code and remove deprecations - WIP/Review - 0.9

evs-chris:

* [x] Remove old-style event, decorator, and transition syntax
* [x] Adjust directive parsing to use the normal expression parser, so that parsing attributes is a bit more robust.

    For instance, this is now possible: `<button on-click="@this.resetPartial("foo", "the new partial <b>with tags</b> and <span data-foo=attributes>inside</span> an attribute" + someRef)">yep</button>`
* [x] Remove partial comments support (`<!-- {{>somePartial}} --> ... <!-- {{/somePartial}} -->`)
* [x] Remove deprecated event object properties
* [ ] Clean up event context helpers/merge more code with regular prototype methods if possible

After converting the tests for the proxy event with args removal, I'm thinking that a fire helper for replacing `<button on-click="eventName:arg1, arg2">`, which now becomes `<button on-click="@this.fire('eventName', event, arg1, arg2)">`, with `<button on-click="event.fire('eventName', arg1, arg2)">`. Note that depending on what the arguments actually are, the proxy version may come up with a completely weird single or multiple args when it actually fires due to proxy event parsing jankiness. Adding that `fire` to the event object would also be a tiny addition.

#2431 

**Is breaking:**
Yep. The template version will need to be bumped again because the directive args got significantly simplified for most cases.

**Reviewers:**
Yes, please.


fskreuz:
`+315 −1,080`

😱 some serious housekeeping going on here.

src/parse/converters/element/readAttribute.js
```diff
@@ -197,57 +189,60 @@ export function readAttributeOrDirective ( parser ) {
...
-				parser.error( 'Cannot use reserved event names (change, reset, teardown, update, construct, config, init, render, unrender, detach, insert)' );
+			// check for a proxy event
+			if ( !readProxyEvent( parser, attribute ) ) {
+				// otherwise, it's an expression
+				readArguments( parser, attribute, true );
+			} else {
+				if ( reservedEventNames.test( attribute.f ) ) {
+					parser.pos -= attribute.f.length;
+					parser.error( 'Cannot use reserved event names (change, reset, teardown, update, construct, config, init, render, unrender, detach, insert)' );
+				}
```
fskreuz: Collapse to `if-else if`?

```
if ( !readProxyEvent( parser, attribute ) ) {
  ...
} else if ( reservedEventNames.test( attribute.f ) ) {
  ...
}
```

src/parse/converters/element/readAttribute.js
```diff
@@ -197,57 +189,60 @@ export function readAttributeOrDirective ( parser ) {
...
 		}
 
 		else {
+			parser.allowWhitespace();
+			if ( parser.nextChar() === '=' ) {
+				const value = readAttributeValue( parser );
+				if ( value != null ) { // not null/undefined
+					attribute.f = value;
+				}
+			}
```
fskreuz:
```
let value = parser.nextChar() === '=' ? readAttributeValue( parser ) : null;
attribute.f = value !== null ? value : attribute.f;
```

src/parse/converters/element/readAttribute.js
evs-chris: Derp :smile:

Issue #1848, opened 2015-03-18
feature request for improved consistency, elegance, and readability


brandondrew:
When I read Ractive code, I find that everything is easy to read and easy to understand.  This is very important to me, and I believe it is very valuable for my productivity.  (Not just mine&mdash;I believe it's valuable to everyone's productivity really, although lots of developers don't seem to realize this, and continue to use obscure or meaningless names for functions and variables and fields in the database). 

When I've worked with Angular I've always felt a little like vomiting in my mouth over all the little "ng-" turds littering my code.  Of course there's much to like about Angular, and I may still use it in the future as appropriate, but so many things attract me much more strongly to Ractive for any cases where I don't have a compelling reason to choose something else.

One reason I find Ractive code easy to understand is that it relies as much as possible on plain English words, rather than making up new and obscure conventions.  To specify a template I use the word... (drum roll, please!) "template"!  When I need to list partials I use... "partials".  When I want to refer to a source of data by variable name, or pass in data as a javascript object, I use... "data".  These are the most obvious and intuitive things possible.  At least if there's anything else comparable I can't think what it would be.  (I suppose you could use "layout" for the full-page template, and "templates" for partial templates, or something like that, but the Ractive way seems at least marginally clearer to my Rails-soaked thinking, and I'd be happy with either.)

When I want to do something on teardown, I use (no big surprise here) "onteardown" in my JavaScript.  The only thing I have to remember is that there is no underscore or camelCase or hyphen (the last of which would be impossible in the given context anyway).  That's not hard to remember, mainly because it's consistent (as far as I can see) throughout Ractive.  (Well, there's "on-teardown" as an attribute in templates and partials, but that's a different context and it makes sense to use a hyphen there to disambiguate between, for instance, "onclick" and "on-click".  So it's still logical and essentially consistent.)

[...]

PR #2696, opened 2016-10-17
Allow target as a config alias for el - #1848

evs-chris:
This allows a target element to be specified at init with `target` as well as `el` for consistency sake. We don't use abbreviations anywhere else, and this brings in a way to set el that matches the rest of the config params.

#1848

**Is breaking:**
Maybe, if you have `target` as a property on a component or instance.

### DIFF
--- a/src/Ractive/initialise.js
+++ b/src/Ractive/initialise.js
@@ -49,7 +49,7 @@ export default function initialise ( ractive, userOptions, options ) {
 
 	if ( fragment ) {
 		// render automatically ( if `el` is specified )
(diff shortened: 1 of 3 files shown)


aharpervc:
I don't get it. Why is this better? If you don't want an abbreviation, why not `element`?

evs-chris:
There's a bit of discussion in #1848, but mostly because it tracks better with a selector being supplied as opposed to an actual element, which is the less common case. I also preferred target because it tracks with the target attribute on anchors and other planned APIs in Ractive :smile:

PR #2700, opened 2016-10-18
 BREAKING: More capable yield based on partial - RFC - 0.9

evs-chris:
This is my attempt to touch up partials and yielders for 0.9. Here's what it does:

* [x] Allow `name` in `{{yield name}}` to be an expression if there is no appropriate `name` partial.
* [x] Allow non-inline partials to be yielded so that generated content can be yielded.
* [x] Allow multiple yields with the same name, including the implicit `content` partial. See next.
* [x] Allow aliases to be specified at the point of yield e.g. `{{yield tpl with foo as bar, baz as baz}}`. I think this solves the yield context problem around iteration where the implicit forced parent context of the yielder is already beyond the component that is yielding. It does so by making the aliases available just before the yield boundary, so that the context remains intact.
* [x] Make partial context only apply _inside_ the partial. Currently, the context can interfere with partial expressions e.g. `{{>foo { foo: 'name' } }}` will make `foo` resolve to the `name` partial regardless of what `foo` is in the surrounding context.

### Valid partial-ish syntax - []s are optional
`{{yield [name/expression][ with ${ref/expr} as ${alias}[, ... more aliases] ]}}`
`{{> name/expression [( ${ref/expr} as ${alias}[, ... more aliases] ) || ${context expression}] }}`

Between the expressions and allowed iteration, it should now be possible to make more complex classes of component without hackishness. I'm specifically thinking of the sins I've committed whilst building a tab component that could add and remove tabs dynamically. Grids/tables would also be a bit less painful to handle.

### Code merger
Since the yielder is now basically just a specially contexted partial, this also merges the partial and yield parsers and vdom items.

Also, there was a bunch of repeated code around the various vdom items that manage a fragment, so I bundled that up into `ItemContainer` and `MustacheContainer` classes.

### TODO:
* [ ] More tests.
* [ ] Move more common code into container classes.

#1833 and probably #2206

## Is breaking:
Yes, in a few different ways.
1. Template version bump is needed.
2. Yielders can now yield any old partial in addition to inlines. It's no longer an error if you forget your inline to yield.
3. Multiple yields of the same (implicit) name are allowed.
4. Partial context no longer applies to the expression

## Reviewers:
Please :smile:


evs-chris:
@JonDum do you think this could solve your paginator case in #2206?

To expand a bit on why yields were only allowed once per target, a yield has always basically been just a specific form of partials. It always has a built-in context that is the same as the context for the parent of the component, because that's the only context that the parent could possibly know about. Since the context _must_ be the same as the parent of the component, there's not really a way to inject a value without tripping over that requirement. That's where aliases come in.

With aliases, a component can provide references to the partial and not interfere with the immediate context. It does basically require the references to be known ahead of time by the parent of the component i.e. part of the component API. Though you can get around that by generating part of the template when the component is constructed.

I'm pretty terrible at explaining things coherently, so here's an example :smile::

```js
const List = Ractive.extend({
  template: '<ul>{{#each items}}<li>{{yield with . as item}}</li>{{/each}}</ul>'
});
const r = new Ractive({
  el: '#someel',
  template: '<List items="{{list}}">{{item.name}} {{item.price}} <button on-click="@this.add(item)">Add</button></List>',
  data: {
    list: [
      { name: 'Spanner', price: 123.45 },
      { name: 'Widget', price: 9182.12 }
    ]
  },
  components: { List },
  add(item) { doSomethingWithTheItem(); }
});
```

Here's the dynamically aliased version:
```js
const List = Ractive.extend({
  template: '<ul>{{#each items}}<li>{{>adjustedYield}}</li>{{/each}}</ul>',
  partials: {
    adjustedYield: ''
  },
  onrender() {
    this.resetPartial('adjustedYield', `{{yield with . as ${this.get('item')}}}`);
  }
});
const r = new Ractive({
  el: '#someel',
  template: '<List items="{{list}}" item="foozle">{{foozle.name}} {{foozle.price}} <button on-click="@this.add(foozle)">Add</button></List>',
  data: {
    list: [
      { name: 'Spanner', price: 123.45 },
      { name: 'Widget', price: 9182.12 }
    ]
  },
  components: { List },
  add(foozle) { doSomethingWithTheFoozle(); }
});
```

I think that in the vast majority of cases, having the aliases as part of the component API is perfectly acceptable.

zuck:
IMHO:

- [ ] Improve docs on template, partial and yielding relationship and use.

JonDum:
1) This is super freaking cool. 👍 
2) I think I ended up using the lovely undocumented `{{> content}}` hackarooney to make `ractive-paginator` work. I haven't looked at this component in like a year, but it appears to still be working on edge. No idea if anyone actually uses it soooo `¯\_(ツ)_/¯`

Issue #2701, opened 2016-10-19
0.8.0: DEBUG mode console output & Ractive.parse


timoxley:
Since `0.8.0`, `Ractive.parse` generates a lot of noisy console output by default.

This is problematic when `Ractive.parse` is being used in node within a build tool e.g.

https://github.com/rstacruz/ractive-loader/pull/5

Yes, it's a simple fix to turn DEBUG mode off, but I'd suggest that Ractive was a better behaved API in `0.7.3`. In the browser, I can agree that `DEBUG=true` by default is fine, however when Ractive APIs are used in any other environment I'd suggest that having verbose logging on by default is generally considered poor behaviour.

timoxley:
Note the 0.7.3 behaviour may have been less of a browser vs node thing and more a `Ractive()` vs `Ractive.parse()` thing. While turning debug logging off for `Ractive.parse` would resolve my immediate grievance, I believe it makes sense that Ractive should be quiet by default everywhere but the browser and even there, I'd be ok if debug mode were opt-in like most other tools.

fskreuz:
Hmm... This sounds more like a tooling issue rather than Ractive itself. Ractive is not aware how it is being used by its dependents. So it might make sense for the dependent (in this case, the tool) to set the conditions instead of Ractive.

Instead of making Ractive do the decision, either:

- Have the tool turn if off,
- or have the tool suppress logging without completely turning off the option globally,
- or have the tool support turning off the errors via an option.
 
Prior to `0.7`, `debug` was by default `false`. There's a longer discussion about why `debug` was set to `true` in https://github.com/ractivejs/ractive/issues/1793

evs-chris:
I could see where that would be quite painful parsing a large number of templates. How about a global option - something like `Ractive.WelcomeMessage` - that defaults to on that can be turned off by tools. That way you can still get deprecation and other warnings from your tooling? I'll have a PR shortly that includes some other useful debug stuff for parsing.

Issue #2553, opened 2016-05-30
Unqualified method events are deprecated. Prefix methods with '@this.' to call methods on the current Ractive instance.


dagnelies:
Hi,

it's nice to provide a warning, but it would be even nicer to write "where" this happens. After all, this is a change impacting lots of users/code/plugins.

kouts:
What does this warning means exactly?
Does it mean that we must prefix '@this.' to all methods inside templates?

So instead:
``` <button on-click="myMethod()">Call my method</button> ```
We should now write:
``` <button on-click="@this.myMethod()">Call my method</button> ```

Is this the case?

**Comment by evs-chris** [comment]
That's correct, because you can now use any valid expression that includes a call in an event handler, not just a simple `methodname(...args)`. For instance, you can call data methods and multiple functions as a comma separated list e.g. `<button on-click=".bar.foo(10, ../something), @this.parent.fire('someEvent')">Click Me</button>`, which would call `foo` on the `bar` object in the current context and then fire a 'someEvent' on the parent instance when the button is clicked.

**Comment by kouts** [comment]
Ok I understand, It would be great if the '@this' part was optional though, at least for single method calls. To me it seems that -at least in my templates- it will just add additional clutter.
But that's just me...

**Comment by fskreuz** [comment]
`@this` was implemented to tell what's from the data and what's from the instance. Without `@this`, how would you know which `foo` is used where?

```
<button on-click="foo('bar')">{{ foo }}</button>`.
```

In the older versions this was easy. `foo` in the event would be a method while `foo` in the mustache would be something in the data. In the upcoming version, where it appears that both can be used, there must be a way to tell which is which.

**Comment by JonDum** [comment]
>there must be a way to tell which is which.

Isn't that already accomplished with the `.` prefix? Now that I think about, the `@this` is redundant if you have that, no?

```html
<button on-click=".bar.foo(10, ../something)">Click Me</button> // access method on data model

<button on-click="@this.parent.fire('someEvent')">Click Me</button> // use @this to access parent 

<button on-click="fire('someEvent')">Click Me</button> // no dot, assume it's @this
```

fskreuz:
`.` and `this` (no `@`) refers to the current data context. `@this` (has `@`) refers to the Ractive instance.

```
const Component({
  ...
  data: {
    foo: function(){/* Data function. */}
  },
  foo: function(){/* instance method */}
});
```


[...]

Issue #2146, opened 2015-09-14
Feature request: more lines output for errors


matthew-dean:
Hi, I've run into this a number or times so I'd thought I'd bring it up.

I currently have a parsing error that looks like this:

```
Uncaught ParseError: Unexpected closing tag at line 164 character 25:
                        </label>
                        ^----
```
However Ractive counts lines, whether it's by counting from the start of the inline template, or somehow using the position in the HTML doc, the line # never amounts to something meaningful / matching to my template.

Rather than focus on the line number issue (although getting insight on that would be nice too), what would help me is if the error output included more than just one line. I have LOTS of </label> closing tags in this particular template, so I end up visually inspecting all of them to try to find a parsing error.

If, however, the error output simply included more than one line, like two lines preceding the error, I would probably be able to find that easily in my code.

This has actually come up a lot for me, where I get an error, but the error information doesn't actually provide anything I can use to find the source of the problem. Is a little more verbosity possible?

matthew-dean:
I was able to figure out after some mathing that the line of the actual error was the position of the template in the HTML doc + the line of the error. Which makes sense, I suppose, but I still think having more lines would still help, since the math may not always be that simple, depending on how the template is imported.

Rich-Harris:
Agree, this is particularly confusing in the templates-in-script-tags case. Perhaps we should have special treatment for those:

```
Uncaught ParseError: Unexpected closing tag at line 164 (relative to
<script id='template' type='text/html'>) character 25:
                        </label>
                        ^----
```

A couple of lines of padding either side wouldn't hurt either.

PR #2705, opened 2016-10-19
Various debug logging changes - RFC - 0.8.1

evs-chris:
This does a few things:

1. Add a new parsing option `contextLines`, that when set to:
  * `-1` - only shows the message
  * `0` - the default - shows the line with the error and the ascii arrow thing beneath at the appropriate character
  * any positive integer `num` - get at most `num` lines from above and below the relevant line and wrap them around the line with the ascii arrow thing
2. Makes method call event deprecation stand out more by using the context helper from 1 above to throw out a message for each instance that needs to be updated. This will get really chatty, but un-prefixed methods are most likely going away in 0.9 entirely. It's kinda hard to find these because you can't just search for `decorator` or `transition` like the other deprecated things.
3. Makes proxy events with args warn for each instance, again with context like above. As with method events, it's harder to find instances of proxy events that are deprecated going forward.
4. Adds a new global `Ractive.WELCOME_MESSAGE` that can be used by tooling to skip or modify the welcome message with values:
  * `false` - just don't
  * some string `str` - display `str` instead of the default message. If there is group support on the `console`, it will still be collapsed when logged.

If there are no objections or issues, I'll land this tomorrow and push out 0.8.1, which already has a number of bug fixes queued.

#2553 #2146 #2701

## Is breaking:
No, but it does take over `contextLines` if some instance happens to be using that as a flag somewhere in the wild.

## Reviewers:
@timoxley @dagnelies @fskreuz @MartinKolarik @martypdx @Rich-Harris and anyone else

### DIFF
--- a/src/Ractive/config/defaults.js
+++ b/src/Ractive/config/defaults.js
@@ -16,6 +16,7 @@ export default {
 	preserveWhitespace:     false,
 	sanitize:               false,
 	stripComments:          true,
+	contextLines:           0,
 
 	// data & binding:
 	data:                   {},
--- a/src/Ractive/config/runtime-parser.js
+++ b/src/Ractive/config/runtime-parser.js
@@ -13,7 +13,8 @@ const parseOptions = [
 	'interpolate',
 	'preserveWhitespace',
 	'sanitize',
-	'stripComments'
+	'stripComments',
+	'contextLines'
 ];
 
 const TEMPLATE_INSTRUCTIONS = `Either preparse or use a ractive runtime source that includes the parser. `;
--- a/src/parse/Parser.js
+++ b/src/parse/Parser.js
@@ -62,6 +62,40 @@ Parser.prototype = {
 		return null;
 	},
 
+	getContextMessage ( pos, message ) {
+		const [ lineNum, columnNum ] = this.getLinePos( pos );
+		if ( this.options.contextLines === -1 ) {
+			return [ lineNum, columnNum, `${message} at line ${lineNum} character ${columnNum}` ];
+		}
+
+		const line = this.lines[ lineNum - 1 ];
+
+		let contextUp = '';
+		let contextDown = '';
+		if ( this.options.contextLines ) {
+			const start = lineNum - 1 - this.options.contextLines < 0 ? 0 : lineNum - 1 - this.options.contextLines;
+			contextUp = this.lines.slice( start, lineNum - 1 - start ).join( '\n' ).replace( /\t/g, '  ' );
+			contextDown = this.lines.slice( lineNum, lineNum + this.options.contextLines ).join( '\n' ).replace( /\t/g, '  ' );
+			if ( contextUp ) {
+				contextUp += '\n';
+			}
+			if ( contextDown ) {
+				contextDown = '\n' + contextDown;
+			}
+		}
+
+		let numTabs = 0;
+		const annotation = contextUp + line.replace( /\t/g, ( match, char ) => {
+			if ( char < columnNum ) {
+				numTabs += 1;
+			}
+
+			return '  ';
+		}) + '\n' + new Array( columnNum + numTabs ).join( ' ' ) + '^----' + contextDown;
+
+		return [ lineNum, columnNum, `${message} at line ${lineNum} character ${columnNum}:\n${annotation}` ];
+	},
+
 	getLinePos: function ( char ) {
 		var lineNum = 0, lineStart = 0, columnNum;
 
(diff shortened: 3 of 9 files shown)


JonDum:
> un-prefixed methods are most likely going away in 0.9 entirely
 
:( :( :(

I feel like #2553 was ended prematurely. 

I'm still in the camp that thinks having no `@this` prefix at all should still work exactly as it is now, just make the warning shut up. 

`<button on-click="myMethod()">Call my method</button>`

Is just a 1000x cleaner and easier to grasp for newcomers than

`<button on-click="@this.myMethod()">Call my method</button>`

@fskreuz @dagnelies @Madgvox @MartinKolarik @martypdx @Rich-Harris Any last opinion on this?

paulocoghi:
My opinion in this is about stability and performance.

If prefixed methods present better results and operations, I accept them perfectly, although I agree that unprefixed methods are a little cleaner.

fskreuz:
@JonDum a good way to look at it is:

- Anything unprefixed (without an `@something`) refers to data.
- Anything prefixed (i.e. `@this`, `@global`, etc.) keys off from something that's not data.

😄

evs-chris:
@JonDum @paulocoghi @fskreuz I've opened #2720 with the `@` short form that would be a bit less ugly for events e.g. `<button on-click="@myMethod()">Call my method</button>`. Is that any more palatable?

paulocoghi:
> Is that any more palatable?

Sure it is! Two birds in one stone.

JonDum:
@fskreuz Not exactly how I had originally thought it worked:

* `@something` (i.e. `@this`, `@global`, etc.) are special references outside of `data`.
* `.`, `../`, and `~/` refer *explicitly* to `data`
* Anything un-prefixed (without an `@something` **or** `../`, `~/`) goes through the resolve ambiguous ref that ractive has had. If there's a `myMethod()` on the prototype call that, if not check the keypath at that level. Still nothing? Walk up or fail. 

Doesn't that solve all the problems while still allowing for the clean, un-prefixed, not confusing `on-click='foo()'`?

evs-chris:
It would. However, as long as we allow ambiguity in templates, introducing another context for resolving ambiguous references would probably lead to tears. Checking `@this` when resolving and ambiguous reference would effectively add a context parallel to the usual data context, which would mean that something like `{{name}}` could resolve to `@this.name` instead of `^^/name` in certain circumstances that would be surprising.

It's an interesting proposition, though. For the reasons above, I think it'd do more harm than good, but I'm open to it if there's significant positive feedback. I suck at usability most of the time :smile:

To be clear, I think disallowing hierarchical contextual ambiguity (`{{#with { foo: 'bar' }}}{{#with somethingWithoutAFoo}}{{foo}}{{/with}}`), would make the resolution path you suggested feel reasonable. So resolution would go:

1. It's already explicit: `.whatever`, `../../yep`, `^^/../still.yep`
2. It's special: starts with `@`
3. It's a member of the current context
4. It's an alias or special ref from somewhere up the template
5. It's a member of the current instance

The only traveling at that point would be for aliases and index/key refs. It would _require_ using aliases or `^^/` paths to get to upper contexts. It would also eliminate implicit mappings  from component templates which may or may not be a good thing. ¯\_(ツ)_/¯

JonDum:
What about for just method calls in event handlers? Just the case where if it isn't special, it checks the proto first? I may be speaking for myself here, but 95% of any method calls I do are `set()` or another ractive proto method.

I'm just not a fan of the keyword spam cluttering templates:

`<button on-click="@this.merge(@this.sort(@this.get(@keypath+'.otherStuff'))),@this.set('dirty', true)">foo</button>`

is pretty gross

evs-chris:
I'm personally not a fan of special-cased behavior in contexts that don't necessarily stand out as having special handling. If there's a change, I'd prefer the new rule to apply everywhere references are used. To that end, I suppose we could add the current instance to the end of the resolution process, so that it would never take precedent over the context hierarchy. Explicit would still be better, but most methods probably won't overlap with data. People using fat models (is that a thing in js? is there an enterprise js lurking out there somewhere with an InstanceMarshallingFactoryFactory waiting behind a bush?) also wouldn't run afoul of the resolution process.

1. It's already explicit, so resolve.
2. It's a special, cause it starts with `@`.
3. It's a member of the current context.
4. It's an alias, special ref, or member of some parent context - check until there is no more template left, mapping if necessary.
5. It's a member of the current ractive instance.

We've had complaints in the past about the cognitive overhead of the resolution process, so I'm not sure how adding _another_ place to check for values would be received. Those complaints are partially what brought this about to begin with. I suppose we _could_ just do it in event contexts, which would remove the cognitive overhead for general template use, but it would re-introduce another special case.

Maybe the best answer is to say "then don't use ambiguous refs" as an answer to the issue of cognitive overhead, and just add the instance on to the end of the process.

Also, note that using the `get`, `set`, `merge`, etc methods, which support context-relative keypaths, on the provided `event` reference are safer and usually a bit shorter than using the plain instance methods with keypath concatenation. They can also do things that are nigh impossible (or at least a big pain) by manipulating keypaths like `{{#each .some.list}}<button on-click="event.splice('../../', @index, 1)">Remove Me</button>{{/each}}` where the `each` is nested deeply in some other context, has more context within it before the button, or is in a computed context.

JonDum:
Your reasoning is sound. It's a very fine balancing act between resolution confusion and symbol confusion. Personally—if there's no major performance concern—I'd go with less symbols. In the end, I think it's your call since no one else is piping up.

I do like the `@` as a shorthand for `@this` btw. 👍 :)

kouts:
+1 @JonDum , ```@this``` works great for me, as long as there are no performance implications ```@``` as a shortcut would be a nice addition.

dagnelies:
Well, to put things concisely, I agree with @evs-chris that, especially for newcomers, it's not always easy to know "where" to put stuff. Especially `{{myFunction(...)}}` looks inside `data`
while `on-click="myMethod(...)"` looks inside root.

In order to make things more explicit, I agree with the proposal to have to be prefixed. In order to avoid clashes with `@index, @key, @keypath...`, I would suggest that people use `@` "isolated" (with a dot afterwards) to refer to the ractive instance. Ideally, it should work both for methods `on-click="@.doSomething(...)"` as well as functions in expressions `{{@.formatDate(...)}}"`. It's concise, explicit and flexible.

...as for the resolution mechanism you talked about, I must confess I wasn't quite able to follow.

PR #2711, opened 2016-10-23
Add support for spread arrays and objects and merge with call spreads - RFC - 0.9

evs-chris:
In the process of working on the tutorial site, I had the desire to have an options object for a decorator that included the keys from another object. Of course, you can already do that with `Object.assign` where available, but the es201x syntax is so much better: `{ foo: 'bar', baz: 'bat', ...someObj, ...otherObj }` === `Object.assign( {}, { foo: 'bar', baz: 'bat' }, someObj, otherObj )`. (_Note_: the empty first object is only there for equivalence to transpiled code - it's not strictly necessary) We already support spread args for calls, but not lists and objects. I also needed a break from docs.

This adds support for spread arrays `[ foo, bar, ...baz, bat, bip, ...bop ]`, spread objects from above, and simplifies the spread arguments code a bit by using the spread array helpers. All of these are nestable, and spreads can now apply to expressions here as well as the existing support for references, so you can `[ foo, ...bar.baz( why, ...not ) ]`. Array spreads are converted to a concat on an empty array `[].concat([foo, bar], baz, [bat, bip], bop)`. Argument spreads are still handled with an IIFE so that the function can be called with the correct context.

This does add about 1k to the compiled codebase and about 350b to the minified code. I've been trying to slim down the code a bit, so I thought I'd ask for opinions on whether es201x features that make sense are worth the extra weight. What say you Ractive users who watch the issue tracker?

Not all of es201x that could be supported is supported. :smile:

## Is breaking:
Don't think so.

## Reviewers:
Yes please.

### DIFF
--- a/test/browser-tests/computations.js
+++ b/test/browser-tests/computations.js
@@ -1038,4 +1038,36 @@ export default function() {
 		expected = 'foogoof';
 		r.set( 'bar', 'goof' );
 	});
+
+	// phantom just doesn't execute this test... no error, just nothing
+	// even >>> log messages don't come out. passes chrome and ff, though
+	if ( !/phantom/i.test( navigator.userAgent ) ) {
+		test( `various spread expressions compute correctly`, t => {
+			const r = new Ractive({
+				el: fixture,
+				template: `{{ JSON.stringify([ foo, bar, ...baz, ...bat( { bip, bop: 42, ...whimmy.wham( ...[ zat, wozzle, ...qux, bar ], zip ), bif: 84 } ), bar, foo ]) }}`,
+				data: {
+					foo: 1,
+					bar: 2,
+					baz: [ 'a', 'b', 'c' ],
+					bat ( obj ) { return [ obj, 123 ]; },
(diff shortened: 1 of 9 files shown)


fskreuz:
Won't be long before we have a full-blown JS parser in here. We have a full HTML parser in here already. 😄

test/browser-tests/computations.js
```diff
@@ -1038,4 +1038,36 @@ export default function() {
...
+					zat: true,
+					wozzle: false,
+					qux: [ 'z', 26, 'y', 25 ],
+					zip: 'zip'
+				}
+			});
+
+			t.htmlEqual( fixture.innerHTML, `[1,2,"a","b","c",{"bip":99,"bop":42,"0a":"z","1a":26,"2a":"y","3a":25,"4a":2,"5a":"zip","bif":84},123,2,1]` );
+		});
+	}
```
fskreuz: I'm currently looking into simplifying the test setup. Nice to know stuff like this can happen. 😄

PR #2720, opened 2016-10-27
Reference refactor with @ short form and context popping - RFC

evs-chris:

This is a refactor of the reference resolution code that pulls all things resolve into one place. It also adds two features:
### @ short form

Starting with 0.8, plain, non-prefixed method events were deprecated because they have caused a bit of confusion before, and more so because parsing them was a bit challenging making what you can do quite limited. They also have a completely different context than the rest of the references in templates that look the same.

One unfortunate side effect is that calling a method from an event directive is considerably more noisy. This introduces the short form of `@this.whatever` as ~~`@whatever`~~ `@.whatever`. This form was also ~~proposed~~ mostly proposed in #2345, but this PR punts on the additional root reference and state mapping for components.

This also adds support for the short form for API methods, which is nice because some of the public constructs on Ractive instances are moving toward observable management. Notably, `@this.children.byName`, which is a maintained object with lists of anchored children. I think `@this.event` will follow shortly as well.

**Question:** There are a handful of reserved names, like `@index` and `@keypath`, that are not available for short form. That bothered me a bit at first, but the more I look at it, the less it bothers me - to the point that I'm almost completely ok with it :smile:. My original thought on how to avoid that was to require a `.` after the `@` e.g. `@.index`. ~~I now think that looks pretty awful, but~~ I thought I'd ask for opinions, as I often lack perspective on usability. Developer experience and whatnot :wink: I've flip-flopped on this again. I'm also thinking of running for some sort of political office...

~~This also makes `@node` a reserved special reference for now, because I've been looking at some event handling, and sometimes it would be really nice to be able to use things like `@node.value` rather than `event.original.target.value` when handling events.~~ No need for this now.

### Context popping

Ractive already supports keypath popping using the familiar path syntax `../../`, but there's no way to have an explicit reference to something in a parent context. This adds a new prefix `^^/` that is used in the same way as `../`, but refers to context instead of keypaths. So, for `{{#some.path}}...{{#~/some.root.path}}{{^^/foo}}{{/}}...{{/}}`, the `^^/foo` will skip the rooted context `some.root.path` and get `some.path.foo`. `^^/` is also stackable with more `^^/`s and `../`s, so you can pop up the context hierarchy to some point and then pop up the keypath hierarchy e.g. `{{^^/^^/../../WHY}}`.

This isn't something that comes up _too_ often, but when you get in a nested context like the root reference context above, it can be really usefuly to get out of it without using a slower path. When trying to keep references unambiguous, explicit is the fastest way, and both aliases and non-scoped paths take the slow route.
### Bonus

This also adds a fast-track for un-scoped references that are properties of the current context e.g.

``` js
context = { foo: { bar: { baz: 'yep' } } }:
tpl = `{{#with foo.bar}}{{baz}} and {{.baz}}{{/with}}`;
```

both `baz` references will resolve fairly quickly.

Merging `resolveAmbiguous` reference into `resolveReference` also allows some somewhat expensive operations to be done only once per resolution (findContext and splitKeypath).

Mostly usability. A little bit performance.

## Is breaking:

Nope.


paulocoghi:
If the creation of `@` as a short form o `@this` brings more problems than solutions, it's perfectly fine to stay only with `@this`.

paulocoghi:
> When trying to keep references unambiguous, explicit is the fastest way

Totally agree. For this reason, `^^/` is very welcome!

fskreuz:
> `@this.whatever` as `@whatever`

Doesn't this potentially collide with future reserved words? What if someone had a property called `something`. This would equate to `@this.something`, whose short form is `@something`. But then when we need to use `something` as a legitimate reserved word, this would unexpectedly break that someone's app who expected `@something` as `@this.something`. And we'll never know until that someone speaks up here in Github.

Am I missing something? 😄

paulocoghi:
@fskreuz @evs-chris From the perspective of collisions with reserved words, maybe it's better to not implement the `@` short form.

paulocoghi:
> both `baz` references will resolve fairly quickly.  :tada: :tada:

So from now on, it will not be necessary to use `.baz` to prevent loss of performance?
_as you previously informed https://github.com/ractivejs/ractive/issues/2366#issuecomment-177652382_

evs-chris:
Well, there is the possibility of overlap with a future special ref, but at this point, there's really not much more that could be usefully provided in a template context. The only other thing I can think of they _may_ be useful would be access to a `getNodeInfo` object, maybe as `@info`, but that's pretty questionable, as it's mostly a way to get to template context from the js side. I think there's actually more danger of Ractive getting a new property or method on instances that collides with something user-defined.

I originally had `@event` reserved too, but then I realized it would be a pointer for the current event, which is stored as the `event` property on an instance - `@this.event` or `@event` :smile:

Can anyone else think of something that may be useful beyond what's already available - context/data as normal refs, the current index or key of an iteration, current keypath in various forms, current ractive instance, and global model. The current DOMNode as `@node` was the only thing I could think of.

> So from now on, it will not be necessary to use .baz to prevent loss of performance?

`.baz` will be a _microscopic_ bit faster, since scoped references are handled first in the resolution process. `baz` will resolve immediately after a check for special refs rather than having to walk up the template checking for aliases and other refs first.

dagnelies:
First off, thanks Chris for all your efforts. I can clearly see you are pushing things forwards.

I think the `@` shortcut is welcome from a user perspective because `@this` is verbose  ...however, great care should be taken not cause trouble by misinterpreting expressions containing `@index, @key, @keypath, ...` (I don't even know everything that is available). And like @fskreuz  said, it strongly blocks the possibility to add "@ keywords" in the future. 

Because of these two reasons, I'd like to suggest a little tweak, that is hopefully OK with everyone.
In order to avoid trouble, I'd vote for `@.methodName(...)`, with a dot in between in order to "isolate" the `@`. That way ambiguities are resolved regarding existing and future @ keywords.

Btw, I know it's a discussion in the past, but I'm still not fan of `@this` because it's similarity to the `this` keyword. But I'd be happy with a "`@.`"

evs-chris:
@dagnelies would like to get your take on adding the current instance to the end of the resolution process: https://github.com/ractivejs/ractive/pull/2705#issuecomment-256833909

That is one of the suggested alternatives to the `@` part of this PR, and part of me quite likes it.

sarendipitee:
> That is one of the suggested alternatives to the @ part of this PR, and part of me quite likes it.

I'd still prefer that and I think that is the least confusing to new users. Even though not the most ideal, I am open to `@` or `@.` if that's what @evs-chris says is easiest or best.

evs-chris:
I've made a couple of changes to this:
1. The shorthand now requires a `.` after the `@` e.g. `@.foo === @this.foo`
2. In addressing recent issues and looking at closing out some backlog, it seems there would be a use for shared global state for ractive instances that isn't susceptible to stomping from outside of ractive. So I introduced `@shared`, which is simply the `@global` special, but based on a ractive-private store rather than `window`. As with `@global` and `@this`, it is addressable both from the template and the API. To be clear, this means you can have multiple instances referencing `@shared.someProp`, update it from one, and all of the other refs will be notified too. It's kinda like adding to `Ractive.data`, but notifications are automatically handled rather than requiring `update` calls on all instances when it changes.
3. `@this.parent` and `@this.root` have been available since 0.8, but they will create parallel models that don't get notified of changes. This adds special handling in the RactiveModel that automatically creates a link for these props, so `parent.set('@this.someProp`, 1)`will notify`@this.parent.someProp`references in the child and`child.set(`@this.parent.someProp`, 2)`will notify`@this.someProp`references in the parent.`attachChild`has been updated so that`@this.parent` changes notify automatically.
4. Since 0.8, `@this.parent.get('someData')` is available to access parent data manually from children. This adds special handling for viewmodel access to RactiveModel via the `data` property, so `@.parent.data.someData` is equivalent to `@.parent.get('someData')`, except it's not a computation, so it acts like a normal model ref e.g. is bindable and mutable.

I'm thinking that adding the instance to the end of the resolution path as an option that defaults to on will be a good thing for usability, so I'm going to go ahead and add that. Explicit will still be better, as is always the case. And it will be able to be turned off.

While I'm mucking about in reference resolution, I think I'll also add a `warnAboutAmbiguity` flag for when references don't resolve.

evs-chris:
This now includes defaulted instance member resolution, so you can, once again, do:

``` html
<button on-click="set('foo', 42)">click me</button>
```

assuming that there is no `set` property in the current resolution path and that the instance wasn't created with `resolveInstanceMembers: false`.

There's also a `warnAboutAmbiguity` flag that, when set to true, will issue a console warning in debug mode if a reference doesn't resolve to immediate context, creates an implicit mapping, or just plain doesn't resolve.

evs-chris:
Here's the final list of changes, which I shall go ahead and merge to make feedback easier on the way to 0.9 releasing:
1. `@.foo` is shorthand for `@this.foo`. This extends to `ractive.set('@.foo', 1)`, so that you can keep refs in sync easily.
2. `^^/` is a new prefix that makes it possible to pop contexts in the same way that `../` pops keypaths. Another case of explicit is always better, and it makes some scenarios involving nested `~/whatever` contexts much easier to navigate.
3. Special refs that resolve to a RactiveModel (`@this.root`, `@this.parent.parent`, etc) or the special `@this.data` (or `@this.root.data`, `@.parent.data`, etc) now resolve using the actual models rather than a computation. This lets them stay in sync. Notably, this is the workaround for the (not recommended, but sometimes necessary) access of root data from a component: `@.root.data.whatever` as opposed to `~/whatever`, which stays strictly in component scope. I was going to add `~~/`, but that seems a step to far for now. We'll try the alternative long form and see what shakes out. This is handled by creating an implicit link to the appropriate RactiveModel or RootModel during reference resolution the first time through, so these are lazy inited. The attachChild API also keeps `@this.parent` and `@this.root` up to date as children are attached and detached.
4. There's a new `@shared` special, which is basically a Ractive-private `@global`, meaning that you can share data across _all_ Ractive instances on a page using `@shared.whatever` and everything will stay in sync as you would expect. It's basically `@global` that can't get stomped on by external code. It is also available to API methods e.g. `ractive.set('@shared.whatever', 'yep')`.
5. There's a new option (defaulted to on) `resolveInstanceMembers` that puts `@this` on the end of the resolution path, so you can do `<button on-click="set('foo', 'bar')">yep</button>` again if there is no `set` in the data context. This is the main one I want to get feedback on in the wild. Again explicit is always better as far as Ractive-the-codebase is concerned, but in most cases, this feels like an improvement in Developer Experience™ and has very little overhead where it would likely be used the most (events).
6. There's a new flag (defaulted to off) `warnAboutAmbiguity` that will issue warnings from the resolution process when a reference:
   1. resolves, but not in the immediate context e.g. `{{#with foo.bar}}{{#with .baz}}{{bat}}{{/with}}{{/with}}` where `bat` is a property on `foo.bar`.
   2. resolves, but also creates an implicit mapping from component to parent
   3. does not resolve
7. Calling `ractive.set('.foo', 'bar')` will now issue a warning (see the `.`?) with some pointers for what you may have been looking for and not create a `data[''].foo` path. So it's a noop with a (hopefully) helpful warning.
8. Calling `ractive.set('~/foo', 'bar')` is now equivalent to `ractive.set('foo', 'bar')` because the context version of `set` allows it and it's easy to address. For reference `ractive.getNodeInfo('#my-selector').set('~/foo')` is one of the context versions, the other being the one on events.

PR #2722, opened 2016-10-27
Binding and/or mutation of children of computations - take two - RFC

evs-chris:

This is a port of #2331 to edge. It actually takes less code now :smile:

**Definition**: For this PR, "computed" means a value that isn't directly in data, so any `computed` entries from instance init and any expression in the template.

Starting with 0.8, binding/setting in children of computed values e.g. `<input value="{{computed.some.child}}" />` was disallowed because it can cause references outside of the computation to get out of sync with their model. See #2123 for more details.

This introduces a new flag `syncComputedChildren`, which defaults to false, that sets up children of computed paths to be mutable. When a child mutates, it causes an updated on all of the computation's dependencies, which makes sure any other references outside of the computation stay in sync. That should cover the most common use-cases for child mutation, like binding fields on objects in a filtered list to inputs and having the change propagate anywhere else the objects are bound.

I changed the name from `derivedBindings` to `syncComputedChildren` because, to me, it is a bit easier to suss out what the flag means/will do.

The reason the flag is defaulted to false is if you're not careful, you can murder the performance of your app by causing a ton of unnecessary cascade updates.
### Bonus

I also DRYed up some of the computation code. ExpressionProxy, Computation, and ComputationChild now use the same function from the appropriate prototype instead of having the same code duplicated.

I also added value caching to ComputationChild, so that they don't _always_ ask for values all the way up to the computation for _every_ access. This should make complex computed structures a bit faster.

#2123

## Is breaking:

Nope

### DIFF
--- a/src/Ractive/config/defaults.js
+++ b/src/Ractive/config/defaults.js
@@ -21,6 +21,7 @@ export default {
 	// data & binding:
 	data:                   {},
 	computed:               {},
+	syncComputedChildren:   false,
 	magic:                  false,
 	modifyArrays:           false,
 	adapt:                  [],
--- a/src/model/Computation.js
+++ b/src/model/Computation.js
@@ -118,27 +118,6 @@ export default class Computation extends Model {
 		return result;
 	}
 
-	handleChange () {
-		this.dirty = true;
-
-		this.links.forEach( marked );
-		this.deps.forEach( handleChange );
-		this.children.forEach( handleChange );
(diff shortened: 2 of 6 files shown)


JonDum:
Heck yea! :)

I'll try this with some of the fiddles from #2331 and see how it fares.

fskreuz:
> I changed the name from derivedBindings to syncComputedChildren because, to me, it is a bit easier to suss out what the flag means/will do.

👍

Issue #2723, opened 2016-10-28
Components no longer inherit parent's adaptors


ADantes:
## Description:
In previous versions of Ractive, which adaptors to use would cascade down the component chain.  This is no longer the case starting with 0.8.x.  I honestly think how it currently works on the 0.8.x branch is the better case since it results in more isolated behaviour.  However, this needs to either be added to the list of breaking changes or fixed.

## Versions affected:
0.8.x

### HTML
    <script id="component" type="ractive/template">
      {{#each collection}}
        <p>{{this}}</p>
      {{else}}
        <p>No Items</p>
      {{/each}}
    </script>

    <script id="view" type="ractive/template">
      <h1>Test</h1>
      <Component />
    </script>

    <main></main>

### JS

    var Component = Ractive.extend({
      // 0.8.x requires this next line, 0.7.3 does not
      // adapt: ['Backbone'], 
      oninit: function() {
        this.set("collection", new Backbone.Collection);
      },
      template: "#component"
    });

    var View = Ractive.extend({
      adapt: ['Backbone'],
      components: {
        Component: Component
      },
      template: "#view"
    });

    new View({
      el: "main"
    });

### JSFiddle

0.7.3: https://jsfiddle.net/ysonpy1u/
0.8.2: https://jsfiddle.net/du1tpn0v/

Issue #2721, opened 2016-10-27
Make patchedArrayProto Configurable


paulie4:
## Description:
@Jeff17Robbins and I wrap the patched Array mutatorMethods to know when Ractive child elements are being created, but we needed to modify ractive.js to allow re-configuring the property.

## Versions affected:
all

## Platforms affected:
all

## Reproduction:
n/a

## **Code:**
From [patch.js](https://github.com/ractivejs/ractive/blob/dev/src/Ractive/static/adaptors/array/patch.js#L38):
```
	defineProperty( patchedArrayProto, methodName, {
		value: method
	});
```

Proposed change:
```
	defineProperty( patchedArrayProto, methodName, {
		value: method,
		configurable: true,
	});
```

PR #2724, opened 2016-10-28
Inherit parent adaptors when initializing a component - #2723

evs-chris:
This checks for a parent instance during config, and if there is one, copies its adaptors to the end of the instance adaptor list. Apparently this changed some time between 0.7 and 0.8. This is my first foray into the construct code, so I'm not positive this is the best way to handle this.

There's also some question as to whether or not components should inherit their parent's adaptors. I think it makes sense for an app that's working with special data to have the component be able to work with it as well. Mapping aren't affected, since the mapping defers wrapping to the source, but any other setting of data is.

This also makes the prototype overrides on arrays from the builtin array adaptor configurable per #2721.

#2723 #2721

## Is breaking:
No?

## Reviewers:
Anyone familiar with adaptors.

### DIFF
--- a/src/Ractive/construct.js
+++ b/src/Ractive/construct.js
@@ -62,24 +62,30 @@ export default function construct ( ractive, options ) {
 	}
 }
 
-function combine ( a, b ) {
-	const c = a.slice();
-	let i = b.length;
+function combine ( arrays ) {
+	const res = [];
+	const args = res.concat.apply( res, arrays );
 
+	let i = args.length;
 	while ( i-- ) {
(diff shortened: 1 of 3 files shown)


skeptic35:
Just a question: Shouldn't that respect the ```isolated``` property? It didn't in 0.7.3, as you can see [here](https://jsfiddle.net/ysonpy1u/1/), so it would be a breaking change. One that might even affect my own code (should check that...) but still: To me it would appear logical to treat ```isolated``` components as isolated as possible.

evs-chris:
@skeptic35 that's a fair point. Updated!

PR #2728, opened 2016-10-30
Linting

ghost:

Good code starts with concise code. 😄 

At the moment, our only defense against potential bugs are tests. This PR lints everything according to the existing rules, clear out the existing flags and ultimately restart squeaky clean. That way, we can have ESLint spot potential mistakes early.

## Is breaking:

No.

## Reviewers:

Everyone.

---

Also, don't worry about conflicts with this PR, just go ahead and merge new stuff. I'll handle fixing this PR when it does conflict with latest developments.


evs-chris:
For the unused deps in the model modules, there's a circular dependency that I gave up on trying to sort out and just manually added methods to the prototype at the end of one of the modules. That should only be an issue with LinkModel, though.

Also big :+1: for the de-lint :smile:. I played with using the whitespace rules to enforce the style  guide, but they don't seem to support both
`foo( 'bar', { bar: 1 } )` and
`foo({ bat: 1 })` at the same time. Spacing out nested braces just doesn't look right to me, but having unbalanced spacing is also visually painful. I'm not sure if there's a way to resolve that that doesn't end in
`foo( 'bar', { bar: 1})` or
`foo( { bar: 1 } )`. Granted those examples aren't awful, but it gets much worse with more nesting.

ghost:
Yep. I tried that the spacing other day. It worked well with pre-parens and within parens. Good readability.

```
foo( bar );
var foo = { bar: 'baz' };
var foo = [ 1, 2, 3 ];
function foo ( bar ) { ... }
```

until you had to deal with stuff like:

```
// The space between parens and brace just looks akward
foo( {
  bar: 'baz'
} ).bind(this);
```

So I just left it out and declared it as a "recommended but use where it makes sense". I could just put it in eslint as a warning and not an error as well.

ghost:
Merging this in early. There's a second wave of linting (~300 warnings) and I don't want this one to grow huge.

Issue #2727, opened 2016-10-30
toHTML does not render class-* and style-*


shirotech:
## Description:

When rendering on server-side .toHtml does not compute the class-\* or style-\* attributes and leave them in their raw state. Client-side rendering is fine though.
## Versions affected:

0.8.2
## Platforms affected:

Maybe all
## Reproduction:

https://plnkr.co/edit/2wU2B1A98W6S6panm6iX?p=preview
## **Code:**

Expected

``` html
<div class="test" style="font-weight: bold;">Test</div>
```

Actual

``` html
<div class-test="true" style-font-weight="bold">Test</div>
```

Have to use {{#if}} and {{#unless}} wrapper as workaround for now.. not ideal as they are messy and inconvenient to use.

Hope this can be resolved in next release, thanks! :)

Edit: I tried with defining a valid "el", seems to work but what can "el" be from server as there is no document?

van-nguyen:
Have to use {{#if}} and {{#unless}} wrapper as workaround for now.. not ideal as they are messy and inconvenient to use.

PR #2729, opened 2016-10-30
Various toHTML-when-not-rendered-to-DOM fixes - #2727

evs-chris:

This modifies the render browser tests to run `toHTML` testing before rendering to the DOM to simulate what happens in node. The _main_ reason I did that was to track down the issue in #2727, but it also turned up a handful of other issues with code that expected a DOM environment where it shouldn't.

The issue #2727 is that the updateAttribute handler that is called on render sets up the special `styleName` and `inlineClass` properties on the `Attribute`. If it's never rendered, `Attribute.toString` won't realize that it's a special directive and render itself as a plain old string attribute. This PR adds a check for style- and class- directives and sets them up to be appropriately `toString`ed if they've not been rendered.

This also removes the test involving having an array as root data, as that has been deprecated for quite a while and was interfering with the way the test resets data now.
#2727
## Is breaking:

Nope.

### DIFF
--- a/src/view/items/Triple.js
+++ b/src/view/items/Triple.js
@@ -91,6 +91,9 @@ export default class Triple extends Mustache {
 			const anchor = this.parentFragment.findNextNode( this );
 
 			parentNode.insertBefore( docFrag, anchor );
+		} else {
+			// make sure to reset the dirty flag even if not rendered
+			this.dirty = false;
 		}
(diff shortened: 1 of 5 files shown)


shirotech:
Nice one @evs-chris thanks for getting this fixed so quick 👍

PR #2734, opened 2016-11-01
Warn when event cannot be cancelled due to missing event-object

### LINKED ISSUE #2731: "Cannot read property 'original' of undefined" in EventDirective.prototype.fire
## Description:
There's a component with a button which, when clicked, does some stuff in the component-context, then notifies the parent using ractve.fire.

If the handler in parent returns false and the compont does not pass the event-object in ```ractive.fire```, the error ```Uncaught TypeError: Cannot read property 'original' of undefined``` is thrown in EventDirective.prototype.fire.

Not sure if this is a bug or a nicer error message should be implemented. Anyway this is not urgent for me, since there's an obvious workaround.

## Reproduction:
https://jsfiddle.net/dpqexp4b/ - please compare CrashyComponent with GoodComponent

PS: I can't thank you enough for the latest releases and plans for the future. So I wont;) \o/

aphitiel:
97b0305 fixes #2731 by silently ignoring the users intent to cancel a component event. This PR just adds a little warning when the event cannot be cancelled due to a missing event-object.

I think otherwise people could get hurt, for example when they move functionality into a component, since
* plain ```<button on-click="@this.a()``` can be cancelled without passing the event
* docs say events can be cancelled by returning false.

#2731 

## Is breaking:
Nope

### DIFF
--- a/src/view/items/shared/EventDirective.js
+++ b/src/view/items/shared/EventDirective.js
@@ -121,10 +121,9 @@ export default class EventDirective {
 			const result = this.fn.apply( ractive, values ).pop();
 
 			// Auto prevent and stop if return is explicitly false
-			let original;
-
 			if ( result === false ) {
-				if ( event && ( original = event.original ) ) {
+				const original = event ? event.original : undefined;
+				if ( original ) {
 					original.preventDefault && original.preventDefault();
 					original.stopPropagation && original.stopPropagation();
 				} else {
--- a/src/view/items/shared/EventDirective.js
+++ b/src/view/items/shared/EventDirective.js
@@ -12,7 +12,7 @@ import RactiveEvent from '../component/RactiveEvent';
 import runloop from '../../../global/runloop';
 import { addHelpers } from '../../helpers/contextMethods';
 import { setupArgsFn, teardownArgsFn } from '../shared/directiveArgs';
-import { warnIfDebug } from '../../../utils/log';
+import { warnOnceIfDebug } from '../../../utils/log';
 
 const specialPattern = /^(event|arguments)(\..+)?$/;
 const dollarArgsPattern = /^\$(\d+)(\..+)?$/;
@@ -127,7 +127,7 @@ export default class EventDirective {
 					original.preventDefault && original.preventDefault();
 					original.stopPropagation && original.stopPropagation();
 				} else {
-					warnIfDebug( 'Handler returns false, but no event object is available' );
+					warnOnceIfDebug( `handler '${this.template.n}' returned false, but there is no event available to cancel` );
 				}
 			}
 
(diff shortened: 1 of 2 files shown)


src/view/items/shared/EventDirective.js
```diff
@@ -121,9 +122,14 @@ export default class EventDirective {
...
-				original.preventDefault && original.preventDefault();
-				original.stopPropagation && original.stopPropagation();
+
+			if ( result === false ) {
+				if ( event && ( original = event.original ) ) {
+					original.preventDefault && original.preventDefault();
+					original.stopPropagation && original.stopPropagation();
+				} else {
+					warnIfDebug( 'Handler returns false, but no event object is available' );
+				}
```
ghost, line 131: Consider putting `original` inside the block. It's only being used in there anyways. Also use `const` instead of `let` to indicate that it's never going to change.

I'm also not a fan of assignment inside conditions. It's hard to read and it's usually late to identify some value was retrieved then used.

``` js
if(result === false){
  // Could have been null, but a missing property is undefined. Consistency. ¯\_(ツ)_/¯
  const original = event ? event.original : undefined;

  if(original){
    original.preventDefault && original.preventDefault();
    original.stopPropagation && original.stopPropagation();
  } else {
    warnIfDebug( 'Handler returns false, but no event object is available' );
  }
}
```

alternatively... avoiding the additional indent.

``` js

const falseReturn = result === false;
const original = event && event.original;

if(falseReturn && original){
  original.preventDefault && original.preventDefault();
  original.stopPropagation && original.stopPropagation();
} else if (falseReturn && !original){
  warnIfDebug( 'Handler returns false, but no event object is available' );
}
```

_By the way, these are just suggestions. The code is totally fine._ 😉

aphitiel: Thanks for your feedback, which I totally agree with. I was trying to add the warning in a minimally invasive way for easy reviewability. Went with your straightforward first suggestion.

BTW: One might consider enabling in eslint: no-cond-assign (though it wouldn't have caught this one) and prefer-const;)

src/view/items/shared/EventDirective.js
```diff
@@ -120,10 +121,14 @@ export default class EventDirective {
...
-			if ( result === false && event && ( original = event.original ) ) {
-				original.preventDefault && original.preventDefault();
-				original.stopPropagation && original.stopPropagation();
+			if ( result === false ) {
+				const original = event ? event.original : undefined;
+				if ( original ) {
+					original.preventDefault && original.preventDefault();
+					original.stopPropagation && original.stopPropagation();
+				} else {
+					warnIfDebug( 'Handler returns false, but no event object is available' );
```
evs-chris: I'd suggest including the event directive name as well, perhaps as something like:

``` js
warnOnceIfDebug( `${this.template.n} handler returned false, but there is no event available to cancel` );
```

evs-chris:
:+1:

Issue #2736, opened 2016-11-02
0.9: Untested codepath with error (arg for event handler from previously uninitialised partial context) 


aphitiel:
## Description:
Partial puts a property of its context into an argument of an event handler. If this property is not initialised, we get ```removeFromArray is not defined``` in ```setupArgsFn``` in ```directiveArgs.js``` when ```set```ing it. Probably just a missing import, but I don't yet understand this stuff well enough. Anyway here's a small test. Anyone know how to use simulant in jsFiddle?

## Reproduction:
http://jsfiddle.net/5jvc5ekw/1/

evs-chris:
It is indeed a missing import from a refactor. I think getting the lint down to the point that the linter can fail the build will help considerably with this sort of issue. My editor environment still hasn't caught up to the switch from jshint :frowning:

I think to use simulant in a test fiddle, you would just need to include it e.g. `<script src="https://unpkg.com/simulant"></script>` in the html pane.

Since you found and diagnosed it, would you like to fix it?

PR #2738, opened 2016-11-02
Test and fix for #2736 (arg for event handler from previously uninitialised partial context)

aphitiel:
* Simple test so there is something along that code path
* Add missing import

#2736

## Is breaking:
No

## Reviewers:
@evs-chris


evs-chris:
:+1:

PR #2741, opened 2016-11-03
Update lint configs

ghost:

Update linting related stuff. Wanted to push this forward first before doing a huge clean-up operation over at the Linting PR.

- Fix the npm lint commands. Better named. Tests lint command now pointing to correct directories.

- Alphabetically sorting the rules so they're easily found.

- Adding `prefer-const` so ESLint can capture attempts to reassign when we shouldn't.
    - We have a lot of patterns and container objects/arrays defined using `var` and `let`.
    - We have a lot of temporary variables between blocks.

- Lenient single-quote policy. Can use backtick without warnings. Can use any quote to avoid escaping.
    - We use backticks on templates in tests.
    - We use double-quotes to use avoid escaping single-quotes (and vice versa) in tests.

## Is breaking:

No

## Reviewers:

Everyone.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -8,20 +8,21 @@
 		"sourceType": "module"
 	},
 	"rules": {
+		"arrow-spacing": "error",
 		"indent": [ "error", "tab", { "SwitchCase": 1 }],
-		"quotes": [ "error", "single" ],
 		"linebreak-style": [ "error", "unix" ],
-		"semi": [ "error", "always" ],
 		"no-mixed-spaces-and-tabs": [ "error", "smart-tabs" ],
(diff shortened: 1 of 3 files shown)


evs-chris:
Was doing some de-linting on some code I was messing with in another PR, and I'm thinking we should switch on-var to `{ "initialized": "never" }` instead of "never". I think `let foo, bar, baz;` is perfectly acceptable, whereas `let foo = true, bar = 'something', baz;` is the problematic form.

ghost:
I just went with the existing `one-var` but yep, I agree with `{ "initialized": "never" }` 😄

Issue #2270, opened 2015-11-10
Set global Ractive variable even when AMD is detected


Ghazgkull:
It would be very helpful if Ractive always set the global Ractive property (e.g. window.Ractive) when it's loaded. Currently, the global property isn't set if define.amd or module.exports is detected.

The current behavior works fine for monolithic environments, where authors can make their own decision to use a module loader like RequireJS or not. But the current behavior doesn't work in the messy reality where Ractive is pulled into a mix of other 3rd party scripts.

The problem case looks like this:
1. Write a plain Javascript library that doesn't use RequireJS.
2. Load Ractive from your Javascript library by adding the appropriate `<script>` tag to the document.
3. Access Ractive in your library via window.Ractive. All is well. You now have a library that you can publish for people to include in their own sites.
4. ...until you're loaded onto a page with some other script that happens to use RequireJS. At this point, a global define.amd will become available, Ractive will see it and stop setting the window.Ractive property. Even worse, Ractive is defined as an anonymous module in RequireJS so there's literally no way to access it if it was loaded via a `<script>` tag instead of through the RequireJS API.

The problem with RequireJS which prevents accessing Ractive if it's loaded via a `<script>` tag is described on the RequireJS site:

http://requirejs.org/docs/errors.html#mismatch

> Be sure to load all scripts that call define() via the RequireJS API. Do not manually code script tags in HTML to load scripts that have define() calls in them.

Basically, RequireJS adopts an "all or nothing" model where once they're loaded on a page, they assume that all JS on the page is using them. But libraries like Ractive can make sure consumers still have a choice by continuing to set a global property in addition to the module system integrations.

Ghazgkull:
A little extra context. [Here's a great article](http://ifandelse.com/its-not-hard-making-your-library-support-amd-and-commonjs/#amdcommonjsandplainbrowsersupport) talking about adding AMD/CommonJS support to your module, which includes a discussion at the end in the [Update section](http://ifandelse.com/its-not-hard-making-your-library-support-amd-and-commonjs/#update) about why they added back the setting of a global variable.

JonDum:
I thoroughly disagree with this. What if you're loading a self-contained Ractive app with a different version of Ractive then what's on the global? What if you need (for whatever crazy reason) to have different defaults, decorators, components on different Ractive instances? 

[...]

PR #2744, opened 2016-11-05
Add a way to force ractive to register as a global - #2270

evs-chris:
This adds a way to force Ractive to be set on `window` regardless of the method used to load it (AMD, require-ish things). Whatever script contains Ractive should be linked with `RactiveForceGlobal` as a query param on the src e.g. `<script src="./ractive.min.js?RactiveForceGlobal"></script>` or `<script src="./mySpecialBundle.js?RactiveForceGlobal"></script>`. As long as the currently loading script tag has `RactiveForceGlobal` somewhere in the src, there's a window available, and it doesn't already have a `Ractive` member, this will set `window.Ractive` to the loading version of Ractive.

@Ghazgkull I tested this with a fake AMD environment, and it seemed to work fine. I _think_ browser support goes all the way back to IE9. Can you see if this will work for you? If it does, I think it should suffice for anyone else with a conflicted environment :smile:

#2270

## Is breaking:
Only if you use a module loader, named your copy of Ractive `RactiveForceGlobal.js`, and have code that can't tolerate a global Ractive being set :smile:

## Reviewers:
@Ghazgkull @fskreuz

### DIFF
--- a/src/Ractive.js
+++ b/src/Ractive.js
@@ -40,6 +40,20 @@ export default function Ractive ( options ) {
 	initialise( this, options || {}, {} );
 }
 
+// check to see if we're being asked to force Ractive as a global for some weird environments
+if ( win && !win.Ractive ) {
+	let opts = '';
+	const script = document.currentScript || document.querySelector( 'script[data-ractive-options]' );
+
+	if ( script ) {
+		opts = script.getAttribute( 'data-ractive-options' ) || '';
+	}
+
+	if ( ~opts.indexOf( 'ForceGlobal' ) ) {
+		win.Ractive = Ractive;
+	}
+}
+
 extendObj( Ractive.prototype, proto, defaults );
 Ractive.prototype.constructor = Ractive;
 


ghost:
Considerations on having two versions of Ractive running on the same page? I think this was a concern on an earlier discussion. Two or more teams, potentially using different versions of Ractive, all pop in their creations on one page.

Brain dump:
- Was thinking of exposing the first Ractive that loaded, but that would be unpredictable. 
- Leave it to developer communication, but fast-moving teams, tight deadlines can undermine this.
- This assumes script-loaded Ractive. But in cases where you use a module loader/bundler?

What prevents one from manually exposing a global? Can we introduce a function that a user can call to expose the global from code in modules? Like

```
Ractive.makeGlobal();
```

Ghazgkull:
This looks great and I would absolutely use it. Thanks!

@fskreuz As far as considerations for running two versions of Ractive on the same page, I think the only additional thing that we'd want is to restore Ractive.noConflict(). As a client, I do my best to implement this myself, but it's not as reliable as if the library did it internally.

There are situations (the one that prompted me to ask for this in the first place) where you _can't_ access Ractive unless it's defined as a global. So it's not possible to make a call like `Ractive.makeGlobal()` unless Ractive is already available as a global. :)

Ghazgkull:
Links for the previous comment:
- I did my best to explain the problem case in #2270 
- The Ractive.noConflict() request is #2066

evs-chris:
`noConflict` should be back as of 0.8.

As far as trying not to mangle multiple versions on the same page, I don't think there's really much we can do from the library side. With this and `noConflict`, any disparate teams should have what they need to not step on each others toes.

evs-chris:
Oh, and as long ractive is in _some_ script tag with a src attribute, this will kick in. It can be bundled, minified, and abused in whatever other ways that don't affect behavior of the abused code (no overzealous dead code removal). From what I can tell, the only place this absolutely won't work in a browser is directly included as the content of a script tag.

Ghazgkull (changes requested):
The fallback `else` case looks like it's just checking the first script tag on the page, which probably has a low chance of getting a hit. Instead, adding a wildcard attribute selector should do the trick to really find any tags with the flag: `document.querySelector(script[src*="RactiveForceGlobal"])`

evs-chris:
Whoops, that's supposed to be pulling the last script in the document, not the first. It's fixed now. Nice catch :smile:

Ghazgkull:
@evs-chris That looks better. But it still seems like it could fail to work for IE11 (ugh, I know) pretty often. If you load Ractive by appending it to `document.head` but there are any script tags in `document.body`, for example, whatever random script happens to be lower in the document will be examined by that check.

Is there any reason not to use the wildcard attribute selector? It seems like it's pretty foolproof as a backup and there's wide browser support for it: http://caniuse.com/#feat=css-sel3

Just asking because I'm pretty sure that this change would force me to drop IE11 support. Not necessarily the end of the world, but...

evs-chris:
The only reason not to use the flag in the selector would be to allow for other such flags in the future. I can adjust it now and it _might_ get changed at some point later on. Is there any reason not to append to body when doing manual async loading like that?

I suppose we could pick a ractive option prefix and look for _that_. Then each flag could run off of the selected src.

Ghazgkull:
I wouldn't mind moving the script tag to the body. But there are so many script tags constantly being added to the DOM that even if Ractive is added to the end of the body there's a good chance some other script will have been appended after it by the time it loads. (It would even mean that I as a client would have to make sure that I always load Ractive last or my own other dependencies could basically prevent it from setting the global that I need.)

evs-chris:
Hmmm... perhaps we should take a different tack. How about an attribute on the script element other than source, like maybe `<script src="/path/to/my-ractive-bundle.js" ractive-options="forceGlobal"></script>`? Then the check to force would look at the last of `document.querySelectorAll('script[ractive-options]')`.

Ghazgkull:
Yeah, I like the direction of an attribute better than a query param. Since it lets the element selector be exact (not just wildcard looking for a string somewhere in the src). Should it be one "options" attribute or should the attribute be specific to the option? `<script src="..." ractive-global="true">`

src/Ractive.js
```diff
@@ -40,6 +40,25 @@ export default function Ractive ( options ) {
...
+		const scripts = document.querySelectorAll( 'script[src]' );
+		if ( scripts.length > 0 ) src = scripts[ scripts.length - 1 ].src;
+	}
+
+	if ( src && ~src.indexOf( 'RactiveForceGlobal' ) ) {
+		if ( !win.Ractive ) {
+			win.Ractive = Ractive;
+		}
+	}
+}
```
ghost, line 55: Hmm. This might be... better.

``` js
// querySelector just gets the first marked script. Less messing around with arrays.
// data-* because valid HTML5.
const forceGlobal = document.currentScript || document.querySelector('script[data-ractive-options="force-global"]');

// Overwrite or use existing.
win.Ractive = forceGlobal ? Ractive || win.Ractive; 
```

ghost:
The only edge case I can think with this "search for marked `<script>`" approach is when a marked script loads Ractive, then a compiled script with a different Ractive loads after. It sees the marked script and overwrites the existing global Ractive. This shouldn't be a bother with code that does use `document.currentScript` tho.

Very edge-case, but something to consider.

PR #2745, opened 2016-11-05
Clean out the remaining src lint and fail the build if lint doesn't pass

evs-chris:
This resolves the remaining linter complaints about the src directory. It even found some code that was no longer in use but included in the build anyways :smile:

It also set the build script to fail if linting doesn't pass, so the code should stay lint-free going forward.

We like cleaner code?

## Is breaking:
Nope.

## Reviewers:
@fskreuz


evs-chris:
I'm gonna go ahead an merge this, cause I'm trying to clean up a few issues and it's starting to overlap. @fskreuze I hope this doesn't step on any of your changes.

ghost:
Totally fine. 👍 You write code fast. 😄

Issue #2339, opened 2016-01-12
Teardown invoked before transition finishes


pqvst:
I recently started using transitions. I have an element which contains a component. However, I noticed that teardown on the component occurs before the transition finishes (which doesn't look very good).

```
<div intro-outro="slide">
    <Component />
</div>
```

https://twitter.com/pqvst/status/686502421032472576

fskreuz:
You might want to post a reproducible demo so everyone can inspect what's happening.

pqvst:
@fskreuz Sure thing.

Github repo:
https://github.com/pqvst/ractive-transition-teardown-bug

Live demo:
https://rawgit.com/pqvst/ractive-transition-teardown-bug/master/demo.html

martypdx:
Confirmed in both `0.7.3` and `edge`.

Seems like a bug, any disagreement?

evs-chris:
Ehhhh. I don't think it's a bug so much as maybe a missing lifecycle event. Perhaps there should be a `torndown`?

The most convenient way I see to resolve this situation is to expose the current batch promise on runloop (`promise() { return (batch || {}).promise || Promise.resolve(true); }`) and then fire the hook off of `runloop.promise().then(() => youGetTheIdea())`. Does that sound reasonable?

pqvst:
What about adding an event called ``destroy`` rather than ``torndown``, so that ``teardown`` is invoked as the teardown process begins (i.e. before any transitions complete), and ``destroy`` is invoked just before the component is actually removed from the DOM? I feel that such an event is missing in any case.

    teardown -> transition begins -> transition ends -> destroy -> removed from DOM

Issue #2034, opened 2015-07-12
Lifecycle hook events should contain info about current instance


Rich-Harris:
My current problem: maintaining a lookup table of Ractive instances corresponding to nodes in a tree of arbitrary depth (an AST). The easiest way would be for the instance that contains the tree to listen for `Node.render` events bubbling up...

``` js
let lookup = {};

this.on({
  'Node.render', event => {
    const { instance } = event;
    lookup[ instance.uid ] = instance;
  },
  'Node.unrender', event => {
    const { instance } = event;
    delete lookup[ instance.uid ];
  }
});
```

...but that doesn't work because the `render` hook doesn't generate an `event` object. (Demo: http://jsfiddle.net/z1ffp1ws/.) Easy enough to work around for the time being, but this'd be nice to add at some point in the future.

PR #2746, opened 2016-11-05
add destruct hook, clean up a bit of cruft, and send the source insta…

evs-chris:
This does a bit of cleanup around hooks, including removing the deprecation warnings and support. It also adds a `destruct` hook that is the counterpart to `complete`. It fires after the transitions have finished on a component/instance teardown.

This also modifies hook events to pass along the ractive instance that originated the event as their last argument. `change` is the odd man out here, so it's `ractive.on('change', (changes, source) => ...)` whereas everything else is `ractive.on('change', source => ...)`.

#2339 #2034

## Is breaking:
Well kinda, if you're using proxy events named `complete`, `destruct`, `attachchild`, or `detachchild`, because those are the additional events that are now checked during parse. So it's not possible to `<button on-click="complete">nope</button>`, but I'd think that would be pretty uncommon.

### DIFF
--- a/src/Ractive/prototype/teardown.js
+++ b/src/Ractive/prototype/teardown.js
@@ -5,6 +5,7 @@ import { cancel } from '../../shared/methodCallers';
 import { warnIfDebug } from '../../utils/log';
 
 const teardownHook = new Hook( 'teardown' );
+const destructHook = new Hook( 'destruct' );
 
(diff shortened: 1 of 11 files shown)


ghost:
Would have gone with `destroy` but remembered we have `construct`. `destruct` is a good name. 👍

evs-chris:
I don't know exactly why, but destroy just didn't feel right.

PR #2758, opened 2016-11-10
A bit of build tweaking - updated devdeps and added es modules

evs-chris:
1. This bumps the versions of rollup and buble to latest. The latest gobble looks to have an issue with file watching, understandably because cross platform recursive file watching with node is a terrible mess, so it stays a version behind.
I had to patch around a bug in either buble or rollup wherein `import { test } from 'qunit';` + the object shorthand `{ test() { ... } }` results in `{ test: function qunit.test() { ... } }`. It only affected a few test though.
1. This adds ES module builds to the npm package and build branch that can be used for smarter bundling without requiring the ES module -> UMD/CommonJS -> ES module circus. I added both "jsnext:main" and "module" entries to the package, and used ".mjs" as the extension, which I _think_ covers all of the whacky ways to tell things we have an ES module in the package.
1. Switches to using a shallow clone when doing a release to the build branch, because it takes about 10 minutes to clone the whole thing from my crappy rural cable, and there's really no need to  pull all 100+ MB of history, especially when adding the ES modules bumps the release size up to 7MB. I'll be cherry-picking the release script change to 0.8.

## Is breaking:
Hope not :smile: - all of the tests seem to pass, though the minified file grew 200B.

## Reviewers:
Sure. I'm only going to subject edge to this, since I don't want the possibility of breaking a stable release in subtle ways if the rollup and buble updates surface something weird.


ghost:
Looks good to me! 😄

Issue #2730, opened 2016-10-31
Event names no longer support dashes


WaltzingPenguin:
## Description:
Inside templates, events handlers that are bound to an event name with dashes no longer trigger.

## Versions affected:
0.8.x

## Reproduction:

I expect to see both "COMPONENT" and "VIEW" logged to console.  This only happens for 0.7.3.  For 0.8.2, it only logs "COMPONENT".

0.7.3: https://jsfiddle.net/6zx12ufu/
0.8.2: https://jsfiddle.net/Le3fhgo8/1/

### HTML

    <script id="component" type="ractive/template">
      <button on-click="event-with-dash">Click Me!</button>
    </script>

    <script id="template" type="ractive/template">
      <p>
        <Component on-event-with-dash="event-with-dash" />
      </p>
    </script>

    <main></main>

### JS

    var Component = Ractive.extend({
      oninit: function() {
        this.on("event-with-dash", function() {
          console.log("COMPONENT");
        })
      },
      template: "#component"
    });

    new Ractive({
      components: {
        Component: Component
      },
      el: "main",
      oninit: function() {
        this.on("event-with-dash", function() {
          console.log("VIEW");
        });
      },
      template: "#template"
    });

evs-chris:
This is odd. According to the docs, `on-event-with-dash` should subscribe to three events: `event`, `with`, and `dash` - see http://docs.ractivejs.org/0.6/event-directives. I went back to the 0.6 docs just to make sure it wasn't a documentation screw up.

Is the issue that only _DOM_ events should allow `-` separators for multiple events? So component events should _not_ split on `-` when listening to events? If that's the case, it's an easy fix.

WaltzingPenguin:
Having different rules for DOM events and component events would be confusing.  This does look like it goes in the category of "things that worked but really shouldn't have".

Can we have better error messaging around this?  If event names shouldn't contain dashes, it would be nice to get an error when attempting to bind to an event like that.

evs-chris:
I agree that having different meanings in different contexts that look the same is not a good idea. That's probably why I assumed that component events should behave like DOM events and accidentally introduced the breakage in the first place :disappointed:

I'm not sure that there's a good way to warn in this particular scenario, unless we warn when you `ractive.fire('some-dashy-event')`. That seems like a legitimate use of the API, though.

ghost:
[...]

PR #2760, opened 2016-11-11
BREAKING: Allow escaping hyphens in on-event-name - #2730

evs-chris:
As of 0.8, component events are treated the same way as element events, so `<cmp on-foo-bar="...">` subscribes to both `foo` and `bar` events like `<a on-click-tap="...">` subscribes to both `click` and `tap` events.

This adds support for escaping `-`s in event directives, so that hypenated event names can be subscribed in templates: `<cmp on-foo\-bar="...">` will subscribe to one event `foo-bar`.

I would have gone with just warning when you fire a hyphenated event, but this seems like it might be a better option.

Opinions?

#2730

## Is breaking
:Yes, because event name splitting is moved to parse time, and the event VDOM node now has an array for `n` rather than a string. Templates are already broken between 0.8 and 0.9.

## Reviewers:
@MartinKolarik, Grand Wizard of the Order of Regex
Without negative lookbehind in js regexes, I'm at a bit of a disadvantage. I copied your keypath splitting, since it accomplishes roughly the same thing. There may be a better way, though. :smile:

### DIFF
--- a/src/parse/converters/element/readAttribute.js
+++ b/src/parse/converters/element/readAttribute.js
@@ -7,7 +7,7 @@ import flattenExpression from '../../utils/flattenExpression';
 
 const attributeNamePattern = /^[^\s"'>\/=]+/;
 const onPattern = /^on/;
-const eventPattern = /^on-([a-zA-Z\\*\\.$_][a-zA-Z\\*\\.$_0-9\-]+)$/;
+const eventPattern = /^on-([a-zA-Z\*\.$_]((?:[a-zA-Z\*\.$_0-9\-]|\\-)+))$/;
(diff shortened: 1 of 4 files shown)


MartinKolarik:
> Without negative lookbehind in js regexes, I'm at a bit of a disadvantage. I copied your keypath splitting, since it accomplishes roughly the same thing. There may be a better way, though. 😄

It's ok, although a more simple approach would also work in this case. Keypath splitting is a lot more complicated because the backslashes in keypaths can also be escaped.

``` js
function splitEvent ( str ) {
    let result = [], s = 0;

    for ( let i = 0; i < str.length; i++ ) {
        if ( str[i] === '-' && str[ i - 1 ] !== '\\' ) {
            result.push( str.substring( s, i ).replace( slashes, '' ) );
            s = i + 1;
        }
    }

    result.push( str.substr( s ).replace( slashes, '' ) );
    return result;
}
```

Issue #2217, opened 2015-10-13
Decorators raise element events? (post 0.8)


martypdx:
Something that always seemed dirty to me in decorators was mucking about with the ractive instance directly from within the decorator. 

You end up doing things like passing keypaths into the decorator or creating coupling by hardcoding them or names to methods or events to be handled on the ractive instance. (see [this group question](https://groups.google.com/forum/#!topic/ractive-js/yKi5YXKDubo), or [this SO](http://stackoverflow.com/questions/33041695/create-jquery-dialog-ractivejs-decorator) for recent examples).

It occurred to me what would make more sense is to allow the decorator to fire element events that could then be mapped back to the ractive instance in the template:

``` html
<div as-dialog='{ show: showDialog }' on-close='set( "showDialog", false )'>...</div>

<div as-someDecorator on-update='doSometihng( event.value )'>...</div>

<div as-sortable='{ allowDrop: true }' 
     on-drag='set( "position", { x: event.x, y: event.y } )' 
     on-drop='move( ...arguments )'>...</div>
```

(Quick note: I'm using the proposed new decorator syntax in these examples that moves the decorator name out into the attribute name with the prefix `as-` and uses implied bindings in the attribute value, so `decorator='tooltip:{{content}}'` becomes `as-tooltip='content'` or more property `as-tooltip='{ content }'` as would be the "recommended", but not required, object literal options syntax. Parameterless decorators can use the value-less attribute syntax like `as-sortable`).

It would require a change to the decorator interface in order to have access to a `fire` method, or perhaps it could be a ractive method accessible on the existing `node` parameter. And I think access to `ractive` would still be supported for those that prefer direct communication between the decorator and the ractive instance.

Details aside, I just wanted to float this out there to see if others were looking for a better bridge between decorators and the template.

JonDum:
If I'm not mistaken, aren't you already passed the instance in the decorator constructor, so you can call `instance.fire()`?

Secondly, I also like the attribute name for decorators. How would it work with multiple decorators? The order they appear on the tag is the order they're called? Also, what's the `as` prefix in `as-someDecorator`. Is that a personal namespace you use, or a suggestion to have all decorators start with that? 

Personally, I would go with no prefix at all:

`<div tooltip='{{someTooltip}}'></div>`

There's still have to be an entry in the `decorators` object that matches the attribute name for the decorator to be added.

martypdx:
[...]

PR #2764, opened 2016-11-14
Context helper expansion including observe and events - 0.9

evs-chris:

This does a few different things with context objects (event and `getNodeInfo` results), mostly looking to fill in some gaps with decorator support. It also refactors a bit in the observer code and updates the changelog a bit. (I'm trying to avoid non-code only commits outside of a PR since we're now publishing to npm automatically with each push to dev.)

1. You can get to decorators to interact with them using event and node info objects e.g. `ractive.getNodeInfo('div').decorators.someDecorator` will give you access to the decorator return object from `<div as-someDecorator />`.
2. There are context-relative `observe` and `observeOnce` methods that can take relative keypaths. This makes observation from within a computed context possible.
3. There's an experimental `raise` method (`raise(eventName, eventObject, ...args)`) available on context objects, so you can trigger events on VDOM elements. So if you have `<div on-foo="@.set('foo', 'bar')" as-fooable />`, then the `fooable` decorator can do `this.getNodeInfo(node).raise('foo');` to trigger the on-foo event on the VDOM node. Note that these are _not_ DOM events. They only look up the VDOM tree for a handler, and while they will look at parent elements until they find a handler, they don't bubble beyond the first handler in this implementation.

#2217

## Is breaking:
No

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 # 0.9.0 (unreleased)
 
+* Bug fixes
+	* Observers on uninitialized data may be added during the `config` event (#2725)
+
 * Breaking changes
 	* All deprecations have been removed, including proxy events with args, un-prefixed method events, decorator="...", transition="...", the ractive.data getter, partial comment definitions, and lifecycle methods like `init` and `beforeInit`.
 	* The template spec is now a bit simpler after the removal of deprecations, and templates parsed with previous versions of Ractive are no longer compatible.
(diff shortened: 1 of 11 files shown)


ghost:
> There's an experimental `raise` method

So essentially a `fire` for the `getNodeInfo` but not quite? 😄

evs-chris:
> So essentially a fire for the getNodeInfo but not quite?

😆 It's kinda the template counterpart to `fire`. `fire` pushes events up the API side of ractive, where `raise` pushes events up the template side. You can think of it as `dispatchEvent` or simulant, but without the extra cross-browser headaches because it's ractive VDOM only.

I considered naming it `fire`, but since it in no way triggers `on` listeners directly, I figured that would be confusing. It _will_ trigger proxy events indirectly, if they're set up like this: `<div on-foo="bar" />` `ractive.getNodeInfo('div').raise('foo')`.

Issue #2001, opened 2015-06-17
Ractive.stringify(parsedTemplate)


fskreuz:
As discussed in https://github.com/ractivejs/ractive/issues/1998, proposing `Ractive.stringify()`, a counterpart of `Ractive.parse()`. Takes in a parsed template and returns the template in string form.

ISNIT0:
Writing a very limited one is fairly easy...
http://jsbin.com/julitaqave/edit?js,console

(Originally came from this:http://jsbin.com/vewefamaje/edit?html,js,console,output)

I may sit down and start one properly at some point using all of these:
https://github.com/ractivejs/ractive/blob/db7a5d61730ad72857ae355597838b3d77f5bc28/src/config/types.js

rstacruz:
why not release this as a separate library like `ractive-stringify`?

ISNIT0:
Well it could always just be added onto the Ractive prototype manually:
Ractive.prototype.strinigyf = function(){...}
I think once the functionality is there, various different things can be done.

Joe

evs-chris:
This turns out to be pretty complicated, and it doesn't fit in core very well. There's a _mostly_ working implementation over at #2766, which I don't think is going to be merged. I'll leave the branch though, in case there's any interest in doing this.

For the original use-case, internationalization of templates, I think it'd actually be a good deal easier to clone the template and then walk it swapping out text nodes. The only thing you'd care about swapping languages on would be text nodes inside elements (maybe a few attributes, but elements would definitely be the major chunk), and walking the element hierarchy is really easy. The root fragment is an array, text nodes are strings in the fragment array, elements are objects with `node.t === 7`, and the  element fragment array is at `node.f`.

I'm going to close this, as I don't think it will happen in core, and if it _does_ happen as an external library, there's a pretty good start in #2766.

Issue #1786, opened 2015-03-10
Allow subsets of templates to be used as templates


Crisfole:
I'd like to use a node's content partial as a template in a decorator.  Here's [a gist with a good use-case](https://gist.github.com/Crisfole/19a5ea343fd152f2ba4e) for that in particular.

Right now, when I try to render a `tooltip` (from the above gist) I get two errors:

```
ractive.js:4501: Uncaught Error: Mismatched template version! Please ensure you are using the latest version of Ractive.js in your build process as well as in your app
```

and

```
ractive.js:9820: Uncaught TypeError: Cannot read property 'teardown' of undefined
```

I don't think I'm doing anything crazy; it seems like something that should be doable...

martypdx:
The content partial is probably not fully formed to the [template spec](https://github.com/ractivejs/template-spec), or it's because you're using just the first element of the array, `template: r.partials.content[0]`. Try wrapping it like:

```js
{
  v: 2, /* Template spec version */
  t: r.partials.content
}
```

Crisfole:
I tried using `r.partials.content` and it didn't work either.

I never tried manually setting the template version...doesn't seem like something I'd expect to have to do, I'll give it a whirl.

martypdx:
@Crisfole [This] seems to work. 

Flagging as an enhancement to see we could either eliminate the need for the version in that case, or provide the version so it could be passed in rather than hard coded.

evs-chris:
You can use fragment arrays from within templates with partials, so I'm going to go with "close enough" on this.

Issue #1535, opened 2014-12-01
Enhancement suggestion: inherit attributes for components/widgets


dagnelies:
Hi,

Components appear to be extremely great stuff. However, as I already mentionned in #1454, they are  IMHO crippled by one detail. One of the main pitfalls for components is that they don't behave like "normal" html dom nodes. You can't set `<widget style="width:200px" />` except if it's specifically defined in the template. This, in turn, leads to component templates like this:

```
<a href='{{href}}' media='{{media}}' target='{{target}}' type='{{type}}' id='{{id}}' class='{{class}}' style='{{style}}' draggable='{{draggable}}' {{#hidden}}hidden{{/hidden}} title='{{title}}' onclick='{{onclick}}' ondblclick='{{ondblclick}}' onmouseover='{{onmouseover}}' onmouseout='{{onmouseout}}' ondrag='{{ondrag}}' ...
```

...even for trivial components you have huge chunks of "attribute forwarding". It is definitely impractical and error-prone.

The enhancement suggestion would be a flag
 `inheritAttributes: true/false`
When set, all top level elements would inherit all default html attributes, except if it's overriden by some other value in the template itself. If it helps in any way, restricting components to a single DOM element would be ok too in my opinion. This would drastically help making simple components/widgets.

Cheers

andreujuanc:
@dagnelies  it'd be nice to have something like this already implemented, but it's not so easy to design, i can image. What if I want to set an attribute jut for the component and not for the first node of the template? or maybe i have 10, and want just half? Someone said about a switch in the options, but I think it can be done as said before using some kind of decorator.

**Implementation:**
`<widget style="width:200px" config1="12345" />`

**Template:**
`<a href='{{href}}' media='{{media}}' target='{{target}}' type='{{type}}' id='{{id}}' class='{{class}}' style='{{style}}' draggable='{{draggable}}' {{#hidden}}hidden{{/hidden}} title='{{title}}' onclick='{{onclick}}' ondblclick='{{ondblclick}}' onmouseover='{{onmouseover}}' onmouseout='{{onmouseout}}' ondrag='{{ondrag}}' ...`

I dont want config1 to be as attribute in my <a>, so I can see two ways to implement this:
1. Have a list of reserved keywords for attributes, exactly those standard for html elements, so they get automagically "copied".
2. Have a prefix for anything else you need copied but it will get things a bit ugly. :(

I like number 1. this way forces developer not to use html keywords as variable names.

[...]

PR #2766, opened 2016-11-16
Template inspection/manipulation helper - RFC WIP - 0.9

evs-chris:
This is my first pass at a template inspection/manipulation helper. It adds a new method, `Ractive.template` that takes a template, or some portion thereof, and wraps it in a helper class that makes it easier to navigate a template hierarchy. For now, it also does a defensive copy, so you can't accidentally modify a template for future instances. Also bear in mind that this is a work in progress, so there are some rough edges. I'd like to get some feedback before I wander too far off.

I think the classes are fairly self-explanatory, and it's only one file to look at for what's available, so I'll just throw out some examples of what I'm aiming to do:
```js
// verify that certain attributes are passed to a component (in onconfig, I think)
const required = ['foo', 'bar', 'baz'];
const tpl = Ractive.template([this.component.template]).elements[0]; // this is actually the element on the component VDOM item that's creating this instance
const provided = tpl.attributes.map(a => a.name); // attributes are only top-level, non-conditional attributes (no events, decorators, etc as those have their own accessors)
required.forEach(attr => {
  if (!~provided.indexOf(attr)) console.warn(`Please supply ${this.component.name} with a ${attr} attribute.`);
});

// gather up non-config attrs and put them in a partial
// using tpl and required from above
// the create functionality is proposed - it's not implemented or even close to final
const extra = tpl.attributes.filter(a => ~require.indexOf(a.name)).map(a => tpl.create({ type: 'attribute', name: a.name, value: { type: 'interpolator', reference: a.name } }));
// we now have an array equivalent to `attrA="{{attrA}}" attrB="{{attrB}}"` etc
// we can turn it into a useful template (wrap up the fragment array) and assign it to a partial name using
this.partials.attributes = Ractive.template(extra).toTemplate();

// you can also access other handy stuff, like inline partials (<cmp>{{#partial foo}}...{{/partial}}</cmp>)
tpl.partials; // this is the array, but it also has named properties assigned
tpl.partials.foo; // this is the foo inline partial
```

For a theoretical tab component:
```html
<TabBox>
  <Tab title="Hello, {{name}}">See, I used your name in a template for this tab entitled "Hello, {{name}}". You could use the title template to do validation icons or message counts or whatever.</Tab>
  <Tab title="Tab 2">And here's another tab...</Tab>
  <Tab title="Yet another">I could go on, but I won't.</Tab>
</TabBox>
```
```js
// in the tab box onconfig
const tpl = Ractive.template(this.partials.content);
const tabs = [];
tpl.elements.forEach(e => {
  if (e.name === 'Tab') {
    const tab = { content: e.content.toTemplate() };
    tabs.push(tab);
    if (e.attributes.title) tab.title = e.attributes.title.value.toTemplate();
    else tab.title = Ractive.template([`Tab ${tabs.length}`]).toTemplate();
  }
});
// this component would use inline partial expressions ({{#each tabs}}...{{yield .title}}...{{yield .content}}...{{/each}}) to render given data
// you could also have an `addTab` function that accepted a title and content and just push them to the tabs array
// closing tabs would be as easy as splicing them out of the tabs array
this.set('tabs', tabs);
```

As you can see, there are still some rough places, and again, please throw in some feedback on how this could be improved or if it's even any better than just manually munging templates when you just have to.

The last thing that this gets you is a stringify for parsed templates, so you can go from `{v:4,t:[{t:2,...},...]}` to something that resembles your original template using `Ractive.template(someTemplate).toString()`. There are some corner cases probably won't ever come back, like method call spreads, but most of the common template constructs are covered.

Known TODOs:
* [ ] create/insert/append template items. Some mutation is available, but not much.
* [ ] Put `toTemplate` method on arrays of proxy objects e.g. `tpl.elements[0].content.toTemplate()`. Maybe add another helper that creates a new empty template.
* [ ] Fix `toString` of `if/elseif/else` sections
* [ ] Tests!

#1535 #1786 #2001

## Is breaking:
No

## Reviewers:
Anyone interested in this sort of thing. I'm not sure that this isn't too heavy to add to core, but if components would use this heavily, it would be better in core than as a strange extra thing that lots of stuff required or tried to supply on their own.

After looking at it a bit, I'm definitely not sure that this is any better than manually munging templates. Then again, I'm familiar enough with the source code that it's not too hard for me to keep track of how the templates are formed and how they need to be abused to accomplish my goals :smile:

### DIFF
--- a/src/Ractive.js
+++ b/src/Ractive.js
@@ -12,6 +12,7 @@ import initialise from './Ractive/initialise';
 import { getCSS } from './global/css';
 import { escapeKey, unescapeKey } from './shared/keypaths';
 import { joinKeys, splitKeypath } from './Ractive/static/keypaths';
+import readTemplate from './view/helpers/template';
 
 export default function Ractive ( options ) {
 	if ( !( this instanceof Ractive ) ) return new Ractive( options );
@@ -56,6 +57,7 @@ defineProperties( Ractive, {
 	splitKeypath:   { value: splitKeypath },
 	unescapeKey:    { value: unescapeKey },
 	getCSS:         { value: getCSS },
+	template:       { value: readTemplate },
 
 	// support
 	enhance:        { writable: true, value: false },
--- a/src/view/helpers/template.js
+++ b/src/view/helpers/template.js
@@ -1,0 +1,237 @@
+import { ELEMENT, COMPONENT, PARTIAL, YIELDER, SECTION, INVERTED, INTERPOLATOR, ANCHOR, ALIAS, TRIPLE, ATTRIBUTE, EVENT, DECORATOR, BINDING_FLAG, TRANSITION } from '../../config/types';
+import { SECTION_IF, SECTION_IF_WITH, SECTION_EACH, SECTION_UNLESS } from '../../config/types';
+
+const delimiters = {
+	static:       [ '[[', ']]' ],
+	staticTriple: [ '[[[', ']]]' ],
+	plain:        [ '{{', '}}' ],
+	triple:       [ '{{{', '}}}' ]
+};
+
+class Template {
+	constructor ( fragment, owner ) {
+		if ( !fragment ) fragment = [];
+		if ( fragment.t ) {
+			this.expressions = fragment.e;
+			fragment = fragment.t;
+		}
+		this.owner = owner;
+
+		// start the root with a defensive copy so that any modifications don't affect other instances later
+		this.fragment = !owner ? JSON.parse( JSON.stringify( fragment ) ) : fragment;
+	}
+
+	get content () {
+		return this.fragment.map( i => {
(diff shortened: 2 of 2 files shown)


dagnelies:
IMHO this doesn't look very intuitive/simple and misses its mark. It's more like internal stuff, but not really something for the average ractive user.

ghost:
Sounds like the kind of stuff debuggers could possibly use alongside `getNodeInfo` and the newer APIs you made.

evs-chris:
I've come to the conclusion that this needs to be a separate library, as the only dependency on the Ractive guts is the type config that has the definitions for numbers that correspond to node types. Those can just as easily be copied over or inlined. I still think that having attribute munging (declared and optionally required attributes) functionality for components would be a good built-in, so I'll see if that can be done reasonably in another PR.

PR #2768, opened 2016-11-18
Fixes #2033: Shows warning when binding a triple curly

### LINKED ISSUE #2033: no two-way binding with triple curlies?
In this fiddle:
http://jsfiddle.net/ejruypwc/

if you type in the top textarea, there is the normal two way binding, but not in the bottom one.

Is this the intended behavior with triple curlies?

(Sorry if this should have gone on SO or elsewhere)

ghost:

Logs a warning when a triple-curly is used to bind to a textarea.

Also cleans up the binding check to be a bit more structured and easy to follow through.

#2033 

## Is breaking:

No... at least according to the tests. 😄 

## Reviewers:

Anyone who knows how binding works.

### DIFF
--- a/src/view/items/element/binding/selectBinding.js
+++ b/src/view/items/element/binding/selectBinding.js
@@ -1,4 +1,4 @@
-import { INTERPOLATOR } from '../../../../config/types';
+import { INTERPOLATOR, TRIPLE } from '../../../../config/types';
 import { warnIfDebug } from '../../../../utils/log';
 import CheckboxBinding from './CheckboxBinding';
 import CheckboxNameBinding from './CheckboxNameBinding';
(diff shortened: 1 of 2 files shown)


evs-chris:
Would it make sense to just warn on any bindable attribute that has a single triple rather than an interpolator? That could simplify the check a bit and have it apply to _everything_, so you could catch stuff like `<input value="{{{foo}}}" />` and `<checkbox name="{{{list}}}" />`, which are also probably mistakes. You could add a parameter for the element name to `isBindable` and split the giant conditional into three branches (intial - single content not an expression, is an interpolator, and is a triple) to bail early, return true, or warn and return false?

ghost:
Yup. That makes sense. 😄

Still trying to wrap my head around how triples differ from doubles internally.

evs-chris:
There's really not a whole lot of difference to ractive's guts. The main difference is conceptual: that triples hold html where plain interpolators hold strings/numbers/not-html :smile:. Ractive _could_ bind using triples, as the interpolator is really just a container for a reference, as is a triple. It's just the presumed output that goes along with a triple doesn't work with attributes.

ghost:
@evs-chris Just to verify, a Binding can only happen on an attribute? I'm only seeing special elements in the `binding` directory and the README suggests the same as well.

evs-chris:
That's correct. The only one that allows a non-attribute binding is textarea, and it just swizzles its template around to make it an attribute binding.

ghost:
@evs-chris is it ok now? 😄

evs-chris:
Yep, LGTM! :+1:

Issue #1458, opened 2014-11-10
Events hash like Backbone


victorwpbastos:
Is there a way to configure events like Backbone does? Example:

``` js
events: {
    'click .btn-save': 'funcName'
}
```

Issue #2566, opened 2016-06-10
*.eventname does not fire in own namespace


lnryan:
Hiya,

I searched but did not find this bug. Apologies if i simply didn't search well enough. I'm newish to searching git issues.

**Description of the bug:**
I find the above mentioned behavior counter intuitive. I would not expect to have to separately register an event for when it fires in its own container and when it fires in a subordinate container.
eg `ractive.on('myevent *.myevent')`, which is what i'm finding.More exactly, since `this` and `.` can generally be used in ractive, I would expect `this.`/`{emptystring}.` to be among the valid prefixes covered by *.

I would rather expect to use a special notation to indicate i'm _only_ interested in it firing in the current container or _only_ interested in subordinate container firings. 

Perhaps `*.myevent` could be distinguished from a more specifically child only notation? 

**Versions affected:**
0.7

**Platforms affected:**
Chrome

**Reproduction:**
_this is exactly how i found this. Obviously, switching around which lifecycle event it's in may have nothing to do with this_
1. Define an `oncomplete` method for a ractive template. 
2. In `oncomplete` add an event with `this.on('*.myevent',function(event){ alert('I fired!'); });`
3. In your template add a button `<button on-click="myevent">Pick Me!</button>`
4. Create a component and include it in your template `<mycomponent />`
5. In the template add a button `<button on-click='myevent'>No, Pick Me!</button>`
6. When you view this, the button within the component will fire the event, but the one within the original template will not.

**Code:**
See steps to reproduce

evs-chris:
I think it seems reasonable that `this.on('*.foo')` should fire on `this.fire('foo')` too. @martypdx?

heavyk:
I dunno, but personally I do not thing it should be so... I always saw the `.` like a namespace separator... so, `this.fire('clicked')` shouldn't fire for `this.on('*.clicked')` unless it was `this.fire('foo.clicked')` cause I would want these to be different:
 - `<button on-click="clicked">...</button>`
 - `<button on-click="button1.clicked">...</button>`

not sure exactly, but I may have to change some code if this were to change.

guilhermeaiolfi:
I think it's a valid change. If you consider ```clicked``` as ```this.clicked``` (```this``` as namespace), you would expect ```*.clicked``` to be fired.

Issue #2323, opened 2015-12-17
Adding a start & stop to events?


donnyv:
I know right now you can do this.

``` javascript
var handler = function () {
  alert( 'Activating!' );
};

ractive.on( 'activate', handler );

//later
ractive.off( 'activate', handler );
```

But it would be handy to be able to do this.

``` javascript
var handler = function () {
  alert( 'Activating!' );
};

var listener = ractive.on( 'activate', handler );

// later
listener.stop(); // stops the event from firing

// more later
listener.start(); // starts it back up again
```

This lets me pass around the listener without me having to make sure the handler is available everywhere.

martypdx:
Depends on whether `start` and `stop` mean a) unsubscribe or b) just stop calling my handler.

For the a) case (see http://jsfiddle.net/2xm64571/), you can do:

```js
const _on = Ractive.prototype.on;

Ractive.prototype.on = function( eventName, callback ) {
	var cancel = null;
	const listener = {
		cancel() {
			if ( cancel ) cancel();
			cancel = null;
		},
		pause() {
			this.cancel();
		},
		resume() {
			if ( !cancel ) register();
		}
	};
	
	const self = this;
	function register() {
        cancel = _on.call( self, eventName, callback ).cancel;
	}
	
	register();
	
	return listener;
};
```

For the b) case (see http://jsfiddle.net/6emw3w03/):

```js
'use strict';
const _on = Ractive.prototype.on;

Ractive.prototype.on = function( eventName, callback ) {
	var notify = true;
	const listener = {
		pause() {
			notify = false;
		},
		resume() {
			notify = true;
		}
	};
	
	var self = this;
    listener.cancel = _on.call( self, eventName, function(){
		if ( !notify ) return;
		callback.apply(self, arguments);
	}).cancel;
	
	return listener;
};
```

Issue #1182, opened 2014-09-04
Silent reset and silent set 


batmandarkside:
### proposal to make silent reset and set model

``` javascript
this.observe('filtersData', function (value) {
    if (value) {

    }
}, {init: false});

this.set({filtersData : {}}, { silent : true })

// filtersData don't observe
```
### this.data.filtersData = {}; does not work if you later make

``` javascript
 this.set({filtersData : { test : 'test' }  })
```

hezki:
+1 this

guilhermeaiolfi:
Yeah, that would be very useful in something I'm trying to do. So +1.

martypdx:
@hezki , @guilhermeaiolfi - can you give an example or more explanation as to how this would be useful?

guilhermeaiolfi:
Sure, I was creating a decorator for ckeditor and wanted to keep the model in sync with the ckeditor instance value. To achieve that I had to observe the model for changes (model  -> ckeditor) and set the model when ckeditor changed (ckeditor -> model). 

The problem is that when I set the model it would trigger the observer and ckeditor would re-set the value it already had. Losing the cursor position and some other issues. 

I solved that by using a variable to know when I was set in that particular case so the observer - even tho it was called - wouldn't do anything.

I had similar cases using ember.js and I think they are planning a mechanism to pause and resume observers. That would be cool too. But maybe for those simple cases the silent mode is easier to understand and use.

hezki:
like {init:false}, if you want replace all data without raise obserev events

if, example, i observe 'foo.bar' and i want replace foo object without raise 'foo.bar' event similar init= false

nb. sorry for my bad english.

maslianok:
+1

maslianok:
Well, you can use something like this:

```javascript
this.observe('data', function (n, o, k) {
    if (this.silent) return;
    alert(n);
}, {init: false});


this.set('data', 1); //alert(1)

this.silent = true;
this.set('data', 2); //do nothing
this.silent = false;
```

Rich-Harris:
Sorry for the slow response. I think that @maslianok's approach is probably the best one - it's what I do when I need to prevent this kind of 'blowback', and it's how adaptors typically guard against infinite loops.

[...]

PR #2769, opened 2016-11-18
Event-ish refactor: suspendable listeners and this. namespaced events - RFC

evs-chris:

This does a few different-yet-related things:

### Silence-able observers

There's been an open issue for a while requesting the ability to temporarily stop observers from firing  (#1182), so this adds `silence()`, `isSilenced()`, and `resume()` methods to the handle returned from `ractive.observe()`. I went with `silence` over `pause` or `suspend` because the observer still gets changes, it just doesn't fire the notification.

This could conceivably be extended to `cork`-like functionality where changes are stashed until `uncork`ing (probably with `resume` too) or a different form of `silence` where the observer ignores notification, and thus, keeps the old value in place for future `resume`. All of that seems a bit YAGNI for now though, whereas built-in locking for observers would be handy in a number of pretty common scenarios. Opinions?

### Silence-able events

Similar to observers, there's been a request around for a while to temporarily stop events from firing (#2323), so this adds the same `silence()`, `isSilenced()`, and `resume()` methods to the handle returned from `ractive.on()`. Again, I went with `silence` because nothing is torn down while the listener(s) is silenced; they just don't fire. It also mirrors the observer version nicely.

### Event listener other stuff

This also fixes a bug with multiple events subscribed with the same listener at the same time e.g. `this.on('foo bar baz')` wherein the returned handle's `cancel()` method wouldn't actually cancel anything.

It also refactors the `on()` method to be a bit less branchy, like the recent `observe()` refactor.

### Events with an initial implicit `this` namespace

One of the things that seems to be a bit counter-intuitive when first dealing with event bubbling is that `ractive.on('foo')` and `ractive.on('*.foo')` won't both be triggered by `ractive.fire('foo')` (#2566). It seems that the general assumption is that `*.foo` means all the `foo`s, which makes sense to me too. So this adds an implicit `this.` namespace to all intial events, meaning that `ractive.fire('foo')` will trigger any of these listeners: `foo * this.foo *.foo this.*`. It notably _won't_ fire `*.*` because that would mean that any lifecycle events would start triggering `*.*` listeners. That may not be a bad thing, but it breaks tests. I'm not sure that it's intuitive behavior, and I can't really see how all-wild listeners are particularly useful anyways. Opinions?

I really wasn't aware of this before I started poking around this code, but you could already fire namespaced events on your own by using names with `.`s in them. Given that, I considered adding `**` flavors of listeners that would "consume" multiple path segments in the `**` portions, but for now that seems a bit YAGNI too. Again, opinions?

This refactor also (kinda) removes one of the vestiges of versions < 0.8, - the star map. The star map helpers were only used by events as of 0.8, as the Model family took over the matching of wildcard keypaths. The star map helpers basically create a cartesian product of a keypath and an equal-length all-star keypath ( `a.b.c` <-> `*.*.*` ), which is still needed for finding matching event listeners. So there's now a slightly simplified, still-caching (since they're not exactly cheap to create and lifecycle events mean that they're not exactly rare - I'm looking at you `change`) version of star maps moved into the event firing code.

#1182 #2323 #2566 #1458

## Is breaking:
Kinda sorta. If you were relying on `foo` and `*.foo` listeners catching only `this` instance events and only child component events respectively, this is breaking. Otherwise, which seems to be the more intuitive assumption based on issues and SO questions, this would actually be a bugfix. If `*.foo` _must_ only be triggered by child events, you can now check the source ractive against `this` to be sure they're different.

## Reviewers:
Yes, please.


mefjuu:
It seems that subscribing multiple events as object map fixes https://github.com/ractivejs/ractive/issues/1458 also.

evs-chris:
@mefjuu it does now :smile:.

I've added changes to resolve #1458:

### Events hash

With this, you can specify an `on` hash in your options that takes key/value pairs. The keys may be anything that's valid to pass to `ractive.on()`, so you can stick the same handler on multiple events with a space separated list e.g. `'foo bar baz.bat'`. The values may be functions or an object with a `handler` function and other options. At this point "other options" for events is just `once: true`, which subscribes the event using `ractive.once()` instead of `ractive.on()`.

Components/extended instances _add_ to their parent event hash, so

``` js
const cmp = Ractive.extend({ on: { foo() { ... } } });
const r = new cmp({ on: { foo() { ... } } });
r.fire('foo');
```

will trigger the foo methods from both configs. The reasoning behind this is that the `cmp` component set the handler up for a reason, and as such, it should not be overridden accidentally by an extension or instance. Handlers are installed from root component to instance, so the root handlers are called first. If necessary, we could specify that one of these listeners returning `false` would halt the subscriber notification, but I don't think that will really come up.

### Observers hash

As with events, you can specify an `observe` hash in your options that takes key/value pairs. Like events the keys can be any string that can be passed to `ractive.observe()`. Also like events, the values may be a function or an object with a `handler` function and additional options. Here's an example with each:

``` js
new Ractive({
  observe: {
    foo() { /* I will fire at init as the observer initializes */ },
    'bar baz': {
      handler() { /* I will not fire at init, and I will not fire until dom manipulation is done */ },
      init: false,
      defer: true
    },
    bat: {
      handler() { /* I will only fire once when bat first changes */ },
      once: true
    }
  }
});
```

As with events, the observer hash is additive, so base component observers and instance observers are all subscribed, with the base observers subscribing first.

ghost:
Yasss! Events hash! I remember requesting that one. Few questions:
- I assume `events` and `observers` are taken names already? Would be more intuitive though.
- Can lifecycle events be put there? Or are those specifically for user-defined events?
- I encountered `this._super()` while reading the components docs. It appears to be used by lifecycle events to call overridden parent lifecycle events. Would it make sense to do the same for regular events?

And a few other things:
- A warning as to why `*.*` is not supported might be good though.
- Starmap... the code that generates all possible combinations of keypaths given a starred keypath? 😁

evs-chris:
1. `events` is taken. I went with `on` and `observe` because they match the method name, and it seemed fitting as they are _just_ shortcuts for calling `on` and `observe` after the `config` event fires.
2. Any old event can be here, including lifecycle and wildcard.
3. I think the `_super()` is only for dealing with the `onwhatever` method flavor of event handling. With this, that stuff could actually go away. Even if you don't call `this._super()` in an `onwhatever` method, the event will still trigger an listeners that super may have installed.
4. Warn about not supporting it, or just don't special-case it and note it as a breaking change?
5. Well sort of :smile:. It's all possible matching starred keypaths for a given absolute keypath. The star map for `a.b.c` is `[ 'a.b.c', '*.b.c', 'a.*.c', 'a.b.*', '*.*.c', '*.b.*', 'a.*.*', '*.*.*' ]` - or the cartesian product of `a.b.c` and `*.*.*`. It's used when a namespaced event fires, say `foo.bar`, to find any listeners that could match, so when you have `ractive.on('*.bar', ...)`, the `foo.bar` event will call that listener.

evs-chris:
Addenda:
1. I also went with `on` and `observe` to keep them consistent _because_ `events` is taken.
2. Any old event, except things that take place on or before `onconfig` :smile:. I _think_ I could move the subscription point to just before `onconstruct`, but I would probably have to split events and observers up. Should I do that?

Issue #2770, opened 2016-11-18
`noIntro` counterpart for outro transitions (`noOutro` ?)


giovannipiller:
## Description:
I think it'd make sense to have a `noIntro` counterpart for outro transitions (`noOutro` ?).
I'm running into many situations in which transitions are only important while the component is on-screen.

What do you think?

## Disable transitions from template:
As a side note: is it possible to conditionally disable specific transitions from templates?

On 0.7.x I used to do something like `<div intro="{{shouldTransition ? 'fade' : ''}}">`, but that currently throws a warning on 0.8 (`Missing "" transition plugin`). Any better way to do it?

Not sure if it's even doable in 0.8's `...-in-out` syntax, doesn't seem like it.
ex. `<div fade-in="{{shouldTransition}}">` ?

evs-chris:
For the `noOutro` side, I think having transitions that can be configured _not_ to run if there is already a parent element transitioning is the answer (#2414).

For conditional transitions in 0.8, the recommended way is `<div {{#if shouldTransition}}fade-in{{/if}}>`. You can also use a plain section for slightly more brevity `<div {{#shouldTransition}}fade-in{{/}}>`.

giovannipiller:
> For the noOutro side, I think having transitions that can be configured not to run if there is already a parent element transitioning is the answer

I still think that `noOutro` would cover a different use-case though.

---

### Component1
```
...
{{#unless hide}}
   <div fade-in-out>...something...</div>
{{/unless}}
...
```

### Component2
```
<section>something else</section>
```

### Page 
Data has `activeComponent: 1`.
```
{{#if activeComponent === 1}}
  <Component1/>
{{/if}}

{{#if activeComponent === 2}}
  <Component2/>
{{/if}}
```
---

Executing `Page.set('activeComponent', 2);` will result in both components temporarily being displayed in the page.
This because `<Component1/>`'s internal transitions will have to be completed before its node gets removed.

`noOutro` seems like a logical counterpart to `noIntro`, and would solve these cases where there's no parent transition involved.

evs-chris:
Ah, that makes sense. I've added a `noOutro` flag to #2773, which disables outros for a component that is unrendering. The inheritance mirrors that of `noIntro` in that if there's an unrendering parent that has `noOutro` explicitly set, components belonging to it that are unrendering and have no explicit setting will use that of the parent.

giovannipiller:
[...]

Issue #2414, opened 2016-02-26
Animations delay removing parent element


ADantes:
If any of an element's children contain transitions, that element isn't removed until those transitions complete.  It feels like it would be more in line with expected behaviour to just remove the element immediately and ignore any child transitions.

http://jsfiddle.net/reLcn1dz/2/

In the above example:
- switch between "Tab One" and "Tab Two" and back to "Tab One"
- Observe that the transition happens instantaneously
- Check the checkbox and wait for intro transition to complete
- Switch to "Tab Two"
- Observe that both #tab-1 and #tab-2 are present and visible while the outro transition plays.

(In my specific use case, having both of these visible breaks my page layout)

martypdx:
Issue is that sometimes you _do_ want the parent to wait for the children. So it's a matter of an api that can work for various situations.

With current edge you can do something like:

`intro-outro="slide:{{ tab !== 'One' ? { duration: 0 } : {} }}"` 

to prevent the transition.  See http://jsfiddle.net/reLcn1dz/3/, I changed your example a bit so you could see the individual item transition on add.

Once #2369 is merged, you will be able to write conditional transitions using:

 `{{#if tab === 'One'}}intro-outro="slide"{{/if}}"`

ADantes:
I don't think this is a practical solution, once you start dealing with component hierarchies. It did, however, lead me down a few paths I like better.

Approach #1: "none" transition that ends immediately, hiding the element
http://jsfiddle.net/reLcn1dz/4/
(This still subjects you to https://github.com/ractivejs/ractive/issues/2027 however, since the animation time hasn't changed)

Approach #2: Disable animations during a component's teardown step
http://jsfiddle.net/hsLyomb6/
(This is probably the better solution, although is slightly inconvenient since disabling transitions does not cascade down to child components)

I still am of the opinion this is a bug, possibly a side effect of https://github.com/ractivejs/ractive/issues/2027.

evs-chris:
What if the transitions got notified of whether or not they were the root element being transitioned? Then each transition site could choose to transition before letting the parent go or not.

PR #2773, opened 2016-11-19
Introduce controls for disabling nested transitions - RFC

evs-chris:

Currently when transitions are registered on an element that is being removed at the same time as a parent element that also has a transition, both transitions will run in parallel. This PR introduces two options, one on the transition and one on the Ractive prototype, that allow nested transitions to be skipped so that only the root transitioning element in a chunk of DOM will transition.

The prototype option is `Ractive.defaults.nestedTransitions`, which can also be passed to an instance or extension as `nestedTransitions`. It defaults to true, which maintains the current behavior of edge.

The transition option is `nested`, and when set to false will disable the transition if it happens to be on a child element of an element that is transitioning.

If the element that is being added/removed is the child of an element with a transition that is _not_ currently being added or removed, the flag will have no effect. For instance, with:
```html
<div fade-in-out>
  {{#unless hide}}
  <div fade-in-out="{ nested: false }" />
  {{/unless}}
</div>
```
When the template is initially rendered, the inner div will not transition, as it is a child of the transitioning outer div. Once everything is rendered, setting `hide` to true will cause the nested div to transition out, as it has no transitioning parent. Likewise setting `hide` back to a falsey value will cause the inner div to transition back in. Unrendering the entire template will cause the outer div to transition, but the inner div will not transition.

`noIntro` is closely related to this, so I may try to add a fix for #1853 here as well.

#2414 #2770 (I think) and #1853

## Is breaking:
Slightly. `noIntro` will now be inherited from the parent instance if it's not set on the component and the parent instance is also rendering.

## Reviewers:
@ADantes @giovannipiller and anyone else that's had issues with transitioning trees

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 	* Any partial may be yielded, so yielding non-inline partials will no longer warn.
 	* The same partial may be yielded multiple times.
 	* Events now fire in an initial implicit `this.` namespace. This means that with `this.on( '*.foo', handler )`, `handler` will be called if and component fires a `foo` event or if the `this` instance fires a `foo` event. 
+	* The `noIntro` option now applies to any nested components that are also being rendered, unless they have their own explicit setting.
 
 * New features (experimental - feedback welcome!)
 	* You can now create cross-instance links by passing an options object with a target instance e.g. `this.link('source.path', 'dest.path', { ractive: sourceInstance })`. This covers many of the cases handled by the `ractive-ractive` adaptor in a considerably more efficient manner.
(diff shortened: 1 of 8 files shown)


evs-chris:
I've added a fix for #1853 here as well. Basically if a component doesn't have its own `noIntro`, it will inherit it from a parent if the parent is also rendering. So an entire rendering component tree can have the intros skipped, whereas a single rendering component, say inside a section, will intro normally.

giovannipiller:
What would happen if I had something like this?

---

### ChildComponent
```
<section slide-in-out>..something..</section>
```

### RootComponent
```
<div fade-in-out>
  <ChildComponent/>
</div>
```

**RootComponent with `nestedTransitions: false`.**

---

Will tearing down RootComponent still trigger ChildComponent's transitions?
I don't think I would expect it, just as:
> if a component doesn't have its own noIntro, it will inherit it from a parent if the parent is also rendering

giovannipiller:
Also, stretching things a bit, would it be possible to combine the two options?

### StandaloneComponent with `nestedTransitions: false`
```
<article fade-in-out>
    <header slide-in-out="{ nested: true }">..something..</header>
    <footer slide-in-out>..something..</footer>
</article>
```

Basically, whitelisting specific nested transitions.

In my experience, I find myself pseudo-whitelisting a lot of the times, although that's because I mostly deal with SPA.

Let me know if you need more info on this.

evs-chris:
@giovannipiller that all makes sense, so I've updated this accordingly:

1. If transitions are enabled on the instance, `noIntro` is not in effect, and there is a `nested` flag on the transition, the flag wins.
2. If there is there is _not_ a flag on the transition, then the nearest ractive instance with an explicit setting wins e.g. root with `nestedTransitions: false`, components that don't have `nestedTransitions` explicitly set will get the root setting.
3. If an explicit `false` setting for nesting is found, then the transition will check to make sure it is not a child of another transitioning element before registering its element to transition.

giovannipiller:
Looks great! Can't wait to use this  🎉😄 Thank you!

PR #2775, opened 2016-11-21
Random refactorings based on coverage report

evs-chris:

This is just trying to increase test coverage and slim down the code a bit outside of the legacy stuff. I'm also seriously considering shortening some of the largish property names to see if it makes any impact on minified size. More to come...

None, but additional coverage may turn up a few new ones...

## Is breaking:

Nope.

### DIFF
--- a/sandbox/sample/index.html
+++ b/sandbox/sample/index.html
@@ -40,6 +40,11 @@
 			<li>Edit the template block and the code at the bottom of the page</li>
 			<li>That's it! You're ready to go.</li>
 		</ol>
+
+		<select multiple value="{{foo}}">
+		{{{html}}}
+		</select>
+		<textarea lazy>{{html}}</textarea>
 	</script>
 
 	<!-- JS -->
--- a/src/Ractive/config/runtime-parser.js
+++ b/src/Ractive/config/runtime-parser.js
@@ -19,11 +19,7 @@ const parseOptions = [
 
 const TEMPLATE_INSTRUCTIONS = `Either preparse or use a ractive runtime source that includes the parser. `;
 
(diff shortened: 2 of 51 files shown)


evs-chris:
Looking at it, it looks like there's at least 10k of minification shrinkage available just by dropping the relatively large names that are constantly used throughout the code base back to a few characters - parentFragment, fragment, and template are common offenders everywhere, children and childByKey are common in models, and allowWhitespace, matchString, and matchPattern in the parser. It would affect code readability _a little bit_, but I think parentFragment -> pf, fragment -> f, template -> t, children -> ks, childByKey -> k, allowWhitespace -> sp, and joining matchString and matchPattern into a single match function would not be awful. I know size isn't everything, but 5% of the code being just a handful of common references seems high. Any opinions?

I'd like to find 40k to shave off the minified file size, and I'd like to do so before 0.10, which will largely be an adaptor refactor where, hopefully, the magic and array adaptors will get pulled out into their own plugins. 0.11 is tentatively scheduled as a transition refactor, which may drop a bit of weight because the transition code is a bit... mysterious :smile:. IE8 no longer being a consideration may help too. I'd like to get the minified file size back down below 140k (that's 72k less than 0.8 and 25k less than 0.7.3 for those who are counting) before 1.0, which I think is doable.

ghost:
I'm ok with name changes as long as they still represent what they mean on first read. Reading through code yesterday, and without looking at the README that came with it, I was like... "What is an 'Item' again?". 😄

evs-chris:
I'm going to punt on changing names for minification purposes for now. I keep going back and forth between shortening some of the members in source and adding a sigil to them to allow replacing them safely in a build step.

PR #2784, opened 2016-11-27
Random cleanups

ghost:

Random cleanups:

- Removing the `clean` script since I discovered that the same can be done using [`git clean -fdx`](http://stackoverflow.com/a/64966/575527).
- Removing `./node_modules/.bin` from the scripts. [npm adds `node_modules/.bin` to `PATH`](https://docs.npmjs.com/cli/run-script) anyways.
- Removes [Testling ](https://ci.testling.com/) from package.json since Testling appears to be non-functional.
- Removes [Jam](https://github.com/caolan/jam) from package.json since it appears to be inactive.
- Removes the `component.json` in `scripts/templates`
- Removes the fluff in the root package.json since it's only used for build and not the one published.
- Minor reorganizing of package.json while I'm in there.

An intimidating codebase 😄 

## Is breaking:

- If some build process used the newly added `clean`.
- If anyone is still using Jam as their package manager.
- If anyone is still using Component as their package manager.
- If there's something in `.gitignore` that should have been ignored/not ignored.
- If there's a reason why `./node_modules/.bin` was in the paths in the first place.
- If anyone is using `npm install ractivejs/ractive` (pulls repo as a dependency)
    - Suggesting you use `npm install ractivejs/ractive#TAG` to pull in the built versions.

## Reviewers:

Everyone, in particular those who know what goes on in the CI realm.


evs-chris:
I recently discovered that the package manifest in the root dir isn't the one that gets published to npm. The manifest that gets published is a template in the scripts dir that gets adjusted during the build process.

LGTM :+1:

Issue #2786, opened 2016-11-27
Merging rcu into Ractive's bin (or vice versa)


ghost:
## Description:

Probably self-explanatory but I'll explain anyways. We have Ractive's bin (which at the moment contains `parse`). Then we have rcu which is a module that contains utility functions for pre-parsing stuff. We have utils in two places!

So I was wondering, to avoid multiple tools:

- The bin go over to rcu, drop Ractive bin (make rcu be both importable and as a bin).
- Bring rcu to Ractive's bin, drop rcu (ractive bin with more functions, not sure about importability).
- Bring rcu to Ractive's bin, and have rcu (the module) be a thin wrapper for the bin functionality.

Any thoughts?

evs-chris:
I was thinking about this too, and I'm very conflicted. On one hand, having a bunch of different ways to do the same things is certainty suboptimal. On the other hand, especially for the precompilation case, it makes a ton of sense to have that as a built-in. That seems to be one of the biggest issues people run into, given that iOS browsers like to die horribly when parsing large templates.

Right now, I'm leaning toward adding a simplified component builder into the bin script and just leaving rcu to more tool oriented build processes. I don't think it would take much at all to convert a component definition into an es module, and it eliminates a bunch of complexity if you drop source maps and AMD/CommonJS support.

For 0.9, it's all flagged experimental anyways, so we can play with it a bit to see what works out best.

dagnelies:
I've never used *rcu* directly, so I'll leave it up to your expertise.

I just now that following plugin:
https://github.com/ractivejs/ractive-load
has both a dependency on ractive and on rcu. Sadly, it's broken and unmaintained.

ghost:
I'll close out this one since rcu is a totally different monster, primarily because it's meant to be embedded in other tools, as opposed to being a build-time dependency.

While it is possible to merge everything (nothing is impossible 😉 ) and make Ractive bin more robust, it makes the Ractive package a bit more complicated with dependencies and all.

Issue #1051, opened 2014-07-30
Link to external stylesheets from a component


KP:
The encapsulation of component-specific styles is awesome. Writing CSS directly in these components, however, is less than ideal (can't use CSSComb). It'd be great to be able to link to external stylesheets from inside a component and have those styles be encapsulated.

Was thinking an implementation like this:

``` html
<link rel='ractive' href='style.css'>
```

evs-chris:
The ractive cli helper can parse components in much the same way as rcu, but aims to produce es modules rather than amd or commonjs modules thus punting loading to the user in a build step/runtime loader. It supports external scripts, partials, and stylesheets.

PR #2787, opened 2016-11-28
Add rcu-ish component building to the ractive bin - RFC

### LINKED ISSUE #1892: Define only partials in component files.
We can do this by javascript now:

```
// From http://docs.ractivejs.org/latest/ractive-extend
BasicModal = Modal.extend({
  partials: {
    modalContent: '<p>This is some important content!</p><a class="modal-button" on-tap="okay">Okay</a>'
  },

  init: function ( options ) {
    // wherever we overwrite methods, such as `init`, we can call the
    // overwritten method as `this._super`
    this._super( options );

    this.on( 'okay', function () {
      this.teardown();
    });
  }
});
```

But in component files there is no "partial block", defining partials in the HTML "block" will override the whole template.
How can I define only partials in component files?

evs-chris:

This splits the ractive bin script up into modules that could conceivably be used by third parties. It stays dep-free, so there are a few things that rcu offers that aren't present here, namely source mapping and mangling various forms of dependency loading.

Instead of being very clever, this takes the dumb approach to compiling a component. It merely collects up the relevant files, converts them to the appropriate string form, does a string substitution on the script portions, and plops the result back out to STDOUT or the designated file. This leaves any module craziness up to the component author, but I'd recommend using es module format, as it translates reasonably well into other formats where needed. It's also a bit less cluttered.

Here's what this will do:

1. Read in a base template file from STDIN, a given file, or a given directory with an `index.html` or `index.ractive` and parse it as a template
2. Iterate through the top-level bits of template looking for script, template, style, and link tags
    * script tags that are `text/ractive` or `text/html` and have an `id` or `name` are turned into partials - if the script has a `src`, it is read from the designated file relative to the component file
    * script tags that are `text/javascript` (or untyped) are concatenated in order and used as the basis of the component. It's generally only recommended to have one, but this doesn't argue about having more.
    * template tags that have a `name` or `id` are parsed as partials
    * style tags have their content added to the component css string
    * link tags that are type `text/css` or `component` and rel `ractive` or `component` with an `href` will have the file at `href` read and appended to the component css string
    * any processed tags are removed from the template
3. The extra partials are parsed and have their template set to the appropriate partial id/name and their expressions added to the main template expressions
4. The css has its whitespace collapsed and it is JSON.stringify-ed
5. The script has any `$TEMPLATE` and `$CSS` identifiers replaced with the stringified template and css respectively

The whole chunk of relevant code ended up being ~120 lines, which I think is acceptable for a component compiler. External deps _should_ be as simple (when using ES module format) as `import Ractive from 'ractive'; import OtherComponent from '../OtherComponent`;` or whatever, and the result can be shared with the world by `export`ing it. I believe module bundlers like rollup would work nicely with this sort of format.

#1051 #2786 and it could fix #1892 by allowing a `$PARTIALS` sub that spit out a stringified object with all of the partials, but I'm not sure how useful that is in the context of csp and more powerful yielding.

## Is breaking:
Nope.

## Reviewers:
@fskreuz and anyone else interested in having a Ractive CLI for (hopefully) less painful component building and template pre-parsing.

### DIFF
--- a/bin/lib/component.js
+++ b/bin/lib/component.js
@@ -1,0 +1,166 @@
+const fs = require( 'fs' );
+const path = require( 'path' );
+
+const stringify = require( './parse' ).stringify;
+
+const help = `
+    component
+
+      Read a single file component definition from STDIN or from a file via the
+      -i option and compile it into an ES module to output either to STDOUT or to
+      a file via the -o option. Linked stylesheets will be imported into the
+      component from their external files and included alongside any style elements.
+      Any script tags will be concatenated an used as the JS body of the template
+      file. The remaining HTML will be turned into a template and inlined into the
+      script anywhere $TEMPLATE is encountered. Any aggregated css will be injected
+      into the script as a JS string (JSON.stringify-ed) anywhere $CSS is
+      encountered.
+
+      Any additional script tags encountered with type text/ractive or text/html
+      and an id or name attribute will be automatically turned into partials on the
+      main template. <template> tags with an id attribute will also be turned into
+      partials on the main template.
+
+      script and link tags can specify the path to a file to include directly using
+      the src and href attributes, respectively. Paths to remote files should be
+      relative to the source file.
+
+      -i, --input {file} - file to parse, otherwise STDIN
+          This may also be a directory containing a component in index.html.
+      -o, --output {file} - file to write, otherwise STDOUT
+      -x, --nocsp - don't output expression functions
+      -d, --delimiters {open} {close} - set the plain mustache delimiters (default {{ }})
+      -s, --static {open} {close} - set the static mustache delimiters (default [[ ]])
+      -r, --triple {open} {close} - set the triple mustache delimiters (default {{{ }}})
+      -p, --static-triple {open} {close} - set the static triple delimiters (default [[[ ]]])
+`;
+
+function build ( Ractive, string, output, opts, base ) {
+	const rootOpts = Object.create( opts );
+	rootOpts.interpolate = { script: false, style: false, template: false };
+	const tpl = Ractive.parse( string, rootOpts );
+	const partials = {};
+	let script = '';
+	let style = '';
+
+	// walk the template finding any script, style, or link tags to process them as appropriate
+	let i = tpl.t.length;
+	while ( i-- ) {
+		const item = tpl.t[i];
+		// top-level elements
+		if ( item.t === 7 ) {
+			if ( item.e === 'script' ) {
+				const type = getAttr( 'type', item );
+				const id = getAttr( 'id', item ) || getAttr( 'name', item );
+				const src = getAttr( 'src', item );
+
+				if ( id && ( type === 'text/ractive' || type === 'text/html' ) ) {
+					if ( !src ) {
+						partials[id] = item.f[0];
+					} else {
+						partials[id] = readToString( base, src );
+					}
+				} else if ( !type || type === 'text/javascript' || type === 'application/javascript' ) {
+					if ( !src ) {
+						script = item.f[0] + script;
+					} else {
+						script = readToString( base, src ) + script;
+					}
+				}
+
+				i = drop( i, tpl.t );
+			} else if ( item.e === 'template' ) {
+				const id = getAttr( 'id', item ) || getAttr( 'name', item );
+				if ( id ) {
+					partials[id] = item.f[0];
+				}
+
+				i = drop( i, tpl.t );
+			} else if ( item.e === 'style' ) {
+				style = item.f[0] + style;
+
+				i = drop( i, tpl.t );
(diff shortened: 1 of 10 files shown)


ghost:
Reading lodash earlier to see how it split up its functions into separate modules while keeping one with everything.

We can have Ractive loaded as is, like how we do:

```
// The regular lib gets returned
var Ractive = require('ractive'); 
```

But then have rcu functions be loaded like this. They're separate modules from the main Ractive lib, kinda like the way how lodash has one gigantic lodash.js, and have multiple independent modules for its functions.

```
var parse = require('ractive/parse');
var make = require('ractive/make');
var generateSourceMap = require('ractive/generateSourceMap');
// and so on...

// Use like rcu
```

Then the bin just becomes a CLI wrapper for those same functionality, except it just takes in options from the terminal instead of a node script.

```
// stdin input, stdout output
ractive make input.html > output.html

// file input, file output
ractive make -i input.html -o output.html

// All the rest
ractive parse ...
ractive sourcemap ...
// and so on
```

That way, it can both serve those who prefer the CLI way of doing things and the ones who make Ractive part of their node tools. Two birds with one stone n_n)b

evs-chris:
I think I'm only one step off of a reasonably consumable API stuck right in the ractive module then. I just need to move the output handling up a layer and let the component module work directly with strings. It will still have to load files in some way. I suppose the polite way to do that would be to accept a loading function like rcu and promisify everything. Then the bin wrapper can provide a default node file loader that handles adjusting paths to be relative to the input.

I don't think this will ever quite _replace_ rcu, as it doesn't quite have the same intentions. It looks like rcu is mostly isomorphic, but I don't see a whole lot of point to that for bundling with the ractive module. There's just so much more mileage you can get out of pre-building and bundling app and library code. I keep hearing that http2 makes having multiple requests actually preferable to bundling, but it seems like the added complexity of having a server smart enough to know which scripts it needs to push out with certain requests would be trounced by simply bundling the scripts up as a build step. I say simply because rollup is _relatively_ simple :smile:. I still haven't muddled through a webpack or browserify build.

I also don't think sourcemap support here is particularly useful. The only _actual_ source that ends up in the output is the script source, which is recommended to come from a single script tag. Everything else, being styles and template chunks, is neither executable nor sourcemap-able. And generating/interacting with the various module formats is way beyond the scope of what I think this should handle.

So, I think that leaves something like `require('ractive/lib/component')` handing back an object with a function that takes a few options (the primary string template, a file loader function, parser options) and returns a promise of a string. The the much simpler `require('ractive/lib/parse')` would hand back an object with a parse function, which takes a string template and hands back a string of compiled template, and a stringify function, which takes a compiled template and stringify-s it. Does that sound reasonable?

I think adjusting the modules like that will make this reasonably testable, which is something I was trying to figure out how best to approach.

ghost:
Oh snap :facepalm: Your second paragraph reminded me that rcu is more than just a node module/build time dependency. It's actually an embeddable script. I was looking at everything from a node build tool dev dependency perspective. They might have to stay separate after all. D:

bin/lib/component.js
```diff
@@ -0,0 +1,166 @@
...
+          This may also be a directory containing a component in index.html.
+      -o, --output {file} - file to write, otherwise STDOUT
+      -x, --nocsp - don't output expression functions
+      -d, --delimiters {open} {close} - set the plain mustache delimiters (default {{ }})
+      -s, --static {open} {close} - set the static mustache delimiters (default [[ ]])
+      -r, --triple {open} {close} - set the triple mustache delimiters (default {{{ }}})
+      -p, --static-triple {open} {close} - set the static triple delimiters (default [[[ ]]])
+`;
+
+function build ( Ractive, string, output, opts, base ) {
```
ghost: I'm curious why `Ractive` has to be handed over via args. I believe you can easily do the following since Node does seem to cache modules.

```
var Ractive = require('../../ractive');
```

evs-chris: I was just being lazy since ractive may exist at `../ractive.js` or `../build/ractive.js` depending on whether you're running from a git checkout or the actual built module.

ghost: I see. Oki doki. 😄

evs-chris:
I've adjusted the library functions so that they could conceivably be used by third parties:

* `require('ractive/lib/parse').parse('some template', { parse: options }).then(str => 'do something with the stringified parsed template');`
* `require('ractive/lib/component').build('some template', { parse: options }, function readFile(path) { return aPromiseForTheFileContent; }).then(str => 'do something with the component js');`

I also added the $PARTIALS replacement, which just subs the JSON.stringify-ed partials from script tags, and adjusted the stringified template output so that it doesn't quote stuff that doesn't really need to be quoted. That saves a fair amount of space for parsed templates.

PR #2791, opened 2016-12-01
Combine and remove duplicate (de)camel methods

benbeaumont:
Absolutely tiny change to remove a couple of lines of duplicated code (there are two camel/decamel methods in the utils).

Really - I'm just trying out the whole git/push/pull thing to make sure I understand how it works.

None - just a small clean up I noticed while passing through the code.

## Is breaking:
No.  All unit tests continue to pass


ghost:
Would be nice to rename those to indicate what case it came from and what case it transformed to. That way, it's not confusing as to what it accepts, and what it returns (had to run the code to see what the other case was).

And to be on the safe side, tests (not sure if any exist specifically for these functions, haven't checked).

Other than that, looks good to me. 👍

benbeaumont:
I changed the names of the methods to make it clear that the functions are actually swapping from hyphenated to camelCase (so are now called 'camelizeHyphenated' and 'hyphenateCamel')

Note that other tests do fail if these two methods are broken (so I don't think there is a need for a separate unit test).

evs-chris:
:+1: Nice!

evs-chris:
Aside, missed an opportunity for camel kebabs :laughing:

benbeaumont:
lol

Issue #1723, opened 2015-02-17
Enhancement suggestion: better parse errors


dagnelies:
Hi,

Here are a couple of examples.
- When an unknown tag is encountered, because misspelled or missing,  it'd be nice to print a warning. http://jsfiddle.net/7xo32g8n/
- When a closing tag is missing, it'd be nice to print a warning/info. Possibly enabled through an optional "strict" mode. http://jsfiddle.net/7xo32g8n/1/
- (optional) When referring to the error position, would it be possible to refer to the line in the file itself rather than within the script tag? http://jsfiddle.net/7xo32g8n/2/

evs-chris:
For the first, what about components (Ractive), components (HTML), and the semi-semantic HTML behavior allowing any tag name?

The second *would* require a strict mode, as Ractive tries to follow the HTML spec where possible, and HTML allows automatically closed tags.

I don't think the last one is possible because Ractive parses the content of the given selector. It has no notion of where the content is on the page, and even if it did figure out where the node is within the current textified DOM, it is probably not the same as the original file.

evs-chris:
Track in #2466

PR #2792, opened 2016-12-02
Better parse errors

ghost:

Adds a bunch of errors where needed. Laundry list includes:
- [x] Warning when encountering unknown, non-custom-element tag names.
- [ ] More to come.

- https://github.com/ractivejs/ractive/issues/1723

## Is breaking:

No?

## Reviewers:

Everyone.

### DIFF
--- a/src/utils/is.js
+++ b/src/utils/is.js
@@ -36,3 +36,7 @@ export function isObjectLike ( thing ) {
 	const type = typeof thing;
 	if ( type === 'object' || type === 'function' ) return true;
 }
+
+export function isNotAnElement ( name ){
+	return document.createElement( name ).constructor === HTMLUnknownElement;
+}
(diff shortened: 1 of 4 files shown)


src/utils/is.js
```diff
@@ -36,3 +36,7 @@ export function isObjectLike ( thing ) {
 	const type = typeof thing;
 	if ( type === 'object' || type === 'function' ) return true;
 }
+
+export function isNotAnElement ( name ){
+	return document.createElement( name ).constructor === HTMLUnknownElement;
```
ghost, line 41: This really looks too easy. Any ideas?

evs-chris: I think that may be a bit too expensive to do for _every_ element that's created. I think #1723 is looking for a parse-time warning, which is nigh impossible if you want to be able to use components. I think the only thing that can be reasonably implemented from that issue is a strict(er) mode that requires closing tags and maybe has a few more smarts about what elements can be nested where.

ghost:
Closing. Solution not viable. 😄

Issue #2597, opened 2016-06-28
Drop IE8 for 0.8? Anyone still have to support it?


evs-chris:
__Please vote__: use the thumb's up reaction on this comment to vote to drop IE8 support, and thumb's down to try to keep it.

I had a nice long rant about trying to debug IE8, but I'll spare you. 😄 You probably already know there's just no reasonable way to debug stuff on IE8. Also, @MartinKolarik bravely tried _actual_ IE8 in a VM and found that _emulated_ IE8 is actually more well-behaved. 😢

The first error in the test suite, in aliasing because the alphabet, is caused by `bind` being called on `Alias`, but the function that actually executes is `bind` on Transition. Obviously `Transition.bind` is going to fail when called on an `Alias`, but I have been unsuccessful in determining how that is happening at all. My only guess was that the `Transition.bind` was somehow becoming positionally global, but it has other `bind`s defined before and after. The prototypes also have references to the correct methods. 😖

All of my clients dropped IE8 early last year, and as of early this year, have also dropped IE9. I don't think maintaining IE9 support will be too cumbersome at this point; not so with IE8. Anyone else?

fskreuz:
Minimum IE at our place is IE9.

heavyk:
@fskreuz yep. IE8 is sooo 2006.

mpowell-atomic:
At the very large corporation that I do some contract work for (using Ractive, of course), they upgraded everyone to IE11 earlier this year.

JonDum:
Yea seriously don't waste any of those precious brain clock ticks of yours on IE8

kouts:
IE8 not needed here too

skeptic35:
Good riddance

aharpervc:
IE10 is an acceptable minimum. Don't need to be concerned with versions older than that.

tomekmarchi:
I would drop all of IE NP switch the focus to Edge.

dagnelies:
A few related tidbits:
https://www.microsoft.com/en-us/WindowsForBusiness/End-of-IE-support
https://support.microsoft.com/en-us/lifecycle?p1=15672
...so I think it's safe to drop it.

MrStonedOne:
Ugh, 

We use ractive in an embedded IE view for a game that still sees some windows xp users. Because microsoft won't release ie9 to windows xp, we are forced to support ie8

fskreuz:
@MrStonedOne Oh cool! What game is this?

[...]

PR #2801, opened 2016-12-15
Moar polyfill removal

ghost:

So I changed the goal of this PR to keep it simple: Remove all polyfills and use native APIs.

Reasoning:

- [It's already a year since all IE < 11 went EOL](https://www.microsoft.com/en-us/WindowsForBusiness/End-of-IE-support).
- Lesser code to deal with.
- Cleanup.
- Size.

A list of APIs affected:

### [ES5](https://kangax.github.io/compat-table/es5/)

* `Array.isArray`
* `Array.prototype.every`
* `Array.prototype.filter`
* `Array.prototype.forEach`
* `Array.prototype.indexOf`
* `Array.prototype.map`
* `Array.prototype.reduce`
* `Date.now`
* `Function.prototype.bind`
* `Object.create`
* `Object.defineProperty`
* `Object.defineProperties`
* `Object.freeze` (used in one of the test support files, not in library)
* `Object.keys`
* `String.prototype.trim`

### [ES6](https://kangax.github.io/compat-table/es6/)
* `Array.prototype.find`
* `Object.assign`
* `Promise`

### Web
* [`performance.now`](http://caniuse.com/#feat=high-resolution-time)
* [`requestAnimationFrame`](http://caniuse.com/#feat=requestanimationframe)
* [`addEventListener`](http://caniuse.com/#feat=addeventlistener)
* [`getComputedStyle`](http://caniuse.com/#feat=getcomputedstyle)
* `Node.prototype.contains` (only used in testing, not in library)

## Is breaking:

For modern browsers, this shouldn't be a problem. This _will_ break on older browsers and environments unless you provide a polyfill/shim for the APIs mentioned above.

Included in this PR is a file that contains polyfills and shims for the APIs mentioned above. Your mileage may vary. Not all APIs are shimmed 100% and I have this feeling IE is blowing up because of these.

## Reviewers:

Everyone. More eyes the better. 😄


evs-chris:
Neat!

Also wanted to note that

```js
export const create = Object.create;
export const defineProperty = Object.defineProperty;
export const defineProperties = Object.defineProperties;
```
and later
```js
import { create, defineProperty, defineProperties } from '../wherever/utils/object';
// use the methods
export function doSomething() {
  const obj = create(null);
  defineProperties(obj, ...);
}
```
actually results in smaller minified code, because it compiles down to (approximately)
```js
var a = Object.create;
var b = Object.defineProperty;
var c = Object.defineProperties;
function doSomething() {
  var d = a(null);
  c(d, ...);
}
```
as opposed to
```js
function doSomething() {
  var d = Object.create(null);
  Object.defineProperties(d, ...);
}
```

So the simple case, where the methods are only used once are a net loss, but when they're repeated more than about three times, it's a net gain. Uglify can't safely mangle the member accesses, but it can safely mangle local variable names. I'm pretty sure the global `Object` gets left in there too, whereas with the helper version, it only occurs once per method. In fact, to get even minifier friendly, you could
```js
const GlobalObject = Object;
export const create = GlobalObject.create;
...
```

I'm thinking about introducing an output step that converts all of the `ClassName.prototype.someMethod = function () {...}` to `var e = 'prototype'; ClassName[e].someMethod = function () { ... }`, which saves roughly 5k off the minified size and introduces almost no performance overhead because the prototype assignments don't happen constantly. See https://github.com/evs-chris/hoist-props

tomekmarchi:
More love <3

ghost:
@evs-chris Yep. Was just staring at the changes and for a while it appeared that the sum of `Object.` 's added outweighs the file it removed. 😄I guess you just confirmed it.

I'll incorporate this optimization once everything is replaced with their native counterparts. 👍

paulocoghi:
Excellent move. My only concern is the removal of `Performance.now`.

The use of `Date.now` alone seems to me as a backup function, because it can generate equal numbers in heavy use cases (and consequently conflicts).

_(to me Ractive is faster than a bullet and these conflicts are very easy to occur)_

`Date.now` had been chosen when `Performance.now` had little support. Today this is different: [caniuse.com/high-resolution-time](http://caniuse.com/high-resolution-time) , and my suggestion is to drop `Date.now` instead (and add performance.now in the polyfill).

aharpervc:
Does this change the minimum supported browser versions?

ghost:
@aharpervc The plan is to push the core code forward and offload as much stuff to native. As it stands, the edge build will need browsers with at least ES6 support because built-in promises was removed in an earlier PR.

But in this PR will be a file that will contain polyfills, only those Ractive will need. This file plus Ractive should, in theory, give the same effect as the legacy build. You can opt-out of using this file if you bring your own polyfills.

Then again, this is still a PR. It's open to everyone's scrutiny. 😄

ghost:
Changed a bunch of stuff in the PR. Now just removing polyfills and self-implemented functions. Tests work with IE11 (Win 10), but blows up on VM'ed IE (Win7). 💣 💥 

Would be nice to have someone test this on IE8-11 without using a VM. I also recall a discussion of Ractive being used in a game patcher (which uses embedded IE). Wondering if that can be tested too.

paulocoghi:
> I also recall a discussion of Ractive being used in a game patcher (which uses embedded IE). Wondering if that can be tested too.

The user is @MrStonedOne (and [here is his message](https://github.com/ractivejs/ractive/issues/2597#issuecomment-246392594))

paulocoghi:
Maybe I'm being too extreme, but if it works in Chrome, Firefox, Opera, Edge, Safari **and IE11**, I suggest to place the additional code (that would make Ractive compatible with the rest of IE versions) in a separated file, the same way as the polyfill (or within it).

Internet explorer has already ruined our lives too much. :face_with_head_bandage:

_(and versions 5.5 to 10 together represent <1% of [global market share](http://caniuse.com/usage-table))_

MrsTonedOne:
I should point out that embedded IE is still a somewhat common thing, and wine can only go up to ie8 in embedded ie because of the directx changes in ie 9, so you also lock out linux users of these windows apps.

(we have ractive 0.7 working in embedded ie on wine)

I'm struggling to understand this hatred for interoperability. ractive 0.7 works in windows, mac, wine, linux, chrome, ie, a toaster, some piece of silicon you found on the ground outside intel's headquarters etc.... like, it works in fucking wine running embedded ie on a laptop made in the 90s! but ractive 0.8 will kinda work, maybe, if the os was released in the last week, and its not a vm, and the stars align?

For what? So the codebase can *feel* tidy?

paulocoghi:
@MrStonedOne , I apologize for my jokes.

Every old browser (non-ES6 compatible) is being supported by a polyfill file (and not discarded). My suggestion is to also include the fixes related to IE 8-10 in it (and not in the core).

The community has always worked hard to support IE, as we previously discussed [here](https://github.com/ractivejs/ractive/issues/2597).

And I mentioned you to help us testing and continue to support it.

_I apologize once again for my jokes giving a false impression of hatred._

evs-chris:
> I'm struggling to understand this hatred for interoperability.
> For what? So the codebase can feel tidy?

I have no hatred of interoperability. I simply lack the skills and resources to make Ractive work in IE8. If someone can get 0.8+ to work in IE8, I would gladly readjust the codebase to accommodate that support. After #2597, which contains the primary reason _why_ I gave up on IE8, and subsequent queries to the community, we came to the conclusion that IE8 support was no longer tenable. So these PRs are more of a cleanup of dead code to make other, potentially better polyfills usable with Ractive. If anyone more experienced with debugging bizarre issues in IE8 would like to try to muddle through the issues, I would be glad to help out the best that I can from the Ractive side, and I'm sure any other maintainer would be happy to help too.

ghost:
The tests on this PR are currently polyfilled with es5-shim, es5-sham and es6-shim so tests should pass (in theory) on legacy IE. Also makes Phantom happy (yey!).  Tests passed in IE11, I just can't get it to work in any IE inside a VM. So if anyone can run the tests with a real legacy IE, that would be great. 👍 

I once added a separate polyfills file but wondered about maintenance. Do we really want to maintain our own copy of polyfills? es5-shim and es6-shim are both battle-tested, maintained by more hands, and can easily be popped in.

evs-chris:
I only have IE11 available, so I'm afraid I'm not much help there. Though, to be fair, I've only _ever_ been able to test IE < 10 using the emulated modes.

I _would_ like to package all of our old polyfills back up into a file in the distribution for "batteries included" purposes. For people who are only interested in getting Ractive to work, it's a whole lot easier to toss in one script before Ractive than figure out which shims and shams to include and in what order :smile:.

tomekmarchi:
Love this progress, it needs to be gutted.

ghost:
Chucked in a polyfills file containing the ones in `legacy.js` and came across 2 issues:

- `array.reduce` where `[1,2,3].reduce(()=>{})` returns `undefined` natively, but throws in the `legacy.js` version. This was spotted while rewriting the polyfills to ES6 to appease the linter and noticed a slight difference in the logic vs the polyfill in MDN (which closely followed spec).

    Luckily, this one was easily adjusted after going over the ES specs. The danger was that it was actually applied to the prototype.

 - `Object.defineProperty` just copies the getter natively but executes the getter on the `legacy.js` version. It also silently fails on features it cannot emulate. This was spotted by forcing Opera (essentially Chrome) to use the polyfills. Tests failed in Opera in roughly the same pattern as IE.

    The danger is that the polyfill acted different and was quiet about it. To compare, es5-shim throws an error when trying to use features it cannot emulate - something Ractive's unit tests can actually capture if used.

I'm cool with polyfills. They just require dedicated tests (to keep them in spec) and documentation (to tell others about their quirks) which require humans - a resource that's not abundant at the moment. Ractive will still need slight adjustments to work in legacy IE. It's just that moving the polyfill responsibility to dedicated tools will focus efforts on Ractive and not the bizarrely failing tests, nor the potentially behavior-altering polyfills, nor the horrible VM'ed IEs.

---

PS: I'm still not sure how `Object.defineProperty` polyfill worked on older browsers when it silently dropped everything aside from assigning a value to the target object... unless Ractive expected it to do just that. If so, it's no different from just using `Object.assign` or `for-in`.

evs-chris:
> .. unless Ractive expected it to do just that

Most of the time. I authored the terrible "just call the getter and hope for the best" hackery, but I can't remember exactly why. I know I was trying to get something to work in IE8 at that point, but it's been a while and I've probably intentionally blocked that out :wink:

Did you get anything to run in IE8? Cause if you did, you get like 1000 internet points for wizardry. As long as the polyfill works for IE9+, it's good with me. So for my part, you can actually drop the `defineProperty` polyfills.

ghost:
> Did you get anything to run in IE8? Cause if you did, you get like 1000 internet points for wizardry.

If you consider finishing the tests after a 5-10 minute wait with mixed results, yes? 😄 I'm checking with IE11 in the VM first to see why tests blow up in there and work my way back. It might just be tests that didn't clean up properly after failure 😁 .

ghost:
So I discovered a little something: Tests run together in QUnit, stuff blows up. Run those same failed tests individually... they pass 😱 I have a feeling this is caused by the combination of IE being very slow, tests failing because of that, and tests not being cleaned up properly afterwards.

Randomly picking out tests across the browsers, IE9-11 pass the majority of the tests. There's a few that just fail, like [`this.node.data = this.getString()`](https://github.com/ractivejs/ractive/blob/dev/src/view/items/Interpolator.js#L69) where the assignment (yes, the assignment, not `this.getString()`) throws `Invalid Argument`. IE8 fails nearly all tests with `Died on test #1 undefined: __proto__ computation must have a `get()` method`.

🤔

evs-chris:
Ractive has some very critical state that tends to get out of whack when an uncaught exception causes some internal functions not to complete. It's particularly bad with the runloop, which will get a batch stuck at the top of its stack and cause any future promises not to resolve.

evs-chris:
I just ran this through (emulated) IE, and I gotta say I'm quite surprised. There's a good bit of breakage in IE8, but it actually loads all of the tests and only ~500 fail. If you patch the computation init to skip `__proto__`
```diff
diff --git a/src/Ractive/construct.js b/src/Ractive/construct.js
index 39283352..5a205b24 100644
--- a/src/Ractive/construct.js
+++ b/src/Ractive/construct.js
@@ -55,6 +55,7 @@ export default function construct ( ractive, options ) {
        const computed = Object.assign( Object.create( ractive.constructor.prototype.computed ), options.computed );

        for ( const key in computed ) {
+               if ( key === '__proto__' ) continue;
                const signature = getComputationSignature( ractive, key, computed[ key ] );
                viewmodel.compute( key, signature );
        }
```
and somehow polyfill `setProperty` on the node style objects, I think it might kinda sorta be usable if you don't get crazy. I can't imagine _wanting_ to tiptoe around things that IE8 can't handle, but if you _had_ to I think you could.

I still wouldn't change the minimum officially supported version to anything less than IE9 though :smile:

evs-chris:
Just for kicks, it probably wouldn't hurt to throw `__proto__` into the non-enumerable properties too:
```diff
diff --git a/src/Ractive/construct.js b/src/Ractive/construct.js
index 39283352..5a205b24 100644
--- a/src/Ractive/construct.js
+++ b/src/Ractive/construct.js
@@ -55,6 +55,7 @@ export default function construct ( ractive, options ) {
        const computed = Object.assign( Object.create( ractive.constructor.prototype.computed ), options.computed );

        for ( const key in computed ) {
+               if ( key === '__proto__' ) continue;
                const signature = getComputationSignature( ractive, key, computed[ key ] );
                viewmodel.compute( key, signature );
        }
diff --git a/src/polyfills.js b/src/polyfills.js
index 1c921cf5..e87a77e9 100644
--- a/src/polyfills.js
+++ b/src/polyfills.js
@@ -311,7 +311,7 @@
        // Object.keys
        if (!Object.keys) {
                const hasDontEnumBug = !({ toString: null }).propertyIsEnumerable('toString');
-               const dontEnumProps = ['toString', 'toLocaleString', 'valueOf', 'hasOwnProperty', 'isPrototypeOf', 'propertyIsEnumerable', 'constructor'];
+               const dontEnumProps = ['__proto__', 'toString', 'toLocaleString', 'valueOf', 'hasOwnProperty', 'isPrototypeOf', 'propertyIsEnumerable', 'constructor'];
                const dontEnumPropsLength = dontEnumProps.length;

                Object.keys = function (obj) {
```

evs-chris:
Is this good to go?

ghost:
I cannot verify IE9 and 10, even without the PR. But IE11 and Edge do work (as well as other modern browsers).

evs-chris:
I can't verify _real_ IE9 and 10, but I'll give it a once through with the emulated versions, which, to be fair, is how the last two versions have been checked :smile:.

evs-chris:
Well, IE10 seems fine, which is what I would expect. IE9 fails 5 tests, one related to finding elements in a triple and four related to weirdness with selects. I'm going to call that close enough to merge this so we can move toward an alpha. Any objections?

paulocoghi:
No objection. For me you can go on.

Issue #1376, opened 2014-10-18
Lightweight component experiment


martypdx:
I did some initial research into the idea I present in #1362 for components that reuse their parent's viewmodel and other potential optimizations. I started with just doing the viewmodel optimization (and inheriting fragment context).

For 1000 items with a few data fields:

``` html
    <script id='inline' type='text/ractive'>
        {{#items}} {{name}} {{number}} {{date}} {{/}}
    </script>

    <script id='partial' type='text/ractive'>
        {{#items}} {{>item}} {{/}}
        <!-- {{>item}} -->
        {{name}} {{number}} {{date}}
        <!-- {{/item}} -->
    </script>

    <script id='componentbound' type='text/ractive'>
        {{#items}} <component item='{{.}}'/> {{/}}
    </script>

    <script id='componentwith' type='text/ractive'>
        {{#with item}} {{name}} {{number}} {{date}} {{/with}}
    </script>

    <script id='componentimplicit' type='text/ractive'>
        {{#items}} <component/> {{/}}
    </script>

    <script id='component' type='text/ractive'>
        {{name}} {{number}} {{date}}
    </script>
```

Here are the results:

| method | low(ms) | high(ms) |
| --- | --- | --- |
| inline | 120 | 150 |
| partial | 150 | 180 |
| bound | 520 | 565 |
| implicit | 525 | 560 |
| lightweight | 370 | 388 |

So ~33% less and over half-way to same time as partial.

Really nice that you can use the context block: `{{#items}} <component/> {{/}}` without having to scope the template: `{{name}} {{number}} {{date}}`

PR #2811, opened 2016-12-30
Better perf for many (empty) components

evs-chris:
This is the result of me trying to figure out whether or not it makes sense to pursue #1376 and #1379 lightweight components any further. I've come to the conclusion that it's not really needed because the introduction of proper models in 0.8 basically killed the model-related overhead involved with components. They're still not super-lightweight because they're roughly 20x more expensive than a plain partial, but to be fair, partial performance is __way__ better than it was before 0.8. The only bits I could think to strip off of components are somewhat necessary to have an encapsulated widget anyways (its own potentially isolated model, registries, and lifecycle events).

When I started profiling with an iterated list of 1000 components with just the letter 'c' as their template and partials with just the letter 'p' as their content, the initial render for the components was ~240ms settling to ~140ms, and the initial render for the partials was ~27ms settling to ~4ms. That's a pretty wide gap. With these changes, the initial render for components goes to ~150ms settling to ~50ms, and the partial times stay about the same.

The biggest change comes from a slight refactor on the code that checks to see if elements can be detached at the end of a runloop, namely, if there are no nodes to detach, there's no need to try to gather outros to check against an empty list.

The second comes in the form of a slight event firing refactor wherein instances that have no namespaced events are entirely skipped during bubbling, so you don't pay a perf penalty unless you're actually using namespaced events. This only accounts for 5-7% of the gap.

The last bit comes from fixing some de-opts around `combine` and `extend` and re-jiggering a few hot codepaths to avoid extra function calls and whatnot.

My final test with 50000 iterations of each shows that for huge numbers of iterations, components are ~10x more expensive than partials on initial run. edge actually bails when trying with 50000, but 20000 took 23523ms where this took 1188ms. 50000 components on this takes ~2800ms. 20000 partials takes ~200ms on edge and ~160ms on this for the initial run.

Now, I'd imagine that components that _actually do something_ other than render a single character are probably even _more_ expensive than partials, but I was trying to get a feel for just how much overhead there is at the minimum implementation. I feel pretty good about this PR though, given the difference from edge when rendering 20000 components.

To some extent, any pertaining to performance with lots of components.

## Is breaking:
Nope.

### DIFF
--- a/src/Ractive/construct.js
+++ b/src/Ractive/construct.js
@@ -37,9 +37,11 @@ export default function construct ( ractive, options ) {
 	constructHook.fire( ractive, options );
 
 	// Add registries
-	registryNames.forEach( name => {
+	let i = registryNames.length;
+	while ( i-- ) {
+		const name = registryNames[ i ];
 		ractive[ name ] = extend( create( ractive.constructor[ name ] || null ), options[ name ] );
-	});
+	}
 
 	// Create a viewmodel
 	const viewmodel = new RootModel({
@@ -112,6 +114,7 @@ function initialiseProperties ( ractive ) {
 
 	// events
 	ractive._subs = create( null );
+	ractive._nsSubs = 0;
 
 	// storage for item configuration from instantiation to reset,
 	// like dynamic functions or original values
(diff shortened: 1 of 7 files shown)


evs-chris:
It looks like the lightweight components proposal was made somewhere around 0.6.0. Roughly duplicating  the test described in #1376, it looks like this PR using components is ~1.7x inline and partials in 0.6.0. 0.7.3 had a decent performance improvement all around, except when removing a bunch of components, which was particularly bad for some reason (like 3x render time to unrender). The 0.8 series improved general rendering further over 0.7.3, but it only improved component rendering by about 10%. Given that, I'm going to go ahead and merge this and say that we're now to the point that at 3x inline time, plain old components are probably fast enough not to need a special slightly hobbled version of them, at least for performance purposes.

PR #2813, opened 2016-12-30
Auto Dependency Updates ...

lguzzon:
Auto Dependency Updates ...


evs-chris:
Unfortunately, we still can't bump the gobble deps because newer versions use a file watcher that doesn't get along with Travis. Thanks for the prod though!

PR #2826, opened 2017-01-14
Lowercase element names if they are in the html namespace - #2821

### LINKED ISSUE #2821: Rendered unresolved component name case different on Edge.
## Description:

This test passes on every other browser I have (IE11, Opera and therefore Chrome, Phantom) but blows up in edge. It appears that there's confusion about the case of a component name when it's unresolved. Edge renders an unresolved `<Foo/>` as `<Foo></Foo>` while other browsers do `<foo></foo>`.

Not sure which one is correct though. The component's name, as used on the template, is `<Foo/>` so it might make sense that the HTML be `<Foo></Foo>`. However, I am not an expert when it comes to HTML spec and behavior. It might be correct to lowercase element names, as HTML is a case-insensitive language.

## Versions affected:

The test has been there since March 2016. Potentially 0.8 and newer only.

## Platforms affected:

Edge

## Reproduction:

untitled

## Code:

```js
	test( 'Components not found in view hierarchy when isolated is true', t => {
		const Foo = Ractive.extend({
			template: 'foo'
		});

		const Bar = Ractive.extend({
			template: '<Foo/>',
			isolated: true
		});

		new Ractive({
			el: fixture,
			template: '<Bar/>',
			components: { Foo, Bar }
		});

		t.equal( fixture.innerHTML, '<foo></foo>' );
	});
```

evs-chris:
This lowercases element names when creating elements if they are in the html namespace. This should resolve #2821 and not affect other namespaces, like svg, that require case sensitive tag names.

There's not really a way to CI test this, other than the failing test on Edge mentioned in the linked issue, because webkit/blink-based browsers (PhantomJS) return lowercased elements regardless.

#2821

## Is breaking:
Nope.

## Reviewers:
@fskreuz

### DIFF
--- a/src/view/items/Element.js
+++ b/src/view/items/Element.js
@@ -191,7 +191,8 @@ export default class Element extends ContainerItem {
 		}
 
 		if ( !node ) {
-			node = createElement( this.template.e, this.namespace, this.getAttribute( 'is' ) );
+			const name = this.template.e;
+			node = createElement( this.namespace === html ? name.toLowerCase() : name, this.namespace, this.getAttribute( 'is' ) );
 			this.node = node;
 		}
 


ghost:
untitled

Tests pass. Looks good. 👍

Issue #1800, opened 2015-03-12
ractive & custom tags


timoxley:
I like to use custom tags in my ractive templates so it's easy to see which elements belong to which ractive component.

Something like:

``` js
import Ractive from 'ractive'

document.registerElement('my-element')

var MyElement = Ractive.extend({
  isolated: true,
  template: `
  <my-element title="{{title}}">
    <h1>{{title}}</h1>
  </my-element>
  `
})

new Ractive({
  el: document.body,
  template: `<my-element title="{{title}}"></my-element>`,
  components: {
    'my-element': MyElement
  },
  data: {
     title: 'hello'
  }
})
```

To produce HTML like:

``` html
<body>
  <my-element title="hello">
    <h1>hello</h1>
  </my-element>
</body>
```

To prevent ractive from creating infinitely recursive components, you need to use `isolated: true` or null out the component reference within itself:

``` js
var MyElement = Ractive.extend({
  components: {
    "my-element": () => false
  },
  template: `
  <my-element title="{{title}}">
    <h1>{{title}}</h1>
  </my-element>
  `
})
```

This nulling out the component idea is a minor pain in debug mode since Ractive whinges and whines about there not being a component reference (gets verbose if you have a lot of elements):

```
Ractive.js: A function was specified for "my-element" component, but no component was returned
```

Perhaps a flag to disable this warning or to simply disable recursive components?

Another issue is that you need to remember to forward class and id type information onto to the custom tag.

It would be good if Ractive had a better story for dealing with custom elements. Obvious solution could be a flag which would just leave the Ractive component in-place as a custom element.

martypdx:
> Perhaps a flag to disable this warning or to simply disable recursive components?

I have apps that use recursive components, so that wouldn't work across the board. What if you could just do:

```js
components: {
    "my-element": null
 },
```

> Another issue is that you need to remember to forward class and id type information onto to the custom tag.

Yeah this comes up from time to time, see #1535 for example. (You could create a decorator or `onrender` handler that did this). The issue is that IMO components as ractive instances shouldn't dictate the structure of the contents (for example even having a single top-level element). OTOH there's this widget/custom element view of the world that behaves as your suggesting. My hunch is that that needs to be its own class of component, either by option or maybe even a separate `Ractive.Element.extend` or something.

Rich-Harris:
The 

```js
components: {
    "my-element": null
 },
```

[...]

PR #2835, opened 2017-01-19
Add support for naming attributes in component definitions - RFC (0.9)

evs-chris:
This adds a new option to `Ractive.extend` that allows component authors to specify which attributes should be accepted in the component. If the component user fails to supply a `required` attributes, then a warning will be issued in `DEBUG` mode. The `attributes` option may be an array of strings, which translates into an array of `optional` attributes, or an object with `optional` and/or `required` keys with array of strings values.

Any attributes that are passed that are not found in either the `required` or `optional` attributes lists will _not_ be added to the component model. This keeps users from accidentally shadowing an internal property with a link/mapping. Further, these additional attributes are collected into a partial, named `extra-attributes`, which can then be yielded in a tag body to make passing html attributes to the main component element simpler. Note that unless there are no mapping-style attributes, the provided `extra-attributes` partial _must_ be yielded because the additional mappings will _not_ be created. Using `yield` keeps the references in the context of the calling instance so that they can resolve properly.

This option _only_ affects components that are instantiated by a template, which leaves out `new MyComponent()` and anchors.

`attributes` are not inherited from their parent component.

This is one of the last things on my list to address before moving 0.9 to alpha. I think there's some merit to pulling the bulk of #2766 into a helper library for more advanced cases, but it's just too heavy and independent to include here. It would also complement this PR for more advanced cases like splitting extra attributes across more than one element in the component body.

### Questions

1. Should there be an additional setting in the `attributes` object form that allows extra attributes to be mapped so that the `extra-attributes` partial ends up using references to the mappings? e.g. `<cmp foo="{{bar}}" />` with foo not specified as an attribute would still create a `foo` mapping to the owner's `bar`, and the extra partial would effectively be `foo="{{foo}}"` i.e. referencing the mapped value.
2. Are `attributes`, `required`, `optional`, and `extra-attributes` good names?
3. I also considered allowing the `required` array to be an object where the key is the option name and the value is a message to display if the attribute is elided. That seems like overkill for now though, right?

#1535 and half of #1800

## Is breaking:
Only if you were already using `attributes` with `extend` to add it to the component prototype.

## Reviewers:
@dagnelies @fskreuz and anyone else interested in better control over component attributes

### DIFF
--- a/test/browser-tests/components/attributes.js
+++ b/test/browser-tests/components/attributes.js
@@ -1,0 +1,108 @@
+import { test } from 'qunit';
+import { onWarn } from '../test-config';
+import { initModule } from '../test-config';
+
+export default function() {
+	initModule( 'components/attributes.js' );
+
+	test( `by default all attributes are mapped`, t => {
+		const cmp = Ractive.extend({
+			template: '{{foo}} {{bar}}'
+		});
+
+		t.equal( cmp.attributes, undefined );
+
+		new Ractive({
+			target: fixture,
+			template: '<cmp foo="{{baz}}" bar="{{bat}}" />',
+			components: { cmp },
+			data: {
+				baz: 1,
+				bat: 2
+			}
+		});
+
+		t.htmlEqual( fixture.innerHTML, '1 2' );
+	});
+
+	test( `attributes as an array of strings are all optional`, t => {
+		const cmp = Ractive.extend({
+			attributes: [ 'foo', 'bar' ]
+		});
+
+		t.deepEqual( cmp.attributes, { optional: [ 'foo', 'bar' ], required: [] } );
+	});
+
+	test( `attributes may contain optional, required, neither, or both`, t => {
+		const optional = Ractive.extend({
+			attributes: { optional: [ 'foo' ] }
+		});
+		const required = Ractive.extend({
+			attributes: { required: [ 'bar' ] }
+		});
+		const both = Ractive.extend({
+			attributes: { optional: [ 'foo' ], required: [ 'bar' ] }
+		});
+		const neither = Ractive.extend({
+			attributes: {}
+		});
+
+		t.deepEqual( optional.attributes, { optional: [ 'foo' ], required: [] } );
+		t.deepEqual( required.attributes, { optional: [], required: [ 'bar' ] } );
+		t.deepEqual( both.attributes, { optional: [ 'foo' ], required: [ 'bar' ] } );
+		t.deepEqual( neither.attributes, { optional: [], required: [] } );
+	});
+
+	test( `only named attributes are mapped into component`, t => {
+		const cmp = Ractive.extend({
+			attributes: [ 'foo', 'bar' ],
+			template: '{{foo}} {{bar}} {{baz}}',
+			isolated: true
+		});
+		new Ractive({
+			target: fixture,
+			template: `<cmp foo="{{foo}}" bar="{{bar}}" baz="{{baz}}" />`,
+			components: { cmp },
+			data: {
+				foo: 'foo', bar: 'bar', baz: 'baz'
+			}
+		});
+
+		t.htmlEqual( fixture.innerHTML, 'foo bar' );
+	});
+
+	test( `leaving off a required attribute issues a warning`, t => {
+		t.expect( 2 );
+		onWarn( msg => t.ok( /.*cmp.*requires attribute.*foo.*/.test( msg ) ) );
+		const cmp = Ractive.extend({
+			attributes: {
+				required: [ 'foo' ]
+			},
+			template: 'yep'
+		});
+		new Ractive({
+			target: fixture,
+			template: `<cmp />`,
+			components: { cmp }
+		});
+
+		t.htmlEqual( fixture.innerHTML, 'yep' );
+	});
+
+	test( `extra attributes are collected into a partial and attached to the instance`, t => {
+		const cmp = Ractive.extend({
+			attributes: [ 'item', 'color' ],
+			template: `<div {{yield extra-attributes}} style-color="{{color}}">{{item}}</div>`,
+			data: { bar: 21 }
+		});
+
+		new Ractive({
+			target: fixture,
+			template: `<cmp class="big" data-foo="{{bar}}" item="{{item}}" color="green" />`,
+			components: { cmp },
+			data: { bar: 'yep', item: 'thing' }
+		});
+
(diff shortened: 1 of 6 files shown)


test/browser-tests/components/attributes.js
```diff
@@ -0,0 +1,108 @@
...
+		new Ractive({
+			target: fixture,
+			template: `<cmp class="big" data-foo="{{bar}}" item="{{item}}" color="green" />`,
+			components: { cmp },
+			data: { bar: 'yep', item: 'thing' }
+		});
+
+		t.htmlEqual( fixture.innerHTML, '<div class="big" data-foo="yep" style="color: green;">thing</div>' );
+	});
+}
```
ghost, line 152: Nice 👍  This will be great for wrapper components. No more boilerplate-y mappings.

test/browser-tests/components/attributes.js
```diff
@@ -0,0 +1,108 @@
...
+
+	test( `leaving off a required attribute issues a warning`, t => {
+		t.expect( 2 );
+		onWarn( msg => t.ok( /.*cmp.*requires attribute.*foo.*/.test( msg ) ) );
+		const cmp = Ractive.extend({
+			attributes: {
+				required: [ 'foo' ]
+			},
+			template: 'yep'
+		});
```
ghost, line 82: Might need a test for when the component tries to render `{{foo}}`, `foo` is required, but is never provided by the enclosing component.

test/browser-tests/components/attributes.js
```diff
@@ -0,0 +1,108 @@
...
+			components: { cmp }
+		});
+
+		t.htmlEqual( fixture.innerHTML, 'yep' );
+	});
+
+	test( `extra attributes are collected into a partial and attached to the instance`, t => {
+		const cmp = Ractive.extend({
+			attributes: [ 'item', 'color' ],
+			template: `<div {{yield extra-attributes}} style-color="{{color}}">{{item}}</div>`,
```
ghost, line 95: Not sure how Ractive deals with duplicate attributes (i.e. `<div data-foo="1" data-foo="2">`) but I'm curious how this will map that scenario.

test/browser-tests/components/attributes.js
```diff
@@ -0,0 +1,108 @@
...
+			data: { bar: 21 }
+		});
+
+		new Ractive({
+			target: fixture,
+			template: `<cmp class="big" data-foo="{{bar}}" item="{{item}}" color="green" />`,
```
ghost, line 101: Also wondering how it would deal with conditional attributes, given that the recent questions have been about them.

ghost:
1. If it doesn't add overhead and makes sense (does not involve a lot of explaining 😄 ), it's fine.
2. So essentially, both `required` and `optional` do the same thing, except ones listed under `required` will fire a warning?
3. I'd say overkill. :D

paulocoghi:
1. Same as @fskreuz . If the performance stays the same, it's welcome!
3. Overkill for now.

test/browser-tests/components/attributes.js
evs-chris: It used to be last one in wins, but as of 0.8 attributes are in an array rather than a map so 🤷‍♂️. The same thing would happen as `<div data-foo="1" data-foo="2">` if those were moved to the `extra-attributes` because it just moves them from the main template to the partial.

test/browser-tests/components/attributes.js
evs-chris: It could still be defaulted by the component, but if it's not, it will just be undefined as usual. I was looking at `required` as a guideline sort of thing. Erroring would cause major breakage.

test/browser-tests/components/attributes.js
evs-chris: Conditional attributes aren't considered because they're not members of the components attribute collection. The structure is basically `[ attr, attr, event, attr, section { attr, attr } ]`. This ignores anything in nested scopes like the `section`, which can get pretty deep and hairy. I suppose we could clone the section and try to split the attrs between the clones, but there are some structures that are impenetrable, like partials `[ attr, partial, attr ]` as in `<div foo="1" {{>extra-attributes}} bar="2">`. Any additional attributes in that partial are unknown until render, and there are similar issues with conditional sections that may not be available at initial render, which would cause problems if the attributes in them are `required`.

TL;DR this ignores conditional attributes

evs-chris:
I've added a mode for still mapping all attributes that is enabled with `attributes: { mapAll: true, ... }`. In that case, you'll want to use a regular partial for the extra attributes because they will now refer to the mapped local names e.g. `<div {{>extra-attributes}}>`. Here are the equivalents with `item` being the only attribute named and the template as:

```html
<cmp item="{{.}}" class="{{some.str}}" />
```

### mapAll: false
the extra attributes partial will effectively be `class="{{some.str}}"`, which must be yielded for the reference to act as expected. No additional mappings other than `item` are made.

### mapAll: true
the extra attributes partial will effectively be `class="{{class}}"` because `class` will be mapped to `some.str` on the parent instance.

I see the original behavior as being the more common expectation, so I left it as the default (not mapping extras).

evs-chris:
Oh, and yes, the only difference between `optional` and `required` is that `required` attributes that are not present at construction will cause a warning to be issued.

dagnelies:
Hi,

for my part, I'm very happy about this enhancement. I think it's a very useful addition for UI wrapper components. Thanks for the great job! The details also look well designed to me.

---

> Should there be an additional setting in the attributes object form that allows extra attributes to be mapped so that the extra-attributes partial ends up using references to the mappings? e.g. <cmp foo="{{bar}}" /> with foo not specified as an attribute would still create a foo mapping to the owner's bar, and the extra partial would effectively be foo="{{foo}}" i.e. referencing the mapped value.

...I thought the extra-attributes were yielded ...that is independent from the component's data. Right? Then what is this mapping about? ...sorry, I think I'm confused about it. 😕  Might you show an example illustrating the difference? ...btw, I'm all about yielding the attributes, I consider it to be ideal.

> Are attributes, required, optional, and extra-attributes good names?

It's fine with me.

> I also considered allowing the required array to be an object where the key is the option name and the value is a message to display if the attribute is elided. That seems like overkill for now though, right?

Indeed looks like overkill to me. A standardized message is enough.

---

I'm looking forward to test this out. As usual, I've lots of other things to do, but I'm very sure it'll be helpful for component libraries. 👍

evs-chris:
Adding an option to map the extra attributes allows their use outside of yielding the extra attributes partial, in case the component needs to `get` or `set` them or use them on an element without the rest of the bundled attributes. Basically it's a way to achieve the current behavior of component attributes with a working extra-attributes partial.

Issue #846, opened 2014-06-21
How can I remove event handler from a specific element?


QuyHP:
For example, I have registered a proxy event on an element

> &lt;a href="javascript:;" on-mouseover="doSomething"&gt;Do something&lt;/a&gt;

Say I have a reference to this element (from event.node inside the event handler), how can I remove/unsubscribe "doSomething" from just this specific element? (I still want to keep this event for other elements)

evs-chris:
Delegation is now in edge, you can wrap events in conditionals, and you could have a sing-fire event decorator, so I'm going to call this one done.

PR #2841, opened 2017-01-23
First swing at event delegation - RFC

evs-chris:
Ractive events currently install a listener on their target node, which works out great most of the time, but if you have a large iterative section, attaching event listeners adds a fairly significant overhead. You can manage delegation manually, but it's a little bit painful, because getting in the correct context requires some strange boilerplate.

To address that, this adds support for setting a delegation target and specifying events to be delegated further downstream. To set a target, you add `delegate-click` as an attribute to the element. Then to specify a delegated event, you set `delegate-click="doSomething"` on the source element. When this sees a delegate target, it installs an (or all) appropriate event listener on the target element. When it sees a delegated event, it sets up the fire handler but does nothing further. When the target catches an event, it starts with the target element and walks back up the hierarchy until it reaches the target node while looking for and firing matching delegated event listeners.

Caveats: this only works with events that fire with an original event, so some custom event plugins may not work. It also doesn't really make sense to support component events as delegated, because they don't involve the DOM.

Example:
```html
<table delegate-click>
{{#each lotsOfRows}}
  <tr>
    <td><button delegate-click="@.select(.)">Select</button></td>
    <td>{{.description}}</td>
    <td>{{.something}}</td>
    <td><button delegate-click="event.splice('../../', @index, 1)">Remove</button></td>
  <tr>
{{/each}}
</table>
```

Note that the delegated handlers are written exactly the same as non-delegated handlers would be.

In a benchmark very similar to that example, using delegation reduces render time by ~25% on my machine with 1000 rows, going from 40ms to 30ms on average. Unrender is also a bit quicker on average, as the listeners don't also need to be removed.

### TODO
* [x] Tests

The remaining bits of #846

## Is breaking:
Nope

## Reviewers:
Yep

### DIFF
--- a/src/config/types.js
+++ b/src/config/types.js
@@ -51,3 +51,4 @@ export const EVENT             = 70;
 export const DECORATOR         = 71;
 export const TRANSITION        = 72;
 export const BINDING_FLAG      = 73;
+export const EVENT_DELEGATE    = 74;
--- a/src/parse/converters/element/readAttribute.js
+++ b/src/parse/converters/element/readAttribute.js
@@ -1,4 +1,4 @@
-import { ATTRIBUTE, DECORATOR, BINDING_FLAG, TRANSITION, EVENT } from '../../../config/types';
+import { ATTRIBUTE, DECORATOR, BINDING_FLAG, TRANSITION, EVENT, EVENT_DELEGATE } from '../../../config/types';
 import getLowestIndex from '../utils/getLowestIndex';
 import readMustache from '../readMustache';
 import { decodeCharacterReferences } from '../../../utils/html';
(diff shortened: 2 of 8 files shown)


ghost:
🤔 

Wondering if delegation can simply be an implementation optimization rather than an additional attribute. Like if Ractive detects an `on-*` in a suboptimal spot (like in a repeating section), it attaches the handler to the most optimal ancestor. This way, there's no change in the template syntax/addition of another attribute.

The idea comes from jQuery's `.on()` method. Aside from the change in signature, the handler stays absolutely the same and is unaware of the fact that the attachment was done direct or via delegation. The only noticeable difference is that `event.delegateTarget` now has a value.

For instance, this would attach the click handler on `<table>` instead (escaping both sections). However, `select` would still be run as if it were in the context of the button.

```js
<table>
{{#each lotsOfRows}}
  <tr>
    <td>
    {{#each moarButtons }}<button on-click="@this.select(.)">Select</button>{{/each}}
    </td>
  <tr>
{{/each}}
</table>
```

My 2c.

MartinKolarik:
I agree with @fskreuz that we probably don't need new syntax for this.

evs-chris:
That's how I wanted to approach it to begin with, but I couldn't come up with a method for checking to see if the event should be delegated that wasn't potentially as expensive as just installing the handler. The other side of that is keeping track of the delegate listener when an event directive decides it needs to be delegated. I suppose it could be reference counted or just left in place until the element is removed.

There are also some corner cases with delegation and decorators or anything else that could potentially add event listeners between the delegate and target node. Since the event has already bubbled through to the delegate node, trying to stop the event from one of the delegates is not possible. That case _is_ covered for ractive-managed events, though.

So, we could probably add a flag to elements to see if they are in an iterative section, and if they are use delegated events. Could the cancellation concern be addressed by using a capturing listener for delegates?

evs-chris:
Hmmm. Even using capture would cause some weirdness with listeners being fired out of expected order.

Thinking about it, I would guess that 99% of the time delegation would be fine, and I suppose we could add an instance flag to disable it for those other 1% cases. If that's the accurate, I think I can come up with a way to do automatic delegation. Would that be be fairly universally preferred over manual control as implemented here?

MartinKolarik:
Yes, capture wouldn't really solve it, but I don't think it's a problem. Especially if there's a way to disable delegation, as you suggested. For me, definitely better than manual control.

evs-chris:
Superceded.

Issue #2798, opened 2016-12-11
Removing a few magical features


ghost:
## Description:

- **Magic mode** - Ractive detecting direct object mutation.
    - Ractive must know about the properties in advance.
    - Can't set new properties directly.
    - Individual setting of properties causes separate renders.
    - Nothing a `ractive.set()` can't do.

- **Array Modification** - Ractive detecting direct array mutation.
    - `__proto__`-modification code. On IE, it's array-modification code.
    - Doesn't really support all array methods, ie `array.fill()`, `array.concat()`
    - There are equivalent instance methods. Much more explicit in saying what is supported.

- **Deprecation warning system** - Ractive throwing friendly warnings on deprecations.
    - If a developer is keen on upgrading, they read change logs, not update and pray.

Ideas are welcome. 😄

evs-chris:
Two more for consideration, though I'm not sure how magical they are:

* __Ambiguous resolution__, which causes all manner of confusion among users, adds quite a bit of complexity to the (Ractive) code, and is the source of some fairly large performance penalties in some situations.
* __Implicit mappings__, which also cause confusion for component users, add a bunch of extra complexity to the (Ractive) code, and add significant performance overhead, especially when combined with ambiguous references.

For the magic and array adaptors, I'd like to distill an API good enough to handle them, a good firebase adaptor, and a proxy-based magic mode for 0.10.

ghost:
Does this mean components become isolated (which also gets rid of `isolated`) by default and that step 7-9 in [reference resolution](http://docs.ractivejs.org/latest/references) won't happen anymore? If so, I'm in 👍 .

evs-chris:
Well, it doesn't quite get rid of `isolated`, which also controls whether or not components inherit _other_ stuff, like adaptors.

It looks like you added `ractive.data` in an edit, but I think it's already removed in edge? I thought that was one of the deprecations I cleaned out.

I was curious, so I created a branch that strips out the array and magic adaptors, removes support for resolving ambiguous references, removes support for implicit mappings, and removes support for keypath expressions (bet you don't even know what those are :smile:) over here: https://github.com/ractivejs/ractive/tree/less-magic. There were a few tests that had to be updated (~50) because they were relying on ambiguous resolution. The rest were specifically testing one of the removed features.

[...]

PR #2843, opened 2017-01-23
Remove support for unresolved references - RFC

evs-chris:
This removes the unresolved references checks that work like this:
1. If a matching reference cannot be found in available contexts or keypaths, register a callback with each potentially matching path
2. Each time data is set, the model checks its list of unresolveds to see if there's a listener that is interested in the data
3. When the callback is fired, it should also remove itself from other registered paths
4. The reference is now resolved.

This is one of those features that is really nice in theory, but in practice can cause weirdness, confusion, and bad performance, which is why it's also against recommended practices. _Only_ ambiguous references can end up in the unresolved queue.

This replaces the above process with this one:
1. If a matching reference cannot be found in available contexts or keypaths, resolve the reference as if it were relative against the current context.

From what I've seen, a large percentage of refs that end up in the unresolved queue end up being resolved in their current context anyways.

**Note:** this is not removing support for ambiguous references. Those will still resolve to the closest matching context as they would before. It just stops ambiguous references from going on unresolved until some data is added somewhere along their context hierarchy that matches their base key (`foo` in `foo.bar.baz`).

I only had to remove or modify 11 tests to get this passing, and a couple of those were tests for warnings about ambiguous references in bindings.

Part of #2798 

## Is breaking:
Yep

## Reviewers:
Yep


aharpervc:
This still works, right?

```js

const MyComp = Ractive.extend({
// ...
});

const my = new MyComp({
  el: '#target',
  template: '{{ wut }}',
  data: {
    idk: "??",
  },
});

my.set("wut", "all good");
// and "all good" is rendered into the dom
```

evs-chris:
Yes, because when `wut` isn't found, it ends up resolving to the `wut` key on the context, which happens to be the root context. If the template was `{{#with something}}{{wut}}{{/with}}` with `{ something: {} }` as data, then `my.set('wut', 'all good')` would not have any effect because the context for the interpolator is `something` rather than root. Behavior in 0.8 and edge right now would still render "all good" because of the unresolved process. The `something` in data is required so that the `with` will render.

aharpervc:
Ok, thanks for clarifying

evs-chris:
Any objections to this? _I_ don't think that actual unresolveds happen too often in the wild because you're usually interested in the current context when it eventually pops into existence. I'd like to have this out for the alpha cycle (beta? rc? not sure what best to call it) for 0.9 so that I can get some more general community feedback.

ghost:
So...

- The "climbing" behavior will still stay.
- Immediately resolve (to the current context) if reference is not found anywhere.

I'm not exactly clear what resolution actually involves. Does it lock in the keypath to the current context? Does that mean "climbing" only happens during the first time a reference tries to resolve?

So in the case of this removed test:

	test( 'Data is synced as soon as an unresolved mapping is resolved', t => {
		onWarn( () => {} ); // suppress

		const ractive = new Ractive({
			el: fixture,
			template: '<Outer/>',
			data: {
				item: { x: 1 }
			},
			components: {
				Outer: Ractive.extend({
					template: '{{#with item}}<Inner foo="{{foo}}"/>{{/with}}'
				}),
				Inner: Ractive.extend({
					template: '<p>foo: {{foo}}</p>'
				})
			}
		});

		t.htmlEqual( fixture.innerHTML, '<p>foo: </p>' );

		ractive.toggle( 'item.foo' );
		t.htmlEqual( fixture.innerHTML, '<p>foo: true</p>' );

		ractive.toggle( 'item.foo' );
		t.htmlEqual( fixture.innerHTML, '<p>foo: false</p>' );
	});

The only way to render `foo` is to set in on `<Outer/>` (which then maps it to `<Inner/>`) or set it on `<Inner/> directly?

evs-chris:
Any time a reference resolves, it locks in the resolved absolute keypath. The only way to change it would be to cause an unbind/bind, which can be accomplished by toggling a conditional section off and back on. That is still the case here, but instead of waiting for the reference to appear up the context hierarchy, it just locks onto the current context. This _only_ affects ambiguous references, so any prefixed refs, like `.foo`, `../../bar.baz`, and `~/bar` are not affected.

~~For the `foo` question, yes. I think that can be addressed though, by having mappings resolve in their parent context. Actually, I'm wondering why they don't already...~~ They do, so never mind. That test was removed erroneously.

evs-chris:
Turns out I removed that test erroneously. I have reinstated it along with a working version of the explicit mapping test, which was just lacking the explicit mapping on the first component.

sarendipitee:
Looks good to me. Looks much simpler with little drawback.

PaulMaly:
Very bad idea, sad. (((

kouts:
This I think is quite dangerous in running projects, can something like a warning be added?

evs-chris:
There is a warning flag [`warnAboutAmbiguity`](https://ractive.js.org/api/#warnaboutambiguity) that will give you a console warning any time a reference doesn't resolve to an immediate context. That will also warn you on references that exist in a parent context, but it should be enough to get an idea of what needs to be stubbed out in data to avoid issues with unresolveds removal.

If you're loading data asynchronously and using non-isolated components, you can stub out the data to avoid resolving in the wrong context e.g. `new Ractive({ data: { foo: { bar: {} }, baz: {} } })` if your components are expecting to be able to access `foo`, `foo.bar`, and `baz` on the parent at some point.

I'm sorry that removing unresolveds broke your workflow. I do still see it as necessary due to the number of very buggy corner cases caused by unresolveds, two-way binding, and shuffling.

ghost:
I remember the days in Angular where I end up wasting my time "scope hunting" - hunting for the scope of some variable, only to find out it actually lives 5 controllers above it, which may or may not exist at times, and is operated by totally unrelated code which can be pulled out at moments notice.

Having a framework that enforces encapsulation/isolation and explicit mapping of data by default is very helpful.

Issue #2833, opened 2017-01-18
checkbox two-way binding issue


Calx-Takeno:
## Reproduction:
you can visit this URL
https://jsfiddle.net/calx/47fosorf/2/

## Code:none

<!--
Hello!
we find an issue above 0.7.3 about check two-way binding.

if you have fixed this , please tell me ,this is my email address 232613143@qq.com.

thank you for your efforts on git.

Calx-Takeno:
when we observe the value change, it will get the same value between oldValue and newValue according to the document said.

ghost:
Works on 0.6 https://jsfiddle.net/47fosorf/4/

evs-chris:
There's a question here as to the correct behavior of the binding. You're observing an array, say `var arr = []` and you push an element onto it e.g. `arr.push(1)`. Now, the array has changed, so the observer will fire, but the old value and the new value are still the same array `arr`. Only some part of the array has changed and not the array itself.

The question is, should the checkbox binding completely replace the array when it updates, so you get a new array, or should it keep the existing one. I personally lean toward keeping the existing one, as a checkbox name binding feels like it should be the equivalent of a series of push/splice operations to match what happens to the checkboxes involved.

Any thoughts?

sarendipitee:
>  I personally lean toward keeping the existing one, as a checkbox name binding feels like it should be the equivalent of a series of push/splice operations to match what happens to the checkboxes involved

Yep that makes most sense to me. Couldn't you still use `arr.*` anyways if you wanted to observe only push/slice ops?


Sidenote: The fact that this works at all is amazing. I think it's super convoluted to be storing multiple two-way bound checkboxes into a single array so am fine with undefined behavior there.

ghost:
I was having doubts earlier, but it does seem related to https://github.com/ractivejs/ractive/issues/2639.

evs-chris:
There is also the undocumented `observeList` function that takes a keypath, a callback, and options, and it will fire with an object that has inserted and deleted lists and a starting index. It doesn't really handle merges very well though, as they don't necessarily cluster around a starting index. I think we should probably rename it to `observeArray`, add in the oldIndex -> newIndex map for handling merges, and document it. It was added as somewhat experimental, but since the policy is now to add most new features as experimental pending feedback, I think this should be exposed more prominently.

https://github.com/ractivejs/ractive/blob/dev/src/Ractive/prototype/observeList.js
[...]

Issue #2529, opened 2016-04-25
change event in Ractive.js v0.8.0-edge -- discussion


Jeff17Robbins:
@Rich-Harris, three different comments in Ractive.js v0.8.0-edge src suggest that the `change` event is to be deprecated. 

version:    

```
Ractive.js v0.8.0-edge
    Fri Apr 22 2016 20:17:22 GMT+0000 (UTC) - commit af9dbf1e958a4c03827bafe0f70295c9a738733a
```

These are the 3 comments and related code:

https://github.com/ractivejs/ractive/blob/dev/src/global/runloop.js#L102

``` javascript
        // TODO deprecate this. It's annoying and serves no useful function
        const ractive = fragment.ractive;
        changeHook.fire( ractive, ractive.viewmodel.changes );
        ractive.viewmodel.changes = {};
        removeFromArray( ractives, ractive );
```

https://github.com/ractivejs/ractive/blob/dev/src/model/Model.js#L147

``` javascript
        // TODO deprecate this nonsense
        this.registerChange( this.getKeypath(), value );
```

https://github.com/ractivejs/ractive/blob/dev/src/model/RootModel.js#L16

``` javascript
        // TODO deprecate this
        this.changes = {};
```

Meanwhile, @evs-chris  has been doing non-trivial work to make the `change` event work in Ractive.js v0.8.0-edge.  

My interest in the `change` event is that we are currently using https://github.com/rstacruz/ractive-ractive (written by @rstacruz inspired by one of your examples, I think) with Ractive.js v0.7.3 and it works great.  

This adaptor relies on the `change` event.  This adaptor still doesn't work 100% correctly in Ractive.js v0.8.0-edge, but I bet we can get there.  It will take some more work.  I'm happy to continue working on it (mostly generating test scenarios -- as soon as I create a reproducible bug, it seems that @evs-chris  knocks it down, hence my confidence that we can get there!)

However, if the intent of the "deprecate" comments is that the `change` event is going to be deprecated, it would be great to understand the reasons and/or the suggested replacement for it.   

For example, I thought that maybe `ractive.observe('*')` would be a potential alternative to keeping up with all the changes, but have a concern about the loss of batching behavior -- a lot of changes get delivered atomically as a lot of .observe() callbacks instead of in one single `change` event callback -- which seems like a step backwards in efficiency?  So I am not confident that `ractive.observe('*')` is a true replacement for `ractive.on('change')`.

[...]

Issue #2014, opened 2015-06-25
State Management through `change` event


Aintaer:
It would be great if the `change` event fired with both the new and old values of the path that was changed. Specifically so that a state manager could track what has been changed and be able to apply the inverse operation.

fskreuz:
I think [`observe`](http://docs.ractivejs.org/latest/ractive-observe) does the same thing. http://jsfiddle.net/qfzq8hrs/.

This works for simple values, especially primitives. But Ractive mutates objects, so be warned. You can either clone them every time, or consider persistent data structures.

Rich-Harris:
I was discussing this in person with @Aintaer yesterday so we talked about whether `observe` would do the trick - tl;dr it doesn't, sadly! The problems start when *anything* might change, and you need to know specifically which keypath changed, and what it changed from and to:

```js
ractive = new Ractive({
  data: {
    foo: {
      bar: 'a'
    },
    baz: 'x'
  }
});

ractive.set( 'baz', 'y' ); // this needs to trigger a change handler...
ractive.set( 'foo.bar', 'b' ); // ...so does this...
ractive.set( 'foo', { b: 'c' }); // ...and so does this
```

Simply observing the root obviously won't work. The `change` event helps, because you get the keypath of the originating change, but it only reports the new value, not the old value. I *think* it would be straightforward to capture both old and new values, though we'd need to figure out what the event would be called and what its data should look like (since updating the `change` event itself would be breaking - though maybe that's okay since we'd be replacing it with something that can take care of existing use cases anyway.)

fskreuz:
@Rich-Harris since `change` just provides the new value as first argument, how about making it like `observe`, in that it has `oldValue` as second argument, and `keypath` as third? This wouldn't break anything since probably everyone expected only the first argument to exist. 

That's unless someone wrote a function that accepts and uses a second and third argument and is hooked to `change`. Reminds me of this funny question from an FP fan on our team before. Explaining made him cringe.

```js
// Why does this map out to [1, NaN, NaN, NaN] and not [1,2,3,4]?
['1','2','3','4'].map(parseInt)
```

aharpervc:
@fskreuz your FP fan should excited for an opportunity to use partial right application ;)

Rich-Harris:
I'd be lying if I said that parseInt thing hadn't caught me out before! You're totally right, using the second argument is logical and is consistent with `observe`. Let's do that.

evs-chris:
[...]

PR #2845, opened 2017-01-26
Various observer changes and additions - RFC

evs-chris:
This merges my list of TODOs for observers in 0.9 into one PR:

#### Make `observeList` an option of `observe`

The currently undocumented `observeList` method fires an observer with the parts of an array that have been added and removed rather than with old and new values. This does away with the special method and turns it into an option of `observe` so that it benefits from `once` and `defer` flags too.

```js
// old syntax
ractive.observeList('some.array.path', obj => ...);
// new syntax
ractive.observe('some.array.path', obj => ..., { array: true });
```

The `obj` in the callback is an object with `inserted` and `deleted` arrays of elements and a `start` index. It should probably be updated to handle the `merge` case as well.

#### Add a hook for managing 'old' values

Since Ractive does no defensive copying of arrays or objects, when you change something downstream of an object or array that is being observed, the old and new values in the observer are the same object, which leaves you with no way to get _actual_ old values out. This adds a hook in the form of an observer option `old`, which is a function that gets called with the current old value and target new old value every time the observer needs to update its old value.

```js
// this creates a shallow copy of the array to be used as the next old value
ractive.observe('some.array', (n, o) => ..., {
  old(prev, next) { if (next) return next.slice(); }
});

// when the observer fires, the old value will be the cloned array without the 42 on the end
r.push('some.array', 42);
```

Further, the context for the `old` function is an object with the Ractive instance as its prototype, meaning you can access any of the usual instance methods with `this`. You can also set whatever properties you like on `this` without affecting the Ractive instance. So if you always wanted the initial old value to fire in the observer callback, you could use

```js
old(o, n) {
  if (!this.hasOwnProperty('old')) this.old = n; // or some suitable clone
  return this.old;
}
```

I'm not positive about the context object. The default observe takes a `context` param that defaults to the Ractive instance. Perhaps this should use the same context or a new `oldContext` option?

#### `change` event is deprecated in code with insufficient observer functionality to replace it

The `change` event is kinda convenient, but it also has some deficiencies, not the least of which is that you _have_ to pay the performance penalty for it even if you aren't using it. A close second is that you don't get the old value along with the new value when the event fires. Another common complaint is that change events don't fire for mapped keypaths, but that is because the instance doesn't own the data that is changing.

To cover all of those issues, this adds support for a recursive pattern observer that uses `**` to indicate that any keys set from this point in the keypath down should be observed. This means that you can effectively simulate `change` events, except with old values included, by using `ractive.observe('**', ...)`. You can also limit the scope of the path that you want to observe by using `ractive.observe('some.path.**', ...)` and only keypaths downstream from `some.path` will fire changes. This is a bit different than the plain pattern observer because with `*` being observed, when setting `foo.bar`, only the `foo` change can be observed. So single `*` patterns only account for a single key in a keypath. The `**` allows _any_ downstream keypath to trigger the observer, so with `**` being observed, when setting `foo.bar`, the observer will be triggered with `foo.bar` rather than just `foo`.

There are some slightly odd scenarios that can arise from that, with the main one being that setting `foo.bar` in an empty model and then setting `foo` will result in the second observer call having an `undefined` old value. This is because the observer has never actually seen the `foo` path before. I think that could be addressed, but it would be a bit expensive, especially since the other case is already covered in a fairly expensive way.

The other case is setting `foo` to `{ bar: 'yep' }` and then setting `foo.bar` to 'something else'. This is currently handled by recursively filling in the old value map on the observer with child keys any time an object-like value is sent to the observer callback. This means that on the first set, the observer sees that the value `{ bar: 'yep' }` is object-like and recursively updates the old value map with the object, which also sets `foo.bar` to `'yep'`.

The final issue regarding mapped data not being observable is also address because there is an option `links` for recursive observers that when set to `true` will also fire callbacks for linked data. This defaults to false to match the behavior of the change event.

__NOTE:__ recursive observers cannot have more than one `**`, can't have any `*`s, and the `**` must be at the end of the keypath. Supporting additional wildcards would take considerably more code than I am comfortable adding for this.

##### Feedback, please
* [ ] Settle on an `old` context
*  ~~Maybe cover the extra corner cases on recursive observers?~~

##### Just needs to be done
* [x] Tests for recursive observers

#2833, #2529, #2014, #1990 

## Is breaking:
Sort of... `observeList` was undocumented, so I don't know if it should be considered a stable API.

I'd also consider tearing out the `change` event without a deprecation cycle. Just to be clear, my justification for that is _any_ time _any_ model changes, Ractive has to track the changes for the change event. Then _every_ time a runloop turn completes, the instance fires its change event, which bubbles all the way up the component hierarchy. With a recursive observer you opt in to expensive change tracking, whereas with the `change` event, there's no way to opt out.

## Reviewers:
Yep


ghost:
I remember @dagnelies using the `change` event.

Also, wondering if `old` was just a boolean config on the observer, and the defensive clone be internal to Ractive. Unless I'm missing something a use case where a function is really needed to defensively clone.

evs-chris:
I think `old` could be a boolean for the array case, especially when combined with `strict`, because it's easy enough to see that the value is an array and that `old === true` to do a `slice`. The object case is where it gets hairy. I originally proposed something similar to having a Ractive `deepClone`, but I think that would end up too complicated when a user could just pull in lodash, wrap `_.cloneDeep` (or whichever is most appropriate) in a helper function, and pass that along to the observer. See https://github.com/ractivejs/ractive/issues/2357#issuecomment-212964852 for the original discussion.

ghost:
But if the array can just `array.slice` (which is shallow), the object can just do `Object.assign` (also shallow). Besides, it's observing the array/object only, not the downstream properties (unless using wildcards), so it sort of makes sense just shallow-copying the object that actually changed.

evs-chris:
Only if the observer is `strict`. Any time a any child of the observed keypath changes the observer will fire. That's the other half of the array case, which is more prominent because when you `push` to an array you'd expect the old value to be shorter by one element. If the array is at `some.array` and `some.array.0.name` changes, the observer will still fire unless it's `strict`. In that case, the slice won't help you either, which is where the hook for a deep clone would come in.

If I added `old: true` support for arrays, doing the same for objects with `Object.assign` makes sense. Is there a better name for `old`, though?

Looking at it further, it seems that having the source keypath that caused the observer to trigger may be useful as well?

ghost:
> Any time a any child of the observed keypath changes the observer will fire. 

Oooh, wasn't aware of this functionality.

> Is there a better name for old, though?

👴 

kidding 😄

dagnelies:
Damn, I typed a lot but lost all because I accidentally closed the browser. :(

Indeed, I used `onchange` in various places. The main reasons were:

1. `onchange` was always stable (working) while *observers were always so full of bugs in every version it's not even funny*. 😛  Just have a look at #2430 to have small impression. It's just the tip of the iceberg.

2. several usage details. It's handier to have `onchange: function(vals) {/*my code*/}` instead of `oninit: function(){ this.observe('*', function(val) [/*my code*/})`. In `onchange` you get the exact keypath/value that changed instead of the first part and a big object. Lastly, you get called only once if multiple values are changed at once.

5. observers are slower than `onchange` by several order of magnitude! (see #2529)

...as a side note, the usefulness of onchange was kind of killed by #1990 since it doesn't work in components anymore.

So, as a summary, I have the feeling we're screwing a *good* `onchange` event in favor of *flaky* observers. ...honestly, in the past, it was really nice to have this `onchange` as fallback since the whole observer thingy was in my experience so buggy (and slow). It was already saddening to see `onchange` on components go away in 0.8.

evs-chris:
I _think_ all of the observer issues are handled, bug-wise on edge, and the rest here.

For the second, you can `observe: { '**'(val, old, path) { ... } }`, and the path will be relative to the observed path, meaning that with `**`, you'll get full keypaths. If you don't need to see _every_ keypath, using something like `foo.bar.**` will give you paths relative to `foo.bar`, and you won't have to pay the perf penalty for changes in other paths.

For the third, unfortunately wildcard observers are a bit expensive to maintain. I _think_ there have been some performance gains on them since that issue was opened, but even so, recursive observers _should_ be faster because they don't have to jump through hoops making sure the changed models line up exactly on the observed keypath.

The final two issues, firing on mapped paths and having the old values available, are also addressed by a recursive observer.

evs-chris:
I decided that trying to keep track of up- and downstream old values in recursive observers is probably more headache than it's worth. So for now, you'll get the last value that fired _on each a particular keypath_, meaning that `ractive.set('foo', { bar: 'yep' })` and `ractive.set('foo.bar', 'yep')` are two very different things, and the second will fire with an undefined value after the first because `foo.bar` has no previously encountered value.

The reverse case is also still true because trying to collect potential future downstream values while processing upstreams on `ractive.set('foo.bar', 'yep')` means that a huge amount of tree traversal could happen for what should be very simple operations.

dagnelies:
thanks for the infos. I really hope they'll be stable from now.

As a side note, I must say I never quite understood why onchange and observers were based on two different implementations. I'd have expected that the onchange part under the hood just notifies the adequate registered observers. That way, the performance issues and buginess would have been solved as well. But this is old history, you're way deeper into this and I assume there were some reasons.

I just care that it works well 😄 

---

...and having back the performance of onchange would still be a nice bonus, although much lower priority. Observers kind of die once you have large data. (with observers http://dagnelies.github.io/ractive-dataset/big_edge_o.html vs without http://dagnelies.github.io/ractive-dataset/big_edge.html)

evs-chris:
@dagnelies in trying to profile your examples to see if I could remove some performance bottlenecks, I think I found the reason that the observer example is so slow. Unless I'm misreading it, it looks like the view is lazy-loading rows, which means that the models and supporting view get created on demand; however when a (non-recursive) pattern observer is used, it has to traverse the _entire_ model, including non-initialized portions, looking for matching keys. The volume of keys that have to be traversed on that set is pretty enormous. With a recursive observer, it wouldn't preemptively traverse the model, it would just fire the observer on matching changes.

dagnelies:
Yay 👍 Everything seems fine then!

...and thanks for your patience during those explanations, I'm a bit critical sometimes.

mefjuu:
Will the changes be documented at docs.ractivejs.org?

evs-chris:
Yes, but they haven't been updated yet.

PR #2846, opened 2017-01-27
Second swing at event delegation - RFC

evs-chris:
This is the follow-on to #2841, where instead of adding manual insertion points for delegates, iterative sections that are nested within an element automatically set their containing element as the delegate target. Fragments automatically inherit their parent delegate, and there's special casing for handling components.

Unlike the first swing, this does not allow delegating custom events because there's really no way to control exactly which events should be delegated. As such, the event directive has to check to see if there is a matching custom event in the hierarchy. This adds a bit of overhead over the first swing too, so it looks like the performance gap drops from 20-30% to 15-25%. On the plus side, you can't accidentally break custom events, because they will still be rendered on the element containing the directive where the DOM events may be shuffled upstream.

To be clear, any events that are within an iterative section that is within an element will automatically be delegated. If there's not a wrapping element, then there's nowhere to install real listener. Nested iterative sections inherit their parent delegate target, so there's only one target element. Turning off delegation is still a TODO.

To complete the comparison, here's the adjusted example from the first swing

```html
<table>
{{#each lotsOfRows}}
  <tr>
    <td><button on-click="@.select(.)">Select</button></td>
    <td>{{.description}}</td>
    <td>{{.something}}</td>
    <td><button on-click="event.splice('../../', @index, 1)">Remove</button></td>
  <tr>
{{/each}}
</table>
```

### Questions

1. Does tracking the delegate target element seem lightweight enough to justify the render gains?
2. The search for ractive events to fire should be pretty darn fast, but I've heard rumors that delegation can cause lag with things like touch scrolling. Would that be a problem here? With jquery, I believe the delegated events have to execute a query selector, which is where the lag generally comes from?
3. If this is ok, should it go out defaulted to on or off?
4. If lag could be a problem, would having a directive for elements to disable delegation be useful? Something like `<div no-delegate>{{#each ...}}...{{/each}}</div>`, perhaps.

### TODO

* [x] Add an instance flag to disable delegation
* [ ] Maybe have a directive for enabling/disabling delegation explicitly at one point (element containing iteration)?
* [x] Tests for disabling delegation

The remaining bits of #846

## Is breaking:
Kinda sorta... if delegation defaults to off, it's not breaking. If it defaults to on, it _may_ be a little breaking due to the way bubbling can get out of order.

## Reviewers:
Yep

### DIFF
--- a/src/Ractive/config/defaults.js
+++ b/src/Ractive/config/defaults.js
@@ -1,7 +1,8 @@
 export default {
 	// render placement:
 	el:                     void 0,
-	append:				    false,
+	append:                 false,
+	delegate:               true,
 
 	// template:
 	template:               null,
@@ -11,7 +12,7 @@ export default {
 	tripleDelimiters:       [ '{{{', '}}}' ],
 	staticDelimiters:       [ '[[', ']]' ],
 	staticTripleDelimiters: [ '[[[', ']]]' ],
-	csp: 					true,
+	csp:                    true,
 	interpolate:            false,
 	preserveWhitespace:     false,
 	sanitize:               false,
--- a/src/Ractive/construct.js
+++ b/src/Ractive/construct.js
@@ -34,6 +34,11 @@ export default function construct ( ractive, options ) {
 	initialiseProperties( ractive );
 	handleAttributes( ractive );
 
+	// if there's not a delegation setting, inherit from parent if it's not default
+	if ( !options.hasOwnProperty( 'delegate' ) && ractive.parent && ractive.parent.delegate !== ractive.delegate ) {
+		ractive.delegate = false;
+	}
+
 	// TODO don't allow `onconstruct` with `new Ractive()`, there's no need for it
 	constructHook.fire( ractive, options );
 
--- a/src/config/types.js
+++ b/src/config/types.js
@@ -51,3 +51,4 @@ export const EVENT             = 70;
 export const DECORATOR         = 71;
 export const TRANSITION        = 72;
 export const BINDING_FLAG      = 73;
+export const DELEGATE_FLAG     = 74;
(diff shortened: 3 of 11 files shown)


evs-chris:
There is now an option `delegate` that defaults to `true`. Delegation is configured by the component hierarchy of the template rather than the component prototype hierarchy alone. If a component has a set `delegate` flag, it will be used. Otherwise, it will use the setting of the parent instance in the template.

ghost:
> I've heard rumors that delegation can cause lag with things like touch scrolling.

Delegation has performance issues when the delegate target is way up in the tree. If I remember correctly, this was why `jquery.live` was replaced with `jquery.on`. `jquery.live` used `document` as the predetermined delegate target while `jquery.on` allows the user to pick an ancestor lower in the tree.

Given that jQuery already provides `jquery.on` to mitigate this issue, any issues that still exist probably developer error, like attaching too many scroll handlers, very slow scroll handler performance, and stuff like that.

This case would be comparable to `jquery.on` as it's scoped within the bounds of the component, which is presumably small and only a few elements deep. 

> If this is ok, should it go out defaulted to on or off?

- If it currently breaks, it's safer to be off. But a global switch (i.e. `Ractive.defaults.delegate = true`) would be nice for those adventurous enough to try it.
- If it is safe to do so or if nobody really cares about order, on by default would be nice.

I personally don't care about bubble order. It's kind of expected especially when dealing with dynamic objects, even when dealing with jQuery.

The only time I care about order is when I need to stop propagation, like stopping clicks inside a modal from closing a modal that is configured to close when clicking outside of it (the close handler was on `<body>`). And this already felt hacky and edge-case to me, considering I could have put a transparent non-ancestor shade to capture the outside click instead of putting the handler on `<body>`.

My 2c

PR #2848, opened 2017-01-29
WIP: Build improvements

ghost:

Goals:

### Karma for testing

Removes the headache of setting up scripts and tools for testing. Setups for Phantom, Saucelabs, Coveralls is anticipated.

### Separation between gobblefile and scripts

Currently, steps for bundling are split between the Gobblefile and build.sh. The idea is to move the bundling entirely to the Gobblefile so you don't need to meddle with scripts on development, and that scripts can focus more on the CI side of things.

### A more predictable import structure

Currently import paths are magical, particularly in tests, because they reflect paths caused Gobble's merging of directories and not the actual filesystem structure. Causes head-scratching at times. The idea is to make Gobble adjust so that imports actually reflect filesystem structure.

### Test only on the latest versions

I intend to only test on a single version of everything, both Node and browsers. That means the latest of Node and the latest of each browser. I may reinstate testing on older versions of platforms in the future, but for simplicity, this will do for now.

ToDos:

- [x] Karma Phantom setup.
- [x] Karma Sauce Labs setup (disabled at the moment, don't want to spam it yet).
- [x] Karma Coveralls setup (just needs credentials, not sure what's encrypted on the yml).
- [x] Prevent Karma's Progress reporter from spamming the CI output.
- [x] Update the gobblefile to do all bundling duties.
- [x] Update gobblefile to make import paths identical to filesystem paths (src and test only).
- [x] Adjust import paths in source and tests.
- [x] Get rid of the 404 error caused by Karma not being aware of that gif Ractive is loading.
- [x] Update build script to do CI stuff.
- [x] Make sure the karma-coverage and karma-coveralls report to Coveralls.
- [x] Reinstate old release scripts.
- [x] Polyfills in mjs format.
- [x] ~Find a way to run coverage on node tests and merge them in the report.~ Postponed.
- [x] ~Investigate why Sauce Labs tests randomly fails.~ It's a Ractive timing issue on slow systems. Sauce labs will still run and show errors but won't break the build.
- [x] ~Review available tools that do auto-versioning and auto-publishing.~ Postponing to post-1.0 since auto-publish tools like semantic-release will release a breaking version when current changes break previous release's tests. Not good if we're still trimming and finalizing features . Otherwise, we would be like v15 by now.
- [x] ~Prevent QUnit cli from emitting Ractive's console messages on the Node tests.~ Replaced QUnit cli with Electron to test Node on the same Karma setup.

Just putting this PR out there for comments. This may never fly or scope may be trimmed down for another PR. Your ideas are very much welcome. 😄


evs-chris:
That build looks much cleaner! I haven't quite made it all the way through yet, but I have two things to note:

1. Publishing to npm from the travis build was, uh, special to get working. The main thing to look for is that it can _only_ publish from one of the node version builds, which is currently 6.0. There are also branch checks that have to run to make sure that only 0.8 and dev get published. When 0.9 is published, that branch will need to be added for tracking.

2. Rollup actually has `noConflict` support built in now. I believe if you pass `noConflict: true` to the rollup options, it will stick one in there for you.

Also, I'm sorry I just made a bunch of conflicts for you.

ghost:
@evs-chris Yep, have to take note of the multi-node build thing. And the conflicts are normal. Happens every time I move around the tests. Not a big deal tho. 😄

ghost:
OMG it actually passed 😱 

Will put this one up for a while before force-pushing another change and/or reorganizing commits. 😄

ghost:
Random tests failing in random browsers, have to check why. Probably Sauce Labs being slow and all. But this one consistently fails in Safari OS X after 3 or so reruns of the same build:

```
Safari 10.0.1 (Mac OS X 10.11.6) failed specs:
  plugins/transitions.js > js timing functions can be used with css transitionable properties (#2152)
	failed, expected argument to be truthy, was: false
	Expected: true
	Actual: false
	promiseReactionJob@[native code]
```

evs-chris:
Interesting. I don't have access to safari, so I can't really debug it. What happens when you set `node.style.transitionTimingFunction` to an invalid value? I was originally going to use a whitelist, but that seemed terribly brittle.

ghost:
Test passes on actual Safari 🤦‍♂️

evs-chris:
Transitions tests are notoriously finicky, especially with browser automation. Is there a way to soft fail tests with karma? Maybe something like a passing quorum? It'd be nice to know they fail, but with more browsers being tested I don't know that conditional exclusion is as feasible as when you mostly just have to watch for phantom issues.

ghost:
Yep. Karma is set to `singleRun` mode so it acts like a script, which I believe can be bypassed with `-e`/`+e`. But it would be nice to get this setup perfect. Just a bit more tweaking. 😄

ghost:
If you're wondering why there's a build in Coveralls whose coverage is 64%, that's me configuring karma-coveralls incorrectly in this PR. 😓

evs-chris:
I think it would be good to export a `polyfills.mjs` too, so that anyone wanting to bundle their app with ractive and legacy browser support could do so by simply `import polyfills from 'ractive/polyfills'; import Ractive from 'ractive';`. The polyfill import may require the extension under rollup and/or webpack, but the module and jsnext:main settings in the package should have the main ractive module covered.

ghost:
Might need eyes on this one. It's sort of done. :D

evs-chris:
Nice!

I made a few minor changes, and merged with dev (sorry, I would have rebased, but I didn't want to blow up the world). It looks like maybe sauce isn't liking something now, and I allowed the electron test to fail for build:local because my dev environment has no available gui.

Do you know why travis might be missing eslint?

ghost:
It might have to do with travis using cached npm dependencies. Not sure if it applies across branches. Try removing the following from the yaml:

```
cache:
  directories:
  - node_modules
```

evs-chris:
Aha! There were no node modules in the build dir. Throwing an `npm i` in the build script fixed that. Now just need to get the release script sorted and all will be well again.

ghost:
I also adjusted the scripts. The `bundle` emits `.build` and ~`.publish`~ `.release` (formerly `tmp` and `build`) and the scripts will look at `.publish`.

evs-chris:
It looks like the shell environment that npm run provides may not work with the script. Setting it up to run directly from the travis command seems to work ok.

Issue #2810, opened 2016-12-26
Make 'event' the original DOM event, deprecate `event.original`


sarendipitee:
This is something that's annoyed me for a long time because it *always* trips up new members of my team.

If you do something like this `<button on-click=' doSomething(event.clientX, event.clientY)'>` I think it's fairly natural to assume that the `event` there is a DOM event. But it's not. And that's seriously confusing.

I feel slightly embarrassed telling a new team member that you have to use `event.original` to access the "actual" event.

Historically it was fine because it was easy access to the [helper](http://docs.ractivejs.org/edge/method-calls) methods, but now that we have `@` references in event method calls, I think it's time to deprecate the special "event object" that ractive creates.

There's no need for `event.pop()`, `event.get()`, `event.animate()` etc when you can use `@this.get()`, `@this.animate()`.

Then this would allow us to make `event` the *real* event object and remove `event.original`.

Proxy events would still behave the same way (for now).

Thoughts?

evs-chris:
Apologies in advance for the rambling, but I've thought a bit about cleaning up events before.

We still need some way to access context methods, because they are already scoped to the current context. For instance, if we're at keypath `foo.bars.7.bazzes.12.bat`, using `event.set('wat', 10)` will set `foo.bars.7.bazzes.12.bat.wat` to 10, whereas `@set('wat', 10)` will set the root keypath `wat` to 10. The context specific methods are _required_ for calling methods on keypaths in a computed context (template expressions).

I _think_ the reason that the DOM event is pushed off into the `original` property is to have a common interface between DOM events and plugged-in events. There were also some additional properties hanging off of the event, like context and index that have been done away with since.

I've thought that perhaps it would be a good idea to split the source event and the event context apart for a while now. I just haven't come up with a good way to do it. There's also a bit of conflation between DOM-triggered proxy events and manually fired events that confuses things. There's some weirdness around `fire` that needs to be addressed, especially when used from an event directive. It's a little better without proxy event args, but I still think it could be made better. Not automatically making the event the first argument would improve the interface, I think. Perhaps moving the args into the event object would be better. I'm not sure, but that doesn't really directly relate to this issue.

[...]

Issue #1671, opened 2015-01-29
ractive.update() does not update data functions


kurdin:
I have code where I want to update whole UI if `lang` set to new value, instead if `custom_fn` I use `translate` function with return from `I18n-js` library but I removed it from example to make it simple. 

`App.update();` does not trigger `data.custom_fn()` so UI not updated. 

``` javascript
App = new Ractive({
    el: '#app',
    template: '#template',
    data: {
      lang: 'en',
      custom_fn: function(s) {
          return s + ' Sergey current Lang is ' + this.data.lang;
      }
    }
});

App.observe('lang', function(lang) {
    console.log('Current lang is: ' + lang);
    App.update(); 
});

App.set('lang','ru');
```

Template:

``` html
{{ custom_fn('Hello') }}
<p>Lang is: {{ lang }}</p>
```

output is:

```
Hello Sergey current Lang is en
Lang is: ru
```

and shoud be

```
Hello Sergey current Lang is ru
Lang is: ru
```

http://jsfiddle.net/du7v1q1c/

Am I doing something wrong?

martypdx:
If you use the api in the custom function, it works automatically without the observe (see http://jsfiddle.net/du7v1q1c/1/):

```js
App = new Ractive({
    el: '#app',
    template: '#template',
    data: {
      lang: 'en',
      custom_fn: function(s) {
          return s + ' Sergey current Lang is ' + this.get('lang');
      }
    }
});

App.set('lang','ru');
```

Does that work for you?

kurdin:
Thanks @martypdx, this works but in theory that should be working with observe and App.update() as well right? 

Let say I want to change language in ``I18n-js`` only once  when ``lang`` set to another value:

```javascript 
App.observe('lang', function(lang) {
     I18n.locale = lang;
     console.log('Current lang is: ' + I18n.currentLocale());
     App.update();
});
````
with code you provided I need to set ``I18n.locale`` for each template value
````javascript 
model = {
      lang: 'en',
      translate: function(s) {
        I18n.locale = this.get('lang');
        return I18n.t(s);
      }
};
````
This will work but not efficient as with observe, where I change ``I18n.locale`` only once and then update UI with ``App.update();``

martypdx:
Ah, I see. If you use the translator directly (or place the function off an object), you can get it to work by using the object as the update keypath (see http://jsfiddle.net/du7v1q1c/3/):

```html
{{ i18n.t('Hello') }}
[...]

PR #2853, opened 2017-02-02
Split event into `@context` and `@event` - RFC

evs-chris:

This adjusts `event` and context support in a few mostly minor ways:

1. All items and fragments can now lazily supply a context object that is a pointer into their context. The context object can resolve relative keypaths as they exist in the template at a given point in the vdom. As opposed to edge as it stands now, where only elements can supply context objects. The difference is subtle, but cases like `<div>{{#with foo}}...here?{{/with}}</div>` are not addressable with element-only contexts. This is not a primary goal for this PR and is mostly exposed for symmetry.

    `@context` is only available as an expression and not a model/reference as `@this` and normal data are, so it can't be used for bindings. It can be used in other expressions though, which makes it handy for decorators, passing context handles out via expressions, and, its primary purpose, events. It might also be pretty handy to set up an alias to a context and then pass it into a component further down the template. Turning a context into a fully fledged model and passing it around might be a good way to achieve out-of-data state support.

2. Instead of taking the ractive event object and adding context helpers to it, this starts with the context object for the event directive, creates a new object with the context as its prototype, and assigns all of the event's properties to the new object. This exposes the context for the event along with any additional properties set on the event while giving the event priority.

3. In addition to `@context.original`, the source event (usually a DOMEvent), is exposed as `@context.event`. This allows the slightly less confusing destructure syntax `ractive.on('someEvent', ({ event }) => ...)` to be used rather than `ractive.on('someEvent', event => /* the actual event is at event.original */)`. In the template, the source event is also exposed as `@event`, which is only a valid reference in an event directive. `@context.original` is left in place so that existing plugins don't break.

    `@node` is also available to event directives as a reference to the DOM element with which the event directive is associated. This is handy for delegated events where the target is not necessarily the same as the element with the event directive. It also means you can do stuff like `<input on-focus="@node.select()" />`.

4. If the current event is being re-fired, its name will be updated to match the new name passed to `fire`. I believe this is the last remaining difference between `<button on-click="foo">` and `<button on-click="@.fire('foo', @context)">`.

    This also means that the default first argument to an `on` callback should now be described as context rather than an event. Having an optional context argument in an event callback makes a bit more sense to me than having an optional event argument.

5. The `event` reference available to event directives is deprecated and to be replaced by `@context` and `@event`.

6. `resolveReference`, which is responsible for taking a reference keypath and mapping it to a model, has been refactored to be a bit easier to follow. I don't know if optimizing compilers still have function length limits, but if they do, this should also be more optimizer-friendly.

7. Eliminates the sometimes there, sometimes not event parameter for the `on` callback by changing the context of the event callback to `@context` with the implicit event's properties assigned to it. That makes `ractive.on('foo', ev => ev.name /*etc*/)` become `ractive.on('foo', function() { this.name /*etc*/ })` but is more developer friendly for events that may be fired from both API and DOM.

8. This also adds a new proxy event handler that is a specialization of an expression event. If the expression returns a single array with a string as the first element, then it will effectively `@this.fire( ...array)`.

### Thoughts?

Looking at contexts from the perspective of decorators and custom events, it seems to me that it would make more sense to call the plugin init function in the context of a context object rather than the ractive instance. The ractive instance is still exposed, but there are more relevant methods directly available with the context. One of the first lines in most of the decorators I write these days is also `this.getNodeInfo(node)` to grab the context.

There's also no way to have decorators and custom events take advantage of event delegation. It would be nice to have a way to do so, and I think augmenting a context object with `listen` and `unlisten` methods would work nicely.

#2810 #1671

## Is breaking:
~~No.~~ Yes. It adds a deprecation ~~, but everything else is backwards compatible~~. The context of the event handler is also changed from the Ractive instance to a context object with the event's properties assigned to it.

## Reviewers:
Yep


evs-chris:
It has also occurred to me that using a context object as the context for event handlers would eliminate the need to pass the ractive event around when using `fire` or proxied events. Any arguments passed to fire either directly of via proxy would be the _only_ arguments rather than having an event first argument in some situations, and `this` in the callback would _always_ have the event info in it rather than only if the `fire` call happened to supply it.

sarendipitee:
So much goodness here, definitely a step in the right direction!

`event.original` was very confusing to every person I've introduced Ractive too.

> The ractive instance is still exposed, but there are more relevant methods directly available with the context.

Isn't the instance exposed as `this` anyways? So you could have `context` be the second arg for decorators? `function(node, context) { ... }`? I think that's deserving of its own topic outside of this pull though.

evs-chris:
What do you think about adding support for argumented proxy events based on the first expression in the list evaluating to a string e.g. `on-click="`foo`, someArg, { and: [ 'an' ], object: too }"` would evaluate the array, see that the first arg is the string "foo", and proxy using those args e.g. `fire('foo', someArg, { and: [ 'an' ], object: too })`. If the `@context` is `this` for the event callback, then there's no need to worry about it being passed as the first arg in some cases but not others. I suppose it could also be made a little safer by checking for an array with a first string element as the only element of the expression list e.g. `on-click="['foo', someArg, ...]" with the plus that that stands out and the minus of magic.

> Isn't the instance exposed as `this` anyways?

It is. If DOM event callbacks get a `@context` context (`this`) though, I think it would make sense to go ahead and do the same for other template-based constructs like decorators, custom events, and transitions. That _is_ a mostly unrelated issue, though. I'm trying to break all the things now rather than later because I'm starting to think 0.10 should be 1.0. The version after 0.9 should have one major breaking change for adaptors, and the rest _should_ be backwards compatible, which would make it a good candidate for a "stable" release and more strict semver.

Issue #2664, opened 2016-09-22
Idea: option to use CDATA sections or comments to split up adjacent text nodes in toHTML for enhance


evs-chris:
HTML merges adjacent text nodes, which is problematic for progressive enhancement where adjacent VDOM nodes may end up being rendered as text. CDATA sections are not rendered, but they are still part of the document~~, unlike comments~~. There could be an option for `toHTML()` to output CDATA nodes next to non-element VDOM nodes that would allow the enhancement process to acquire individual text nodes instead of grabbing the first (and only) one and leaving the remainder of the textual VDOM nodes to insert additional text nodes.

Perhaps a similar solution could be applied to triples where a special CDATA content could be used as a marker for all of the nodes that should belong to the triple.

martypdx:
>  still part of the document, unlike comments

Comments _are_ accessible via `childNodes` of parent element. But maybe you're referring to something else?

evs-chris:
Huh. So they are. I'm not sure where I got that.

MartinKolarik:
Wouldn't it be better to simply treat `Hello {{name}}!` like one text node instead of three? This idea came up before and I think I even started with the implementation back then but didn't finish it.

evs-chris:
I also started trying to implement something like that, but I gave up when trying to figure out how to handle things like `foo{{#if bar}}<span>yep</span>{{/if}}bar` where nodes would need to split and rejoin.

MartinKolarik:
Hmm that indeed looks problematic. I guess using comments would be easier.

Issue #2403, opened 2016-02-24
Progressive enhancement with triple mustaches doesn't work


monoblaine:
I think the progressive enhancement feature has some problems regarding triple mustaches.

The following example (http://jsfiddle.net/monoblaine/h516ecvz/1/) runs fine:

``` js
setTimeout(() => new Ractive({
    el: 'ractivable-el',
    template: '{{txt}}',
    enhance: true,
    data: { txt: 'Hello!' }
}), 6000);
```

This one (http://jsfiddle.net/monoblaine/tqhs0bmb/1/), however, unnecessarily recreates the text node: (not the best example for triple mustaches, I know)

``` js
setTimeout(() => new Ractive({
    el: 'ractivable-el',
    template: '{{{txt}}}',
    enhance: true,
    data: { txt: 'Hello!' }
}), 6000);
```

evs-chris:
Triple mustaches and progressive enhancement together are problematic conceptually too. We can probably fix the text node issue with triple mustaches, but I don't think their actual use-case will work with with progressive enhancement - at least not as it exists now. Since there are likely tags inside the triple, it would have to render itself to a document fragment, then match the actual DOM with the fragment DOM and try to correct as it went. Is that better than throwing in the rendered fragment and stripping out the leftover tags on the way to the next element?

martypdx:
What about an `.innerHtml` type compare for triples? Broad try to see if html is same.

evs-chris:
Ah, but `.innerHTML` on what? Triples may be other-thing™ adjacent.

monoblaine:
> Is that better than throwing in the rendered fragment and stripping out the leftover tags on the way to the next element?

I can't say anything about this, but if implementing it isn't worth the trouble (or the performance benefit, if any) using triple mustaches may be discouraged for those who are obsessed with progressive enhancement.

PR #2855, opened 2017-02-02
Support non-destructive sets in conditional sections - RFC

evs-chris:

When a rendered conditional section's condition goes false-y, the section is completely torn down and unrendered. For tiny chunks of template, that doesn't mean very much, but for larger things that means that the teardown is expensive and setting it back up when it goes truth-y again is very expensive. It also means that any components within that section are entirely destroyed and recreated in the same cycle, which can have unfortunate side-effects for components with private internal state.

This is my attempt to address those issues by adding a flag to the mutation methods `set`, `add`, `subtract`, and `toggle` that tells any sections that unrender during the course of updates propagating from the model mutation not to completely destroy their fragment but, instead, just unrender it and hang on to it for future re-rendering. The flag is passed in a new options object as the last parameter of those methods e.g. `ractive.toggle('foo', { keep: true })`. The default behavior is still to tear everything down.

### TODO

* [x] Tests

I haven't done any extensive testing yet, but this theoretically supports transitions properly. I also haven't specifically tested it with components, but I have verified that DOM is not completely torn down and recreated when `keep === true`.

Just aggravation I've experienced since trying to write a tab component that used conditional sections before giving up and resorting to `display: none` instead, which is not at all transition friendly.

## Is breaking:
Nope.

## Reviewers:
Yep.

### DIFF
--- a/perf/src/app/runSuite.js
+++ b/perf/src/app/runSuite.js
@@ -157,6 +157,7 @@ function runTest ( context, test, version, ractiveUrl, callback ) {
 function runStep ( context, test, callback ) {
 	let start = now(), count = 0, runStart;
 	let duration = 0, totalDuration = 0;
+	const times = [];
 
 	console.group( test.name );
 
@@ -170,17 +171,24 @@ function runStep ( context, test, callback ) {
 			return callback( e, { name: test.name } );
 		}
 
-		duration += now() - runStart;
-		totalDuration = now() - start;
+		const end = now();
+		duration += end - runStart;
+		totalDuration = end - start;
+		times.push( end - runStart );
 
 		if ( test.maxCount && count >= test.maxCount ) break;
 	}
 
+	const avg = times.reduce( ( a, c ) => a + c, 0 ) / times.length;
 	callback(null, {
 		name: test.name,
 		duration,
 		count,
-		average: duration / count
+		first: times[0],
+		average: avg,
+		variation: times.reduce( ( a, c ) => a + Math.abs( avg - c ), 0 ) / times.length,
+		minimum: Math.min.apply( null, times ),
+		maximum: Math.max.apply( null, times )
 	});
 
 	console.groupEnd( test.name );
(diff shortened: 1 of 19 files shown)


ghost:
Sounds like another one of those "potentially an optimization" things or just do manual hiding.

In the case of optimization, what's the downside of hanging onto it by default? If I remember correctly, I read somewhere that Inferno does this to make it real fast, where it doesn't completely destroy stuff but instead holds on to them for future use.

But then, one can simply just do manual hiding using CSS if one really wants to preserve the DOM. In the case of tabs, it's simply a `ractive.set` toggling the some value that controls the presence of a class on each tab and tab content. If I need animation, I jujst pop in CSS transitions with numeric properties like `max-width`, `max-height` and `opacity` in place of `display:none`.

evs-chris:
There are pros and cons to both keeping and throwing away DOM and vDOM. If you keep it and you actually don't want it any more, then that's a bit more garbage sitting around that can't be reclaimed. It can add up pretty quickly with complex apps, and ractive is already at a bit of a disadvantage there because the model tree is never pruned. This is probably a bigger deal on mobile. For reference, my tests show a ~75%/25% improvement on render/unrender when `keep: true` is used.

As for manual CSS transitions, that's how I've handled it too, but it leaves a bit to be desired. Ractive intro transitions have the tremendous benefit of being able to measure exactly how big an element needs to be when the transition ends, which makes scaling/sliding/fly type transitions much more pleasant. Getting a transition sequence to end with a `display: none` and start with `display: block` (or whatever) also requires a bit of timing and class voodoo.

evs-chris:
There are now tests, and in adding them, I discovered that nodes were being recreated rather than reused. That is now fixed.

I also got a bit carried away and did some progressive enhancement fixes while updating kept fragment support. Triples will now reuse nodes during initial render with `enhance: true`. Text nodes and interpolators will use `Text.splitText()` rather than overwriting the initial node and adding new ones. I'm gonna call that close enough for now, so this will now fix #2664 and #2403.

PR #2858, opened 2017-02-03
Remove support for live queries - RFC 0.9

evs-chris:
A live query is a `findAll` or `findAllComponents` query that when passed `{ live: true }`, will keep the returned array up to date as new elements/components are added and removed. Note that there is no notification that goes along with that, it just keeps the array updated, and doing so adds a fair bit of housekeeping every time an element/component is added or removed. That housekeeping happens whether you're using live queries or not.

This feature can be emulated by using a special decorator for the `findAll` case or listening to component `render`/`unrender` events for the `findAllComponents`.

I'll leave this open for a week or so, ask the community for feedback, and if there are no objections, merge it before 0.9.

Part of #2798 

## Is breaking:
Yep.

## Reviewers:
Yep.


sarendipitee:
Didn't even know this existed! All for removing it if it improves perf/reduces complexity.

giovannipiller:
Never used them.
I can't think of any good reason to keep them around, especially if removing them would result in better performance/maintenance.

aharpervc:
> Didn't even know this existed! All for removing it if it improves perf/reduces complexity.

Ha, same. I wonder how many other "hidden gems" there are going on inside ractive?

Issue #1907, opened 2015-04-08
merge() on Hashmaps


StritziK:
Please correct me if I'm wrong but I didn't find a good way to perform a merge of two hashmaps.
I have for example two hashmaps with different users:

``` javascript
Ractive.get('users') // returns something like { 'user_1': {...}, 'user_2': {...},  'user_3': {...} }
var newUsers = { 'user_4': {...}, 'user_5': {...},  'user_6': {...} };
```

It would be nice if there would be a support for:

``` javascript
Ractive.merge('users', newUsers, true)
```

-> http://jsfiddle.net/ayc8h9b0/
Currently with this approach nothing happens...
Is there an actual solution I am missing, except doing it with Ractive.set() and overwrite all Users which triggers rerendering of the whole userlist ...

martypdx:
Not currently supported, through an interesting idea - there seems to be increasing requests for dealing with hashmaps.

The merge function _replaces_ the existing array with the new array while reusing any surviving members. In this case don't you want to append the keys as well as replace?

In any case, you can add your own function to `Ractive.prototype`, for example (see http://jsfiddle.net/zbanyr3p/):

```js
Ractive.prototype.mergeHash = function(keypath, hash){
    var obj = this.get(keypath), key;
    for(key in hash){
        if ( hash.hasOwnProperty(key) ) {
            obj[key] = hash[key];
        }
    }
    this.update(keypath);
};
```

> which triggers rerendering of the whole userlist

Take a look at the nodes in the example, Ractive reuses them. :)

evs-chris:
`merge` on a hashmap is effectively a deep set, and that is now implemented in edge.

Issue #550, opened 2014-04-06
Question: how to deep set or extend an object


jonathantneal:
How might I deep set or extend an object within data in Ractive?

If my data contains nested objects like so:

``` json
{
    "deep": {
        "foo": true,
        "bar": false
    }
}
```

How might I update the object without clobbering the other properties?

``` js
ractive.set('deep', {
    bar: true
});
```

**Example**: http://codepen.io/jonneal/pen/1e15853515369b6d362729fb8fe029a0

**Related**: #79

Don’t mistake my question for a complaint. I prefer that set overwrites objects, but I would also like the ability to update an object.

---

If not already available, could this be a separate method? I am looking for something, perhaps, like this:

``` js
// NOTE: identical function exists in project
function utils_isObject(value) {
    return value && Object.prototype.toString.call(value) === '[object Object]';
}

// extend first object with all remaining objects
function utils_extendObject() {
    var
    // move all objects into collection
    objects = [].slice.call(arguments),
    // separate first object
    object = objects.shift();

    // for each object in collection
    objects.forEach(function (nextObject) {
        // set or extend its keys to first object
        Object.keys(nextObject).forEach(function (key) {
            if (utils_isObject(object[key]) && utils_isObject(nextObject[key])) utils_extendObject(object[key], nextObject[key])
            else object[key] = nextObject[key];
        });
    });
}
```

evs-chris:
edge now has `ractive.set('some.path', someComplexObject, { deep: true })` which overwrites the overlapping leaves of the model tree with `someComplexObject` aka a deep set.

PR #2859, opened 2017-02-04
Add deep set/object merge as an option on set

evs-chris:
This adds an option `{ deep: true }` on `set` that, instead of replacing the existing data, merges the new data into it. So with `{ foo: { bar: 1, bat: 2 } }`, calling `ractive.set('foo', { baz: 99 }, { deep: true })` will result in the data being `{ foo: { bar: 1, bat: 2, baz: 99 } }`.

Handily, this also works with arrays using both arrays and objects with numeric keys e.g. with `{ foo: [ 1, 2, 3 ] }`, `ractive.set('foo', { 1: 42 }, { deep: true })` and `ractive.set('foo', [99], { deep: true })` will both do what you'd expect. The keypath can also be root e.g. `''`.

#550 #1907

## Is breaking:
Nope.

## Reviewers:
Yep.

### DIFF
--- a/src/Ractive/prototype/set.js
+++ b/src/Ractive/prototype/set.js
@@ -1,8 +1,10 @@
 import { build, set } from '../../shared/set';
 
-export default function Ractive$set ( keypath, value ) {
+export default function Ractive$set ( keypath, value, options ) {
 	const ractive = this;
 
-	return set( ractive, build( ractive, keypath, value ) );
+	const opts = typeof keypath === 'object' ? value : options;
+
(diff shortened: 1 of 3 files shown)


evs-chris:
In an interesting pivot, Travis has decided to be fickle while AppVeyor is actually working. For the last three days or so Travis has been failing jobs that say they're successful with exit 0 at the end. Any ideas?

ghost:
I think it's a Travis bug. The build results look flaky, they appear to fail on a random Node.js version across your recent PRs.

Issue #1885, opened 2015-03-31
Support for CoffeeScript classes syntax


suprMax:
Hi!

I'm using CoffeeScript extensively in my app. It provides it's own classes system (http://coffeescript.org/#classes) with ability to call `super` to invoke parent's class method.

When I tried to do `class View extends Ractive` it works for the most part, but fails once you try using partials or components.

Here's an example from Backbone how they handle extends. It works great and I can extend any Backbone class "natively" instead of calling `Backbone.extend`: https://github.com/jashkenas/backbone/blob/master/backbone.js#L1779

Would it be possible to have similar functionality? It seems that most of the functionality you already have anyway (`this._super()` etc). I think CoffeeScript users such as myself would greatly appreciate if it works out of the box natively!

martypdx:
Does it work to do something like:

```js
function CoffeeRactive(options){
   return Ractive.extend(options);
}
```
and go from there?

Rich-Harris:
I think @suprMax is talking about using standard inheritance semantics. I've wondered about this, in the context of ES6 classes (which are somewhat similar to their CoffeeScript counterparts).

The problem is that the semantics are a bit different. When you call `A = Ractive.extend(...)`, followed by `B = A.extend()`, you're not just creating a constructor whose prototype inherits from the parent; 'registries' (such as `partials` and `components`, as you note, but also `transitions`, `events`, `decorators` and so on) plus `data` and `computed` all extend the corresponding properties on the parent.

More than that, it allows Ractive to do things like only compiling templates once at `extend` time, rather than at instantiation time. So it's a lot more efficient if you're extending a component and only making minor changes (for example, `WarningModal = Modal.extend({ data: { type: 'warning' } })`, or something) - everything 'rolls up' neatly, saving both you and the computer work.

So I don't really know. React recently moved in this direction (preferring `class Foo extends React.Component` over `React.createClass()`, and it's fair to say it's causing some of their users a bit of pain. Going with the grain of the language is generally a good thing, but I think we'd lose a lot of convenience in the process. It would involve quite a wide-ranging rethink, I think.

martypdx:
As long as you're willing to forgo registry rollup (or handle on your own), I'm thinking that the extend options could be treated as the "class":

```js
function RactiveClass(options) {
    var Extended = Ractive.extend( Object.getPrototypeOf(this) );
    return new Extended(options);
}
```
Then that should enable:
```js
class Widget extends RactiveClass {

    template: '#widget'

[...]

PR #2866, opened 2017-02-09
Allow use of existing constructors (classes) in extend - RFC

evs-chris:
This adds support for calling Ractive.extend with an existing constructor, meaning that you can use an ES6 class or the appropriate function construct usually created by js-sugar languages to create a new component/view. The only two constraints are

1. the constructor has a Ractive as a prototypal ancestor
2. the constructor calls super

Both of these are checked during `extend`, though the super call is a bit of a loose check. If the checks don't pass, `extend` will throw an error.

```js
// Ractive could also be another extension
class Foo extends Ractive {
  constructor(opts) {
    super(opts);
    this.foo = doSomethingMeaningfulOnEveryConstruction();
  }
  
  go() {
    this.set( 'gone', 'probably' );
  }
}

// you still have to call extend, which handles
// setting up all of the fun things like parsing
// templates, munging attributes, and rigging
// on and observe listeners
Ractive.extend({
  class: Foo,
  attributes: { ... },
  on: { ... },
  template: '...'
});

// again, Ractive could be other extensions

const f = new Foo({
  target: '#some-el'
});

f.go(); // probably gone
```

### Why?

I was wondering how we could go with the ES6 (and maybe typescript, coffeescript, livescript, etc) grain a little more readily, and it occurred to me that there's no hard requirement for Ractive to create the constructor. It just needs to decorate it if everything lines up right.

That and it really irritates me that _every_ component is just `Child`. With this you can control the constructor name 😄 

#1885 and maybe a few other closed issues

## Is breaking:
Not really, unless you use `class` as an attribute in `extend`

## Reviewers:
Yep.

### DIFF
--- a/src/Ractive.js
+++ b/src/Ractive.js
@@ -3,7 +3,7 @@ import easing from './Ractive/static/easing';
 import interpolators from './Ractive/static/interpolators';
 import { magic, svg, win } from './config/environment';
 import proto from './Ractive/prototype';
-import extend from './extend/_extend';
+import { extend, initClass } from './extend/_extend';
 import parse from './parse/_parse';
 import getNodeInfo from './Ractive/static/getNodeInfo';
 import construct from './Ractive/construct';
@@ -48,6 +48,7 @@ Object.defineProperties( Ractive, {
 
 	// static methods:
 	extend:         { value: extend },
+	initClass:      { value: initClass },
 	escapeKey:      { value: escapeKey },
 	getNodeInfo:    { value: getNodeInfo },
 	joinKeys:       { value: joinKeys },
--- a/src/extend/_extend.js
+++ b/src/extend/_extend.js
@@ -8,9 +8,7 @@ import unwrapExtended from './unwrapExtended';
 
 const callsSuper = /super\s\(|\.call\s*\(\s*this/;
 
-export default extend;
-
-function extend ( ...options ) {
+export function extend ( ...options ) {
 	if( !options.length ) {
 		return extendOne( this );
 	} else {
@@ -18,7 +16,11 @@ function extend ( ...options ) {
 	}
 }
 
-function extendOne ( Parent, options = {} ) {
+export function initClass ( Class, options = {} ) {
+	return extendOne( this, options, Class );
+}
+
+function extendOne ( Parent, options = {}, Target ) {
 	// if we're extending with another Ractive instance...
 	//
 	//   var Human = Ractive.extend(...), Spider = Ractive.extend(...);
@@ -29,12 +31,10 @@ function extendOne ( Parent, options = {} ) {
 		options = unwrapExtended( options );
 	}
 
-	let Child, proto;
+	let proto;
+	let Child = typeof Target === 'function' && Target;
 
-	if ( typeof options.class === 'function' ) {
-		Child = options.class;
-		delete options.class;
-
+	if ( Child ) {
 		if ( !( Child.prototype instanceof Parent ) ) {
 			throw new Error( `Only classes that inherit the appropriate prototype may be used with extend` );
 		}
@@ -64,6 +64,7 @@ function extendOne ( Parent, options = {} ) {
 
 		// extendable
 		extend: { value: extend, writable: true, configurable: true },
+		extendClass: { value: initClass, writable: true, configurable: true },
 
 		// Parent - for IE8, can't use Object.getPrototypeOf
 		_Parent: { value: Parent }
--- a/src/extend/_extend.js
+++ b/src/extend/_extend.js
@@ -6,6 +6,8 @@ import initialise from '../Ractive/initialise';
 import Ractive from '../Ractive';
 import unwrapExtended from './unwrapExtended';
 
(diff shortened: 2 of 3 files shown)


sarendipitee:
This is super cool! I recently just spent a while trying to figure out that I had to now call `_super()` in 0.9.x, so this looks much cleaner. I think if we can document this well we can market Ractive with better first-class Components a la React/Inferno.

evs-chris:
I must confess I've never had occasion to call `_super` anywhere. Out of curiosity, what have you had to add that to that you didn't also have to do in 0.8? I was under the impression that it would only be required for lifecycle methods like `onconstruct` and `oninit` for extended components.

ghost:
Wondering if the extra call to `Ractive.extend` can go away though. That way, the component is purely just a class like React. The presence of the extend call kinda defeats the purpose of the class syntax, and would tell one to just roll with the old functional syntax to avoid complexity.

Also, with regards to `_super()`, I have never used it ever. Inheritance was a fun idea before, until it doesn't work. I prefer composition, only extend from `Ractive` and compose components from smaller components. The last time I tried inheritance resulted to this https://github.com/ractivejs/rvc/issues/5. 😄

evs-chris:
The `extend` is required because there's a bunch of static setup that Ractive needs to do (attribute handling, events and observers, etc) that aren't available on the prototype. You _could_ put those in the constructor, but that would be pretty terrible to use from my perspective (`constructor (opts) { opts.template = '...'; opts.data = function() { ... }; opts.css = '...'; super(opts); }`). There's also a very decent perf bump for pre-parsing templates and partials and having the registries set up before instantiation. According to Rich's comments in #1885 React caught some grief for the perf hit when they switched to classes over factories.

If ES ever adds static constructors, then we can probably sunset `extend`. Until then, I think the benefits far outweigh the slightly awkward extra call.

ghost:
Wondering if we could also avoid that `class` initialization property so that it won't be misused on regular `Ractive.extend` or `new Ractive`. Maybe a separate static function like `Ractive.initializeClass()` would do to avoid confusion. Also, I'm wondering if `static` ever landed on ES6. [Babel seems to have it](https://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=false&presets=es2015%2Creact%2Cstage-2&targets=&browsers=&builtIns=false&code=%20%20class%20Bork%20%7B%0D%0A%20%20%20%20%2F%2FProperty%20initializer%20syntax%0D%0A%20%20%20%20instanceProperty%20%3D%20%22bork%22%3B%0D%0A%20%20%20%20boundFunction%20%3D%20()%20%3D%3E%20%7B%0D%0A%20%20%20%20%20%20return%20this.instanceProperty%3B%0D%0A%20%20%20%20%7D%0D%0A%0D%0A%20%20%20%20%2F%2FStatic%20class%20properties%0D%0A%20%20%20%20static%20staticProperty%20%3D%20%22babelIsCool%22%3B%0D%0A%20%20%20%20static%20staticFunction%20%3D%20function()%20%7B%0D%0A%20%20%20%20%20%20return%20Bork.staticProperty%3B%0D%0A%20%20%20%20%7D%0D%0A%20%20%7D). If so, then that would avoid defining things in 2 places by defining properties that need to be set up as static. That's unless the properties space of the constructor function is already taken for something else.

```js
class Foo extends Ractive{
  // Stuff for initialization
  static template = `
    // HTML
  `
  static css = `
    // CSS
  `
  // Prototype stuff
  constructor(){
    ...
  }
  yourMethod(){

  }
}

// Returns an initialized Foo
export default Ractive.initializeClass(Foo);
```

evs-chris:
Yeah, I think a method specifically for this makes more sense. `extendClass` doesn't seem to fit quite right, and I wanted to avoid `initialize` vs `initialise`, so I went with `initClass` 😄.

evs-chris:
After sleeping on it, I don't like `initClass` as much, as it doesn't really say that the new class will/needs to definitely be a sublcass of the callee. I'm going to call this good:

```js
class Page extends Ractive {
  constructor (opts) { super(opts); }
  navigate() { ... }
}
Ractive.extendWith(Page);

class Home extends Page {
  constructor (opts) { super(opts); }
  something() { ... }
}
Page.extendWith(Home, {
  template: '...'
});
```

If that turns out to be lousy, it's going to be flagged experimental and can change before or after 0.9.

ghost:
Been reading about Angular 2 (or whatever version it is now) and I'm curious of their use of [decorators](https://www.typescriptlang.org/docs/handbook/decorators.html). For vanilla, decorators [are still a proposal](https://github.com/wycats/javascript-decorators) and Babel [has a transform](https://babeljs.io/docs/plugins/transform-decorators/) for it. The way I understand, the following could be possible:

```js
// MyComponent.js
@ExtendWith({
  // Stuff needed on extend
})
export default class MyComponent extends Ractive{
  // Stuff needed on instantiation
};
```

evs-chris:
It looks like that might be a possibility at some point, but right now the Bable repl just throws an error if you try any of their examples 😞. Based on playing about in the typescript playground (supposedly tracking es closely, right?), I'd say it will work nicely.

PR #2868, opened 2017-02-09
BREAKING: Second run at normalizing the event things (also proxy syntax) - RFC

evs-chris:
This is take two on #2853. Instead of trying to move the event context object to the context of the callback, this takes the opposite tack and makes the context object _always_ the first parameter. So no matter what or where the event is `fire`d, the first argument to the callback is always a context that will have the name, original event (also named event for `({ event }) => ...` reasons), component, and other relevant properties stuck on it. ___Always_ passing the context as the first arg, event for lifecycle and component-only events, is the breaking change__. Detecting that the first argument to `fire` is already a context/event got much easier, so you can re-fire events without sometimes getting an extra argument passed to your callback. Additionally, the context is not considered the first argument in `arguments[n]` or `$n` notation anymore, because it's not really an event argument.

The other bits are mostly here too, just in a cleaner way. This adds `@context` everywhere for between-elements scopes and `@node` and `@event`, which are the relevant dom node and dom event, for event directives.

It also adds special handling for event directives that gives back most of the convenience of the old proxy events with none of the drawbacks. With a minor patch that went in prior to this PR, it is now possible to have decent unquoted attribute support for directives, so `<a on-click=['foo', bar, 444]>...</a>` is now possible. There's nothing really sugared about that, it just kinda looks like it because the whole quote pileup is avoided. When the `a` is clicked, the directive will see that its expression list resolved to a single array with a string as the first element and will effectively `@this.fire('foo', @context, bar, 444)` for you. This means that `ractive.on('foo', ({ event }) => alert(event.clientX))` will do exactly what you'd expect it to (alert with the x coordinate of the mouse click event).

#2810

## Is breaking:
Quite.

## Reviewers:
Yes.


ghost:
I personally don't mind having context as first arg for `ractive.fire`. It's common practice to receive some sort of information object as first arg of event handlers (jQuery for one). And 👍 for the `event` destructuring thingy. I also assume this doesn't affect method calls, that method calls remain explicit (what you pass in is what the other end receives).

I'm a bit confused on this one tho: 

> Additionally, the context is not considered the first argument in arguments[n] or $n notation anymore, because it's not really an event argument.

Also, what happened to the `proxy-click=""` something syntax to separate things that accept expressions vs ones that have the funky proxy synax? I sort of kinda liked that (just by the looks of it, dunno about implications).

evs-chris:
> I also assume this doesn't affect method calls

That's correct. The behavior of `fire` has changed slightly, such that if you pass a context as the first argument, it will be merged into the fired context - basically proxying the existing event. There was already sort of a check for that, but it's much more effective now.

> I'm a bit confused on this one tho:

When you fire and event from a component e.g. `this.fire('myEvent', 42)`, on the component template you could have a handler installed that does something with the event at the boundary e.g. `<myComponent on-myEvent="doFoo()" />`. There is special handling in event directives for arguments passed to events that exposes them as `arguments` e.g. `on-myEvent="doFoo(arguments[0])"` or numbered `$n` vars starting with `1` e.g. `on-myEvent="doFoo($1)"`. Prior to this change, where the event/context may or may not have been present, the event/context was considered and argument. Now it is only exposed as `@context` and the arguments are just those passed beyond the context.

> Also, what happened to the proxy-click=""

I figured it may end up being event more confusing to have two different directives for the two different event types. I'm actually wondering if we shouldn't kill off the special handling for `<a on-click="foo">` and just add a check for a single string value in the expression result to do the same thing. With the patched up unquoted arguments, that would make ```<a on-click=`foo`>``` be the equivalent. Without using backticks, it's a little less pretty though `<a on-click="'foo'">`. Of course, since the special funky parsing for event directives was removed in 0.9, any of the quotes can be mixed and matched because the expression parser will consume the nested quotes e.g. `<a on-click=""foo"">`, not that that is any better. Just worth noting 😄

sarendipitee:
> so <a on-click=['foo', bar, 444]>...</a> is now possible. There's nothing really sugared about that, it just kinda looks like it because the whole quote pileup is avoided.

That... is... kinda strange.

> Without using backticks, it's a little less pretty though <a on-click="'foo'">. Of course, since the special funky parsing for event directives was removed in 0.9, any of the quotes can be mixed and matched because the expression parser will consume the nested quotes e.g. <a on-click=""foo"">

And now we're in Season 2 of Stranger Things. 

Maybe it's time we kill off all these weird syntaxes and just go with method call. That's it; just method calls. Events are still there, you just always have to explicitly `fire()` them. It just seems like you're trying to bend over backwards trying to support all these sugar syntaxes and it's ending up even more convoluted?

evs-chris:
> That... is... kinda strange. And now we're in Season 2 of Stranger Things.

Fair enough 😆. I was also looking at it from the perspective of decorators and transitions where you often have a string as the first/only argument and this kinda hurts the eyes `<div as-tracked="'foo'" fade-in="'fast'">...</div>`. Whereas the unquoted string template version results in exactly the same thing ```<div as-tracked=`foo` fade-in=`fast`>...</div>``` because backticks can't be used for attribute quotes but can be used for strings. So that gives you a way to use strings in unquoted directive values.

It's expressions all the way down, so the code is actually simplified on the Ractive side.

LeJared:
> and this kinda hurts the eyes `<div as-tracked="'foo'" fade-in="'fast'">...</div>`

Actually it does not. This kind of mixed quoting is fairly common in JS world. That's why lot of JS developers stick to single quotes for strings anyway. And every HTML editor handles those kind of attribute content quite well.

> ``<div as-tracked=`foo` fade-in=`fast`>...</div>``

_That_ looks strange to me. Does any HTML editor even support that kind of attribute, at least not failing with syntax highlighting?

BTW: those ` are quite hard type on a none-english keyboard and some layouts don't have that key at all. That's why most programming languages avoid them whenever possible.

evs-chris:
To be clear, I'm not suggesting we switch to backticks for quoting anywhere. I was just noting that they're an option for less noisy string literal attributes now that we have string template support.

PR #2869, opened 2017-02-09
Better CSS at-rules handling (fixes #2854)

### LINKED ISSUE #2854: Using `from` or `to` in CSS animations will break transformation
## Description:

Code like the following:

```
  @keyframes someAnimation {
    from {
      transform: scale3d(1.5,1.5,1) rotate(0deg);
    }

    50% {
      transform: scale3d(1,1,1) rotate(90deg);
    }

    to {
      transform: scale3d(1.5,1.5,1) rotate(0deg);
    }
  }
```

Will be wrongly converted to the following. Notice the `[data-ractive-css]` attributes;

```
@keyframes someAnimation {
    from[data-ractive-css~="{328a91fd-7095-2748-09c4-42b189b78cb8}"], [data-ractive-css~="{328a91fd-7095-2748-09c4-42b189b78cb8}"] from {
      transform: scale3d(1.5,1.5,1) rotate(0deg);
    }

    50% {
      transform: scale3d(1,1,1) rotate(90deg);
    }

    to[data-ractive-css~="{328a91fd-7095-2748-09c4-42b189b78cb8}"], [data-ractive-css~="{328a91fd-7095-2748-09c4-42b189b78cb8}"] to {
      transform: scale3d(1.5,1.5,1) rotate(0deg);
    }
  }
```

In the meantime, a workaround to this is using `0%` and `100%` rather than `from` or `to`, in your CSS animations.

## Versions affected:
tested on 0.8.10 and edge

## Reproduction:
https://jsfiddle.net/7qxqrkks/1/

MartinKolarik:
Excludes contents of `@keyframes someAnimation {...}` blocks from the transformation process.

#2854

## Is breaking:
No.

### DIFF
--- a/src/Ractive/config/custom/css/transform.js
+++ b/src/Ractive/config/custom/css/transform.js
@@ -1,7 +1,7 @@
 import cleanCss from '../../../../utils/cleanCss';
 
-const selectorsPattern = /(?:^|\})?\s*([^\{\}]+)\s*\{/g;
-const commentsPattern = /\/\*[\s\S]*?\*\//g;
+const selectorsPattern = /(?:^|\}|\{)\s*([^\{\}\0]+)\s*(?=\{)/g;
+const keyframesDeclarationPattern = /@keyframes\s+[^\{\}]+\s*\{(?:[^{}]+|\{[^{}]+})*}/gi;
 const selectorUnitPattern = /((?:(?:\[[^\]]+\])|(?:[^\s\+\>~:]))+)((?:::?[^\s\+\>\~\(:]+(?:\([^\)]+\))?)*\s*[\s\+\>\~]?)\s*/g;
 const excludePattern = /^(?:@|\d+%)/;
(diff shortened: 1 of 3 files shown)


MartinKolarik:
Note that this doesn't make `@keyframes` scoped, it only fixes the cases when Ractive would break the CSS. Making `@keyframes` scoped would most likely be quite complicated and I think the chance of collision is relatively small. Maybe a `warnOnceIfDebug()` message when we detect `@keyframes` would be a good idea here?

evs-chris:
Having not authored many keyframes myself, this may be a terrible idea, but could the transform just note the names and do a search and destroy with scoping at the end of the process?

MartinKolarik:
@evs-chris if you mean renaming the animation and all its usages the problem is that simple search & replace would break stuff if the animation name was also used in other places (part of property name/value). To do it safely, we'd need more complicated parsing.

evs-chris:
Yeah, probably not worth it, but at least there's a solid starting point if it comes up in the future. Good to go?

PR #2872, opened 2017-02-11
fix #2871 infinite loop on non-ractive el

### LINKED ISSUE #2871: Infinite loop in EventDirective
## Description:
The EventDirective gets into an infinite loop when it hits a node without a `_ractive` property.

## Versions affected:
Affected: 0.9.0-build-72
Okay: v0.8.1

## Platforms affected:
All

## Reproduction:

I'm having trouble creating a test case for this, but you can reproduce it on a project of mine by:

1. CAUTION: be aware this test gets in a loop and will lock Firefox requiring a force-close. Chrome is fine, but the tab needs to be force-closed.
1. Visit http://proofer-polls-apib.surge.sh/#answers
2. Double-click the paragraph text

Expected: text is selected
Actual: this bug is hit. 100% CPU, event loop is blocked, need to force-close the tab.

## Code:
???

AshKyd:
Check for failure case & bail on the DelegateProxy while loop.

Most of this is whitespace changes. No test here because I'm having trouble reproducing it outside of my project. Please let me know if there's anything else I can do with this one.

#2871

## Is breaking:
No.

### DIFF
--- a/src/view/items/shared/EventDirective.js
+++ b/src/view/items/shared/EventDirective.js
@@ -29,17 +29,18 @@ export const DelegateProxy = {
 			// starting with the origin node, walk up the DOM looking for ractive nodes with a matching event listener
 			while ( bubble && node !== end ) {
 				const el = node._ractive && node._ractive.proxy;
-				if ( !el ) break;
 
-				// set up the context for the handler
-				event.node = el.node;
-				event.name = name;
+				if ( el ) {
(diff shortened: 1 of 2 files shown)


evs-chris:
Nice! I added a test and updated it a bit so that the non-ractive nested node events can still trigger the ractive listeners. Thanks!

Issue #753, opened 2014-05-28
Problem with ractive.merge?


martypdx:
Am I completely misunderstanding `merge`? Is it a union or intersection?

Take a look at http://jsfiddle.net/LxAG7/, it replaces the first item in the array and I would expect it to add it because `timestamp` is not equal. It doesn't work with latest either, which makes me think I'm not getting it.

``` js
var ractive = new Ractive({
    el: document.body,
    template: '#template',
    data: {
        message: []
    }
})

var i = 1;
setInterval(function(){

    var message = { 
        message: 'hello' + i++,
        timestamp: Date.now()
    }

    ractive.merge('messages', [ message ], { compare: 'timestamp' } )

}, 1000)
```

guilhermeaiolfi:
Well, until now I've never used ```merge``` because I thought it would merge/union the data. #1947 brought me here. So, +1 from me to change this behavior for/before 1.0.

PR #2873, opened 2017-02-11
BREAKING: Move `merge` to the `shuffle` option of `set` - #753

evs-chris:
Per #753, the `merge` method is confusing because it doesn't actually do a merge of the data at all. What it _actually_ does is set the target array to the new one while shuffling any associated DOM based on the new indices for the elements in the array, which are determined using the comparator.

By far the most common case for this would be to use object identity to merge an array into itself or merge an array created from the target array back into it. So this makes `{ shuffle: true }` the identity merge. That moves the `JSON.stringify` comparator to `1` ( `{ shuffle: 1 }` ), which is probably entirely unnecessary, but the option was already there. The equivalent call would be `{ shuffle: JSON.stringify }`.

At some point in the future, we _may_ want to add and actual array merge that does a union of the two arrays using the comparator, but I think that's best left to another library for now.

#753

## Is breaking:
Quite.

## Reviewers:
Yep.


evs-chris:
On second thought, `JSON.stringify` is much less mysterious than `1`, so I've removed the built-in support for JSON compare.

PR #2874, opened 2017-02-11
BREAKING: Remove support for mixin components - RFC

evs-chris:
There was some extension code that had no test coverage that I was trying to get covered, and nothing that I did with mixing `extend` with mixins and all manner of functions and settings would get it to trigger. The harder I looked at it, the less sense it made, and it was all in effort to support
```js
const Human = Ractive.extend({ ...human stuff?... });
const Spider = Ractive.extend({ ...spider stuff?... });
const Spiderman = Ractive.extend(Human, Spider);
```
That seems lovely, but I've never seen it actually used and would be seriously confused if it was anything other than simply adding methods and properties. With that, you'd be better off just using object literals for `Human` and `Spider`. In practice, the only multiple inheritance that I've ever seen play out in a reasonable way is scala's traits, and even they have some sharp edges.

I decided that instead of bashing my head against this trying to get it tested, it would probably be better to simply remove it. I haven't even seen a component extending a component in the wild at this point, never mind using component mixins, so I'm guessing it's fairly safe to do.

I'll leave this open for a few days to see if anyone has any objections, and if not, I'll merge it for 0.9.

## Is breaking:
Yes, but only if you use component mixins.

## Reviewers:
Sure.


ghost:
If one really wants to use mixins, it's better to do it explicitly rather than let Ractive decide for them.

```js
const SpiderMixin = {...};
const HumanMixin = {...};

const Spider = Ractive.extend({ ...SpiderMixin });
const Human = Ractive.extend({ ...HumanMixin });
const SpiderMan = Ractive.extend({ ...SpiderMixin, ...HumanMixin });
```

evs-chris:
Note that you can still pass multiple config objects to `extend`, so
```js
const SpiderMixin = { ... };
const HumanMixin = { ... };
const Spiderman = Ractive.extend(HumanMixin, SpiderMixin);
```

giovannipiller:
Just to be clear, it's still going to be possible to do:

```javascript
const Human = Ractive.extend({ ...human stuff?... });
const Spiderman = Human.extend({ ...spider stuff?... });
```

However, it will NOT be possible to do something like:

```javascript
const Human = Ractive.extend({ ...human stuff?... });
const Spider = Ractive.extend({ ...spider stuff?... });
const Spiderman = Human.extend(Spider);
```

Am I getting this right?

evs-chris:
@giovannipiller that is correct. You can extend multiple times, but the hierarchy must be linear (to the root, the same super can be extended multiple times, but a sub can only extend a single super).

giovannipiller:
Ok. Will things like this still work as of now?
https://jsfiddle.net/rtw5eLnt/1/

From my basic understanding of the code changes (and `_extend.js`), the answer is yes.

I'd love to try it myself, but for some reason my environment is going crazy today 😭
(failing `npm run build` on branch 'no-multi')

---

Pasting the code here for future reference:

```javascript
const additionalData = {
  data() {
    return {
      data1: false,
    }
  },
  onrender() {
    console.info('Additional render #1')
  }
}

const additionalData2 = {
  data() {
    return {
      data2: true,
    }
  },
  onrender() {
    this._super();
    console.info('Additional render #2')
  }
}

const MyComponent = Ractive.extend(additionalData, additionalData2, {
  data() {
    return {
      myComponentData: 2,
    };
  },
  el: '#main',
  template: '#template',
  onrender() {
    this._super();

    // expected data: {data1: false, data2: true, myComponentData: 2}
    console.info('Main render, with data:', this.get());

    // onrender, 3 console.info should have been fired by now
  }
});

const r = window.r = new MyComponent();
```

evs-chris:
@giovannipiller I really wasn't sure, so I added roughly that as a test case. All is passing, so, yes, that still works as expected 😄. The only things that should be affected by this are `Ractive.extend(OtherComponent)`. Plain old objects should still work as expected because `Ractive.extend(add1, add2, add3)` is effectively `Ractive.extend(add1).extend(add2).extend(add3)`.

giovannipiller:
Awesome!
In our app, we have a couple occurrences to refactor, but nothing that really makes sense keeping around 😊

---

Last point which might be worth considering: should Ractive throw some sort of error/warning?
It might help other people upgrading to 0.9, especially since things might not explicitly break (extend would still run).

```javascript
if ( options.prototype instanceof Ractive ) {
  // throw error?
}
```

evs-chris:
@giovannipiller that's probably a good idea, so done. :+1:

aharpervc:
> However, it will NOT be possible to do something like

Does this mean that making a ractive component "base class" won't work any more? For example:

```js
const CompBase = Ractive.extend({ pojo });
const MyCompInstance = CompBase.extend({ other_pojo });
```

or, will this keep working? It's hard to tell from the discussion so far.

evs-chris:
@aharpervc yep, that still works.

PR #2885, opened 2017-02-17
BREAKING: Remove magic and array adaptors for 0.9 - RFC

evs-chris:
I was originally going to put off removal of the magic and array adaptors until 0.10, but thinking about it, they're a bit buggy and not really all that convenient with their gotchas. They _could_ be pulled into plugins as they stand now, but unless someone actually needs them, I'm not in any hurry to do so. Getting #2512 solved would make the magic adaptor much more useful, but I'm not convinced that the array adaptor should ever come back.

Any thoughts?

#2798

## Is breaking:
Yep.


Jeff17Robbins:
This one caught us by surprise -- must have missed the posting that said that the `modifyArrays` option would be going away.  What's the level of effort to keep it available as a plugin? 

We're not sure how much code, if any, we have that relies on the feature.  But we are running with `modifyArrays: true`.

evs-chris:
It would probably take a few hours to take the array adaptor and remove all of the dependencies on internal APIs. If you _aren't_ using modified arrays anywhere, you're definitely better off without the adaptor from both performance and potential footgun perspectives.

ghost:
@Jeff17Robbins If I remember correctly, both magic and array adaptors are essentially just regular adaptors. Ractive just happens to come with them. It might be possible to just grab the old code, modify it a bit, and hook it up like a regular adaptor.

Here's [the adaptors page from the docs](https://ractive.js.org/Extend/Adaptors/) should you need any help in migrating it.

Jeff17Robbins:
Thanks!  We will have to weigh what to do. I think removing any use of this
from our code makes the most sense, but nice to know we have an option.

PR #2887, opened 2017-02-20
Post-parsing transforms, view-local data, and maybe `@state` with nested computations - RFC

evs-chris:
In looking at how to implement something like materializecss in pure ractive, it feels excessive to create a component for every button and card you add to the view. That lead back to the lightweight components PR, but even that seems like too much for simple things that could easily be a partial with a few parameters.

That brought me around to "what if we could just turn what looks like a regular component into a sort of macro that replaced itself with other template". So that's what this is:

1. `Ractive.parse` now considers `Ractive.defaults` rather than using hard-coded values if no options are passed, meaning that `Ractive.defaults.delimiters = ['<%', '%>']` will actually work with `Ractive.parse('...')` now. It also means that `Ractive.defaults.parserTransforms` would actually mean something.
2. The parser now accepts a `transforms` array or the global/instance `parserTransforms` which is an array of functions to be called with each element in the parsed template. Each function may remove, modify, or replace each element template with which it is called. It is called with `Ractive` as the context, so `parse` is available locally without having to pull in Ractive as a dependency.
3. To facilitate having small bits of local state, this introduces `@data`, which is a root model that is created as needed on the current `@context`, with the idea being that the transform sets up something like `<div>{{#with @data as state}}...{{/with}}</div>` as the replacement template. Then any `state` references within the template won't interfere with normal data, but will instead live only for the lifetime of the outer `div` element in the replacement.
4. `@data` refs are also available to `@context` objects, so you can interact with template-local context from outside of the template.

This is sort of the reciprocal to the `@state` PR, which would be the same thing, but hanging off of the internal model object rather than the view. I considered introducing `@state` here too, but there are a few questions surrounding that and probably a little bit of confusion around the difference between `@state` and `@data`. One of the questions is should `foo.bar.@state.foo.bar` be the same as `foo.bar.foo.@state.bar` or should it be its own independent tree? There's also some question as to what should happen with state on linked models.

At any rate, `@data` is not really meant to be used for general purposes, nor are transforms for that matter. Transforms require a fairly intimate knowledge of the compiled template format, but I don't expect many people to want to write their own widget libraries beyond the level of components. If you _do_ need to achieve something like a partial with paramaters, including passed-in inline partials, this would allow it, though in a component-like form rather than `{{> partial}}` form. Attributes, child elements, inline partials, and other content are all fairly easy to access, move around, and massage.

### Why?

One of the major goals for 0.9/0.10 is to have a nice component suite like materialize or vuetify available to use alongside a ractive-based router/app framework that uses anchors to swap views in and out. Part of any of the better component suites are small things like buttons and cards that may be rendered _a lot_ in various places, so low overhead is important. Most of those _could mostly_ be handled with css alone, but they would be a lot more ergonomic and complete as component-like constructs.

### Questions

1. Is there a better name for `@data`.
2. Right now the return of the transform determines what happens with the node being transformed. I think it's probably a bit limited with `false` removing the node and an object replacing the node. I think maybe it should be an object that specifies what to do instead, like maybe `{ remove: true }` for remove, `{ replace: { ... } }` or `{ replace: [ ... ] }` to replace it with one or more nodes, and `undefined` for nothing?
3. Should figuring out what exactly to do with `@state` go ahead and happen? It would give stable out-of-data storage, meaning it wouldn't interfere with normal data, it wouldn't disappear with an unrender/re-render, and it has been requested a fair amount. It would also give a nice entry point for nested computeds, which remains one of the oldest requested features.

## Is breaking:
Nope.

## Reviewers:
Yep.

### DIFF
--- a/gobblefile.js
+++ b/gobblefile.js
@@ -155,7 +155,7 @@ function skipModule(excludedModules) {
 			// Gobble has a predictable directory structure of gobble/transform/number
 			// so we slice at 3 to slice relative to project root.
 			const moduleRelativePath = path.relative(__dirname, modulePath).split(path.sep).slice(3).join(path.sep);
-			return excludedModules.indexOf(moduleRelativePath) > -1 ? 'export default null;' : src;
+			return excludedModules.indexOf(moduleRelativePath) > -1 ? 'export default null; export const shared = {};' : src;
 		}
 	};
 }
--- a/src/Ractive.js
+++ b/src/Ractive.js
@@ -38,6 +38,7 @@ Ractive.defaults = Ractive.prototype;
 
 // share defaults with the parser
 shared.defaults = Ractive.defaults;
+shared.Ractive = Ractive;
 
 // static properties
 Object.defineProperties( Ractive, {
--- a/src/Ractive.js
+++ b/src/Ractive.js
@@ -11,6 +11,7 @@ import initialise from './Ractive/initialise';
 import { getCSS } from './global/css';
 import { escapeKey, unescapeKey } from './shared/keypaths';
 import { joinKeys, splitKeypath } from './Ractive/static/keypaths';
+import shared from './Ractive/shared';
 
 export default function Ractive ( options ) {
 	if ( !( this instanceof Ractive ) ) return new Ractive( options );
@@ -34,6 +35,9 @@ Ractive.prototype.constructor = Ractive;
 
 // alias prototype as `defaults`
 Ractive.defaults = Ractive.prototype;
+
+// share defaults with the parser
+shared.defaults = Ractive.defaults;
 
 // static properties
 Object.defineProperties( Ractive, {
--- a/src/Ractive/prototype/fire.js
+++ b/src/Ractive/prototype/fire.js
@@ -4,7 +4,10 @@ import Context from '../../shared/Context';
 export default function Ractive$fire ( eventName, ...args ) {
 	// watch for reproxy
 	if ( args[0] instanceof Context ) {
-		return fireEvent( this, eventName, Context.forRactive( this, args.shift() ), args );
+		const proto = args.shift();
+		const ctx = Object.create( proto );
+		Object.assign( ctx, proto );
+		return fireEvent( this, eventName, ctx, args );
 	} else {
 		return fireEvent( this, eventName, Context.forRactive( this ), args );
 	}
--- a/src/Ractive/shared.js
+++ b/src/Ractive/shared.js
@@ -1,0 +1,1 @@
+export default {};
--- a/src/parse/_parse.js
+++ b/src/parse/_parse.js
@@ -1,4 +1,5 @@
 import { TEMPLATE_VERSION } from '../config/template';
+import { ELEMENT } from '../config/types';
 import Parser from './Parser';
 import readMustache from './converters/readMustache';
 import readTriple from './converters/mustache/readTriple';
@@ -69,6 +70,8 @@ const StandardParser = Parser.extend({
 		this.includeLinePositions = options.includeLinePositions;
 		this.textOnlyMode = options.textOnlyMode;
 		this.csp = options.csp;
+
+		this.transforms = options.transforms || options.parserTransforms || shared.defaults.parserTransforms;
 	},
 
 	postProcess ( result ) {
@@ -87,6 +90,42 @@ const StandardParser = Parser.extend({
 			const expr = {};
 			insertExpressions( result[0].t, expr );
 			if ( Object.keys( expr ).length ) result[0].e = expr;
+		}
+
+		const transforms = this.transforms;
+		if ( transforms && transforms.length ) {
+			const tlen = transforms.length;
+			const walk = function ( fragment ) {
+				let len = fragment.length;
+
+				for ( let i = 0; i < len; i++ ) {
+					let node = fragment[i];
+
+					if ( node.t === ELEMENT ) {
+						for ( let j = 0; j < tlen; j++ ) {
+							const res = transforms[j].call( shared.Ractive, node );
+							if ( !res ) {
+								continue;
+							} else if ( res.remove ) {
+								fragment.splice( i--, 1 );
+								len--;
+								break;
+							} else if ( res.replace ) {
+								if ( Array.isArray( res.replace ) ) {
+									fragment.splice( i--, 1, ...res.replace );
+									len += res.replace.length - 1;
+									break;
+								} else {
+									fragment[i] = node = res.replace;
+								}
+							}
+						}
+					}
+
+					if ( node.f ) walk( node.f );
+				}
+			};
+			walk( result[0].t );
 		}
 
 		return result[0];
--- a/src/parse/_parse.js
+++ b/src/parse/_parse.js
@@ -14,6 +14,7 @@ import readPartialDefinitionSection from './converters/readPartialDefinitionSect
 import readTemplate from './converters/readTemplate';
 import cleanup from './utils/cleanup';
 import insertExpressions from './utils/insertExpressions';
+import shared from '../Ractive/shared';
 
 // See https://github.com/ractivejs/template-spec for information
 // about the Ractive template specification
@@ -29,11 +30,11 @@ const defaultInterpolate = [ 'script', 'style', 'template' ];
 
 const StandardParser = Parser.extend({
 	init ( str, options ) {
-		const tripleDelimiters = options.tripleDelimiters || [ '{{{', '}}}' ];
-		const staticDelimiters = options.staticDelimiters || [ '[[', ']]' ];
-		const staticTripleDelimiters = options.staticTripleDelimiters || [ '[[[', ']]]' ];
+		const tripleDelimiters = options.tripleDelimiters || shared.defaults.tripleDelimiters;
+		const staticDelimiters = options.staticDelimiters || shared.defaults.staticDelimiters;
+		const staticTripleDelimiters = options.staticTripleDelimiters || shared.defaults.staticTripleDelimiters;
 
-		this.standardDelimiters = options.delimiters || [ '{{', '}}' ];
+		this.standardDelimiters = options.delimiters || shared.defaults.delimiters;
 
 		this.tags = [
 			{ isStatic: false, isTriple: false, open: this.standardDelimiters[0], close: this.standardDelimiters[1], readers: STANDARD_READERS },
@@ -42,14 +43,14 @@ const StandardParser = Parser.extend({
 			{ isStatic: true,  isTriple: true,  open: staticTripleDelimiters[0],  close: staticTripleDelimiters[1],  readers: TRIPLE_READERS }
 		];
 
-		this.contextLines = options.contextLines || 0;
+		this.contextLines = options.contextLines || shared.defaults.contextLines;
 
 		this.sortMustacheTags();
 
 		this.sectionDepth = 0;
 		this.elementStack = [];
 
-		this.interpolate = Object.create( options.interpolate || {} );
+		this.interpolate = Object.create( options.interpolate || shared.defaults.interpolate || {} );
 		this.interpolate.textarea = true;
 		defaultInterpolate.forEach( t => this.interpolate[ t ] = !options.interpolate || options.interpolate[ t ] !== false );
 
--- a/src/parse/converters/expressions/primary/readReference.js
+++ b/src/parse/converters/expressions/primary/readReference.js
@@ -9,7 +9,7 @@ const globals = /^(?:Array|console|Date|RegExp|decodeURIComponent|decodeURI|enco
 const keywords = /^(?:break|case|catch|continue|debugger|default|delete|do|else|finally|for|function|if|in|instanceof|new|return|switch|throw|try|typeof|var|void|while|with)$/;
 
 const prefixPattern = /^(?:\@\.|\@|~\/|(?:\^\^\/(?:\^\^\/)*(?:\.\.\/)*)|(?:\.\.\/)+|\.\/(?:\.\.\/)*|\.)/;
-const specials = /^(key|index|keypath|rootpath|this|global|shared|context|event|node)/;
+const specials = /^(key|index|keypath|rootpath|this|global|shared|context|event|node|data)/;
 
 export default function readReference ( parser ) {
 	let prefix, name, global, reference, lastDotIndex;
--- a/src/parse/converters/expressions/primary/readReference.js
+++ b/src/parse/converters/expressions/primary/readReference.js
@@ -9,7 +9,7 @@ const globals = /^(?:Array|console|Date|RegExp|decodeURIComponent|decodeURI|enco
 const keywords = /^(?:break|case|catch|continue|debugger|default|delete|do|else|finally|for|function|if|in|instanceof|new|return|switch|throw|try|typeof|var|void|while|with)$/;
 
 const prefixPattern = /^(?:\@\.|\@|~\/|(?:\^\^\/(?:\^\^\/)*(?:\.\.\/)*)|(?:\.\.\/)+|\.\/(?:\.\.\/)*|\.)/;
-const specials = /^(key|index|keypath|rootpath|this|global|shared|context|event|node|data)/;
+const specials = /^(key|index|keypath|rootpath|this|global|shared|context|event|node|local)/;
 
 export default function readReference ( parser ) {
 	let prefix, name, global, reference, lastDotIndex;
--- a/src/parse/utils/refineExpression.js
+++ b/src/parse/utils/refineExpression.js
@@ -10,7 +10,8 @@ export default function refineExpression ( expression, mustache ) {
 		}
 
 		if ( expression.t === REFERENCE ) {
-			if ( !~expression.n.indexOf( '@context' ) ) {
+			const n = expression.n;
+			if ( !~n.indexOf( '@context' ) ) {
 				mustache.r = expression.n;
 			} else {
 				mustache.x = flattenExpression( expression );
--- a/src/shared/Context.js
+++ b/src/shared/Context.js
@@ -7,6 +7,7 @@ import { update as protoUpdate } from '../Ractive/prototype/update';
 import runloop from '../global/runloop';
 import findElement from '../view/items/shared/findElement';
 import getRactiveContext, { extern } from './getRactiveContext';
+import Model from '../model/Model';
 
 const modelPush = makeArrayMethod( 'push' ).model;
 const modelPop = makeArrayMethod( 'pop' ).model;
@@ -16,11 +17,29 @@ const modelSort = makeArrayMethod( 'sort' ).model;
 const modelSplice = makeArrayMethod( 'splice' ).model;
 const modelReverse = makeArrayMethod( 'reverse' ).model;
 
+class ContextData extends Model {
+	constructor ( options ) {
+		super( null, null );
+
+		this.isRoot = true;
+		this.root = this;
+		this.value = {};
+		this.ractive = options.ractive;
+		this.adaptors = [];
+		this.context = options.context;
+	}
+
+	getKeypath () {
+		return '@context.data';
+	}
+}
+
 export default class Context {
 	constructor ( fragment, element ) {
 		this.fragment = fragment;
 		this.element = element || findElement( fragment );
 		this.ractive = fragment.ractive;
+		this.root = this;
 	}
 
 	get decorators () {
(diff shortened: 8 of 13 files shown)


evs-chris:
1. I think I like `@local` better.
2. Yep. Done.

paulocoghi:
`@local` is excellent!

ghost:
Post-processing templates looks neat! Although not everyone goes that low.

evs-chris:
3. I explored the `@state` a bit, and there's definitely some merit to it. I think it's not quite time for that now though, as there's a bit more depth to it than I'm comfortable sounding right now. So anyone looking for `@state`, feel free to leave comments about how you'd expect it to work. Here's where I left it:

---

1. Each model can potentially get its own state tree accessible by special key `@model`, because `@state` is ambiguous. This includes links.
2. Any keypath can contain a `@model` ref inline e.g. `foo.bar.@model.baz`, which would get the `baz` key in the state model attached to the `foo.bar` model. This also works with relative keypaths.
3. Computations would be introduced at any keypath with a new config property (registry) next to `computed`. When a state model is asked for a key that it doesn't already know about, it first checks with the config registry to see if a computed matches, and if it does, grafts it in. This only works on the root of the state model e.g. `foo.bar.boxes.0.@model.area`. Not the prettiest thing in the world, but also non-polluting and it doesn't nerf perf for plain models.
  a. In order for the computation to work, it has to take a parameter for its context.

evs-chris:
If anyone's curious what a more in-depth transform looks like, [here's](https://ractive.js.org/playground/#N4IgFiBcoE5SBTAJgcwSAvgGhAF3gDwDOAxjAJYAOuABLgJ6UIC8AOnggB64D0YuAWwA27GuSRsOAykICGuBOwB8rAHatcBErJhI65XEJbtgwAQiJFZaDBmVqNGgCphyRMe9k1tugHQPNHh8kFXVcAIJcWQAjIlDHTSjo-UNjEFMkgEYnAyNbezCEgAkEISEAewCNAh4k+PDEmJpo8lUkAFokNxijCXYAQgRVHoQnGIAmAoSNAGUEEnK2qsC65cimrqto3vqEl3JdNdqY3cbkolL5hRDljQRcEiPVwrPTjQBZCys0SBoCVsoAFdaAA3WRCQFpUzmSzWBD5EA0HhKAjRGDI264MbJTI0HKpX7-VRA0HgyGSDIxbK5eF2RHI1Hot6aOTRUoogHA7xgeYAa2QFOAQxG2PGCLojDSJB5JF50XKnFEyJoAFFhtsEHimuMAPw1Vnsp4nNbPaYKB6YxkYl4aACSNFkAjoPJoIIQMHoNAE5SQpRo5QAZg6vT73aovbIAFblGA0NCqd3gy2UJS2gDkbrEqgDMYE8nIi1dCDQSSMDtU5DzQnLegErUTQhqKeTqYzmtaOZgedwBfDbpLI3LlfBNa99Zg4KbzIIKfTmY7ufzhf790HsgrVdHdYTE8bPGbNs0s7bWc73d7RYHGqHm-XtfHk-30+P8+zi57y+Lq+v6+H1bvY47o+B5mjOravmeS59l+paar+t5tIBDZTi2c7tm+XZQZe35lvBI4AduyFPka0T1DUSRxBEKayAGCixhRKGqDUpAUNQ8RMV0IJiH0eA6CWyg1JxSggDgRDwDwPDeDoegAO7iPxqgAEqyCQPZur4voBrIgJCLgRC+JQOgXDATgTqoRBnvpQJEGAAAUAaAqoqm9rZqgAJQ0MAATkEGrm+JqzCBTQabBGmHleS8CzmbQUUKKotDMJ5dC-AA7FgNAIL8aacWm6UCL8ADawDJTQmQAMzpaoWUkHIli5TQAZZV27TBC1ixxbgaY0BgAC66WNTQqi+EGGAANzLFFRC0DACAyCpAVJbgqXpZlwU5XlhXFUtpUVYN1W1UQ9UDWmzWhd1fUNYVsVDLgPXdeNkWLFNKRlolQ1OgAZB9g2+AIw2tEgtleMwSgOr44aBYlaY9qkYUPQkPk0LZMNGOFmLXfFw2+I5Nk+bgtlbctGVZetXqbSV5WVftsh1f1TUCC10ntCjCBdb1dMvQgw3dW58MNHYLwzbggIwOGxUzXNJCamNAQCxgvNqBJGWIcENByag9wOKoStJGr8ma0pKlqVzmnabpVlGe6pnrhZuZWYCNn2Y5zmLK5aNhIjfkBUF0MxId7sJJNMXtTdNCJYTpUAJxUzQACsAAs6UAF7k1VwUAPrp126dJPVnC-OLWUAALXdwbPdelxVp2mmfZ0k6dIPIsh5wXNAwMXFTaEI5fs5dNAFXdMuPdFbezXIUth4tROrdl5AgvV+X9xHlN7cFNU04dHMnQzFFsxdA0Fd4IfxTQg98xoQec+44clWlxPBTpR2FWfE3H7gWM464tHIzSRAKzaw9noUUnkNAM-1dLulcmHUGQ1vZQ1zv-BIRhaAXCEOfXAFFhoxhVCpOyyN0rkA8iDTy6Mnq0BZpPd+n1vpUP+m0IG0CwYQx9izOGpCR6oKuMgShv0aBfToL9OhgNgag1kODMOPtOGqWQGww8l9NgjD0IlKhfCaGCIDADBhxCxHMKhgojUSBZFmnYUApoN9tp3xnkIcgT9OY0B1BKJggZObc0hsFKaFBVAoC6g4gqLNhp3V+P4oMhUAAG2IaAABJgDkBoAAalKhgUJF1F5FQprtau69LDtGakkdoUjri2LSdtcY6V85JXbv3GuWcBA5xiA3JuaYLpEF+KE9OAAGXwBTuFuOieQJJ3VzpDKHmaT2-R9G9ADg0BISRBHWTshHFK7SY4FRCtY2UTSOaFyqbXWpud0ppiLqoUMmyaAtJoKEgqHSun3Fsoczu4IblGGkYY9KfT5bpWuRcfG9zyhd18EYWi9V06ZF8IGAM3yAAyCBaJuU+Z075dyi4PKEL4dWuAwDAtBeC75AB1cQGK3I9VCYM+W6CNAYAykIC4JDDwaFmX9eZBN0kxzWRvHJO8YidG6AY2xEzkDcwcfypA3NW4lLKa3SpA90rnOhjASEPcebkoaJ7YVw0pnGIwTEOZDsFklUTqvWOmQObFNSss1eqyarkA2fvSVhVql1xiPVQ5xzfSnPOZcr5tzfn-O6a8qJMSPk0C9T85FfzHmAs6p87FAYIX3GhbC+FNzQ0orRQSzF0awWxrxemolJLKW90qWqgMBV2k9V8DACVNBi2lvLZwdKMByk1rLRWzgSrMQUsxALM0LN9JgKZUkRB0zwiHjGcK1RZzLgvInagiRiVHKaXrEgDys7ErkGVewjqWMmWLJWiTOetj34Bg2kvFlq82W0z7tvZmXKMadQrlPHarKslEA5Te6I+Sp2FJNRHUpNBynbNWbsup0QGlRHda0r1zzrhzoDf0-NwzB5DopchkdCRV2TurAAHywzQdp6C73bt1cy8V-67U7JqSBiDFyOkTpDUi1Nfr6rRNQfLCdaY0wIbJcsXtw0CqoPLYy4jizjXnr9e+tmqGeO-yI47Xd99Z7zxPaap956X22OvXkgG1r5AxjZpXM91cpr0CMEUtMkasoGdI5Uh1ez6mNyiACmFnVsDBUoJwUaeskAYss4+v9NngP1wc2I9FmKH1pnc6NJpwypOC3uCLMWo9JbSz5nLRBhtnLqV9AsCcuAYxEFWeJ3Od1EoOSch+VQbtaUXzId4XAbblGuH0iWAAcqGW0b43a+BLHcwLTqh2qU4MmgmywfURuc2mX4Q0cXxuc1gUbYb-mhcmz9GbuB8XebALLZDYQhYJeqw0BQ0lygyUq+FbtuAh5DyVkMPQut1YKUVpJb5NBASUGdJqX03o1BgljLGN6CAZI0GUplhAI2whRBgCWLKABiSHJZcoBAUNIOQChYfI7mgoRHYRgsF2WDCb4M99juAoW4csWY6KUHKKjmMSMImxzctjmZVJ8RGCyrgTIaZZZqDJSJEAJBxIACoIz8g+w1dqg1yi0AOBQKIPZWgoBoILngah+DCAOzmeK7QtJ1iEPQX4VhzKfooAGB63btbC9Vg9+4SuVdDVOtJA7VOiAGF7L8GaqO54ID5nmKHrRfidNjjNJ07TSrB7w3zaIKleQoBgOUBdbUKiVJh7G03AR5S6HdDk2QXQHa-HGJFpHZkXcVd+PKTg+SwA55O3h3w4xY5EEjwqSv1eZIB5oAXttne8NtxQFH2yyzB9YE6ZkeOcKe+ZHc3HKfoeof96H4P3wmRxjj9D2VKfk+23tG73P2QA-h8H86Svs3WsK0M1VmfxmuhmY0gO9YhM7QeTkBQPwX4ZVO8+74q0do8oMHlEXgAByF5hCa64D5LkBJyrTjDxzAEaCgHtAyQIDP6v40BlTtL4Y86n4O66A0CX6tR3pO455dBeL54wEebp4xi+gwDZ657nKh6h7d4f6YHqAW50BNDW60DK5qCX7AIRQaDO6u6LDu6lD5hugn4sE8F+w0Cgw6R35uBgHGZs6S4Jh8yGRIDEEoAB6qHlAl5u6jye5iEBDlBugwABgVAyTtDlLaR5Z8zGHuhmEnbtD640CuDqFDB8wySuAKD5KGRSxTYnYTiUBN4V42St7t6MEz697z6H7D5L5j7pSh6b7T5tqz59574L6xHL6r6oEb5T7b5T67774L517-zm6SGxDSGvbVigzWIHabBzTOGtD34IA-6dy8h8wKDcDtDgjP5pxSzxTuiqFEEK4B6+ApSxwIBOiZCTF8zNGP5IEv7bTTECCf5+5pwYFhDl7gFJwjHNCUFZ7l58wkAixEAxi-BU6tB0TiHcG5JSEyHVG4G3Efrjp8EPAnFnHVowpmy4C2G+EGDOGB7XH26coVHlHtDabaB5axivECGl4OixDU7Aje4UF-6LwbEaBP6LH56wG4BR6yix7x5tCJ4fGx4IDuFF42yCFpyRq17jDuAIA0wtHf7x64DpSha0n0mMngmqDtAslAnlHuBgkEEwk6FUm-B+yIkKAdHlCUBv6+ATErEBCha-CZDoEACkvxKk-xWhAQSc3Jvo5S7QmQfMmJKB6JGCxeYp-ofxDAtJDeGUXJzJwIyc+pXAeG7gzROgdpRA6UpptA7SHp9YXpR+De-JTxgpTxbUAxmMkZfqTuopcJHuohyJEOMp4pwI5QmpJA2ppUfMepAMXAKpJpCxKB1hmZFJ5kVpMpWptpIZnJFw3JvJzpNA+ZbQbpAZNAnpsYdZvpJZ-pgZCYwZjezBmAQAA) a demo playing around with copying materializecss cards and tabs.

If 4k links don't work for you, you can paste this into the content textarea on the settings tab:

N4IgFiBcoE5SBTAJgcwSAvgGhAF3gDwDOAxjAJYAOuABLgJ6UIC8AOnggB64D0YuAWwA27GuSRsOAykICGuBOwB8rAHatcBErJhI65XEJbtgwAQiJFZaDBmVqNGgCphyRMe9k1tugHQPNHh8kFXVcAIJcWQAjIlDHTSjo-UNjEFMkgEYnAyNbezCEgAkEISEAewCNAh4k+PDEmJpo8lUkAFokNxijCXYAQgRVHoQnGIAmAoSNAGUEEnK2qsC65cimrqto3vqEl3JdNdqY3cbkolL5hRDljQRcEiPVwrPTjQBZCys0SBoCVsoAFdaAA3WRCQFpUzmSzWBD5EA0HhKAjRGDI264MbJTI0HKpX7-VRA0HgyGSDIxbK5eF2RHI1Hot6aOTRUoogHA7xgeYAa2QFOAQxG2PGCLojDSJB5JF50XKnFEyJoAFFhtsEHimuMAPw1Vnsp4nNbPaYKB6YxkYl4aACSNFkAjoPJoIIQMHoNAE5SQpRo5QAZg6vT73aovbIAFblGA0NCqd3gy2UJS2gDkbrEqgDMYE8nIi1dCDQSSMDtU5DzQnLegErUTQhqKeTqYzmtaOZgedwBfDbpLI3LlfBNa99Zg4KbzIIKfTmY7ufzhf790HsgrVdHdYTE8bPGbNs0s7bWc73d7RYHGqHm-XtfHk-30+P8+zi57y+Lq+v6+H1bvY47o+B5mjOravmeS59l+paar+t5tIBDZTi2c7tm+XZQZe35lvBI4AduyFPka0T1DUSRxBEKayAGCixhRKGqDUpAUNQ8RMV0IJiH0eA6CWyg1JxSggDgRDwDwPDeDoegAO7iPxqgAEqyCQPZur4voBrIgJCLgRC+JQOgXDATgTqoRBnvpQJEGAAAUAaAqoqm9rZqgAJQ0MAATkEGrm+JqzCBTQabBGmHleS8CzmbQUUKKotDMJ5dC-AA7FgNAIL8aacWm6UCL8ADawDJTQmQAMzpaoWUkHIli5TQAZZV27TBC1ixxbgaY0BgAC66WNTQqi+EGGAANzLFFRC0DACAyCpAVJbgqXpZlwU5XlhXFUtpUVYN1W1UQ9UDWmzWhd1fUNYVsVDLgPXdeNkWLFNKRlolQ1OgAZB9g2+AIw2tEgtleMwSgOr44aBYlaY9qkYUPQkPk0LZMNGOFmLXfFw2+I5Nk+bgtlbctGVZetXqbSV5WVftsh1f1TUCC10ntCjCBdb1dMvQgw3dW58MNHYLwzbggIwOGxUzXNJCamNAQCxgvNqBJGWIcENByag9wOKoStJGr8ma0pKlqVzmnabpVlGe6pnrhZuZWYCNn2Y5zmLK5aNhIjfkBUF0MxId7sJJNMXtTdNCJYTpUAJxUzQACsAAs6UAF7k1VwUAPrp126dJPVnC-OLWUAALXdwbPdelxVp2mmfZ0k6dIPIsh5wXNAwMXFTaEI5fs5dNAFXdMuPdFbezXIUth4tROrdl5AgvV+X9xHlN7cFNU04dHMnQzFFsxdA0Fd4IfxTQg98xoQec+44clWlxPBTpR2FWfE3H7gWM464tHIzSRAKzaw9noUUnkNAM-1dLulcmHUGQ1vZQ1zv-BIRhaAXCEOfXAFFhoxhVCpOyyN0rkA8iDTy6Mnq0BZpPd+n1vpUP+m0IG0CwYQx9izOGpCR6oKuMgShv0aBfToL9OhgNgag1kODMOPtOGqWQGww8l9NgjD0IlKhfCaGCIDADBhxCxHMKhgojUSBZFmnYUApoN9tp3xnkIcgT9OY0B1BKJggZObc0hsFKaFBVAoC6g4gqLNhp3V+P4oMhUAAG2IaAABJgDkBoAAalKhgUJF1F5FQprtau69LDtGakkdoUjri2LSdtcY6V85JXbv3GuWcBA5xiA3JuaYLpEF+KE9OAAGXwBTuFuOieQJJ3VzpDKHmaT2-R9G9ADg0BISRBHWTshHFK7SY4FRCtY2UTSOaFyqbXWpud0ppiLqoUMmyaAtJoKEgqHSun3Fsoczu4IblGGkYY9KfT5bpWuRcfG9zyhd18EYWi9V06ZF8IGAM3yAAyCBaJuU+Z075dyi4PKEL4dWuAwDAtBeC75AB1cQGK3I9VCYM+W6CNAYAykIC4JDDwaFmX9eZBN0kxzWRvHJO8YidG6AY2xEzkDcwcfypA3NW4lLKa3SpA90rnOhjASEPcebkoaJ7YVw0pnGIwTEOZDsFklUTqvWOmQObFNSss1eqyarkA2fvSVhVql1xiPVQ5xzfSnPOZcr5tzfn-O6a8qJMSPk0C9T85FfzHmAs6p87FAYIX3GhbC+FNzQ0orRQSzF0awWxrxemolJLKW90qWqgMBV2k9V8DACVNBi2lvLZwdKMByk1rLRWzgSrMQUsxALM0LN9JgKZUkRB0zwiHjGcK1RZzLgvInagiRiVHKaXrEgDys7ErkGVewjqWMmWLJWiTOetj34Bg2kvFlq82W0z7tvZmXKMadQrlPHarKslEA5Te6I+Sp2FJNRHUpNBynbNWbsup0QGlRHda0r1zzrhzoDf0-NwzB5DopchkdCRV2TurAAHywzQdp6C73bt1cy8V-67U7JqSBiDFyOkTpDUi1Nfr6rRNQfLCdaY0wIbJcsXtw0CqoPLYy4jizjXnr9e+tmqGeO-yI47Xd99Z7zxPaap956X22OvXkgG1r5AxjZpXM91cpr0CMEUtMkasoGdI5Uh1ez6mNyiACmFnVsDBUoJwUaeskAYss4+v9NngP1wc2I9FmKH1pnc6NJpwypOC3uCLMWo9JbSz5nLRBhtnLqV9AsCcuAYxEFWeJ3Od1EoOSch+VQbtaUXzId4XAbblGuH0iWAAcqGW0b43a+BLHcwLTqh2qU4MmgmywfURuc2mX4Q0cXxuc1gUbYb-mhcmz9GbuB8XebALLZDYQhYJeqw0BQ0lygyUq+FbtuAh5DyVkMPQut1YKUVpJb5NBASUGdJqX03o1BgljLGN6CAZI0GUplhAI2whRBgCWLKABiSHJZcoBAUNIOQChYfI7mgoRHYRgsF2WDCb4M99juAoW4csWY6KUHKKjmMSMImxzctjmZVJ8RGCyrgTIaZZZqDJSJEAJBxIACoIz8g+w1dqg1yi0AOBQKIPZWgoBoILngah+DCAOzmeK7QtJ1iEPQX4VhzKfooAGB63btbC9Vg9+4SuVdDVOtJA7VOiAGF7L8GaqO54ID5nmKHrRfidNjjNJ07TSrB7w3zaIKleQoBgOUBdbUKiVJh7G03AR5S6HdDk2QXQHa-HGJFpHZkXcVd+PKTg+SwA55O3h3w4xY5EEjwqSv1eZIB5oAXttne8NtxQFH2yyzB9YE6ZkeOcKe+ZHc3HKfoeof96H4P3wmRxjj9D2VKfk+23tG73P2QA-h8H86Svs3WsK0M1VmfxmuhmY0gO9YhM7QeTkBQPwX4ZVO8+74q0do8oMHlEXgAByF5hCa64D5LkBJyrTjDxzAEaCgHtAyQIDP6v40BlTtL4Y86n4O66A0CX6tR3pO455dBeL54wEebp4xi+gwDZ657nKh6h7d4f6YHqAW50BNDW60DK5qCX7AIRQaDO6u6LDu6lD5hugn4sE8F+w0Cgw6R35uBgHGZs6S4Jh8yGRIDEEoAB6qHlAl5u6jye5iEBDlBugwABgVAyTtDlLaR5Z8zGHuhmEnbtD640CuDqFDB8wySuAKD5KGRSxTYnYTiUBN4V42St7t6MEz697z6H7D5L5j7pSh6b7T5tqz59574L6xHL6r6oEb5T7b5T67774L517-zm6SGxDSGvbVigzWIHabBzTOGtD34IA-6dy8h8wKDcDtDgjP5pxSzxTuiqFEEK4B6+ApSxwIBOiZCTF8zNGP5IEv7bTTECCf5+5pwYFhDl7gFJwjHNCUFZ7l58wkAixEAxi-BU6tB0TiHcG5JSEyHVG4G3Efrjp8EPAnFnHVowpmy4C2G+EGDOGB7XH26coVHlHtDabaB5axivECGl4OixDU7Aje4UF-6LwbEaBP6LH56wG4BR6yix7x5tCJ4fGx4IDuFF42yCFpyRq17jDuAIA0wtHf7x64DpSha0n0mMngmqDtAslAnlHuBgkEEwk6FUm-B+yIkKAdHlCUBv6+ATErEBCha-CZDoEACkvxKk-xWhAQSc3Jvo5S7QmQfMmJKB6JGCxeYp-ofxDAtJDeGUXJzJwIyc+pXAeG7gzROgdpRA6UpptA7SHp9YXpR+De-JTxgpTxbUAxmMkZfqTuopcJHuohyJEOMp4pwI5QmpJA2ppUfMepAMXAKpJpCxKB1hmZFJ5kVpMpWptpIZnJFw3JvJzpNA+ZbQbpAZNAnpsYdZvpJZ-pgZCYwZjezBmAQAA

PR #2890, opened 2017-02-21
Test changes to make backporting fixes easier

evs-chris:
This switches to using `import { test } from 'qunit'` in all of the tests so that backporting bugfixes is easier between 0.10, 0.9, and 0.8.

It also adds a runner for the node tests to the build scripts.

## Is breaking:
Nope.

### DIFF
--- a/gobblefile.js
+++ b/gobblefile.js
@@ -128,7 +128,7 @@ function buildNodeTests() {
 			format: 'cjs',
 			entry: 'tests.js',
 			dest: 'node.js',
-			external: ['qunitjs', 'cheerio']
+			external: ['cheerio']
 		}).moveTo('tests');
 }
 
--- a/tests/browser/aliases.js
+++ b/tests/browser/aliases.js
@@ -1,11 +1,12 @@
 import { initModule } from '../helpers/test-config';
+import { test } from 'qunit';
 
 export default function() {
 	initModule( 'aliases' );
 
 	/* global navigator */
 
-	QUnit.test( 'simple template aliases', t => {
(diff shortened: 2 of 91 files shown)


ghost:
BTW, the electron test is actually the node test. I just used electron to make node look like a browser to Karma since Karma doesn't run Node. That way, tests, reports, etc. (and soon, benchmarks) all run Karma in a consistent config. 😄 

Also made QUnit consistently global since you can't use module to mean QUnit.module consistently across tests, node in particular. Also didn't like using the other globals. Using the QUnit global explicitly sounds like a better practice.

Given that they're tested and written the same way, the added bonus is that they get built the same way. 😄

fskreuz:
Also made QUnit consistently global since you can't use `module` to mean `QUnit.module` consistently across tests, node in particular.

evs-chris:
Ah! I can't run electron in my usual dev environment because it has no X. I'll back that part out and verify that Rollup handles module correctly.

ghost:
If it's a Windows issue, it must be some magic sauce on my system. I'll check that out later. 😄

evs-chris:
Nah, it's because I usually work in vim in a tmux session over ssh.

Also, verified that rollup turns `import { module } from 'qunitjs'` safely into `qunitjs.module`.

ghost:
Aha! I think I ran into this problem in travis. This portion is what I added to make it work.

```
addons:
  apt:
    packages:
      - xvfb
install:
  - export DISPLAY=':99.0'
  - Xvfb :99 -screen 0 1024x768x24 > /dev/null 2>&1 &
```

Issue #2904, opened 2017-03-06
Remove src dependency of tests


ghost:
## Description:

Haven't looked into this deeply, but tests depend on some `src` modules to work. Few things of concern:

- Tests should be independent from the source, in theory.
- Some are internal APIs. If they break, the tests break.
- It causes tests to be recompiled when those things from `src` change.

![untitled](https://cloud.githubusercontent.com/assets/1582932/23593930/4f2dbbae-01e3-11e7-9b91-c563ae969239.png)

evs-chris:
Yeah, I've wanted to break the src dependency for a while, but not enough to throw in the effort. It certainly doesn't help test speed any, especially since those few deps end up pulling in 95% of the ractive source for the test output. It gets quite confusing when you search for say 'RepeatedFragment.prototype' in all.js by accident and don't notice that your breakpoints won't ever trip because you're in the wrong file.

PR #2912, opened 2017-03-18
Remove src from tests

ghost:

Removes direct `src` dependencies on tests.

https://github.com/ractivejs/ractive/issues/2904 . Should also speed up `src` change for `dev:browser` since altering `src` won't cause tests to recompile anymore.

## Is breaking:

No.

## Reviewers:

Everybody!


evs-chris:
Looking good! 🎉

evs-chris:
Was just looking at creating a sort of meta transition to run multiple transitions in sequence (possibly parallel, but that's trickier), and having `findInViewHierarchy` exposed as maybe `Ractive.findPlugin('transition', 'bounce', instance)` would be super useful.

evs-chris:
@fskreuz I completely forgot about this PR before I did the `getContext` switchover and created a bunch of conflicts 😞. Do you think there's any chance this could land before 0.9?

ghost:
No worries. I'll look into this later today. 😄

ghost:
Still contemplating whether or not I should expose those APIs above. While `normaliseKeypath` is a good candidate to expose since it goes well with the escape and unescape as utility functions, `parseJSON` and `findInViewHierarchy` feel too internal to me.

Still looking for a good way to test them from the public API.

evs-chris:
I've been thinking about dropping the `parseJSON` support for a while. It seems kinda surprising to me actually with the way it's used. If you pass and attribute like this `<div data-foo="{ bar: 42 }">`, which would normally be a string, ractive will try to parse it with `parseJSON` first. I kinda get why it was done originally, but it has some weird implementation details. It can also be replaced with `bind-data-foo={ bar: 42 }` as of 0.9.

I still like `findPlugin` as a front for `findInViewHierarchy` for library purposes - not really meant for general use, but nice for decorator and component authors.

evs-chris:
Pondering for a tiny bit more, I'd say go ahead and expose `parseJSON` for now to land this for 0.9. Since most people probably don't know about the whole bound attribute without an interpolator thing, we can leave it undocumented and pull it out before 1.0. Has the benefit of speeding up attributes slightly and dropping a little more code weight.

ghost:
I think this is good to go.

evs-chris:
The tests! They stand alone! 🎆 🍾

PR #2916, opened 2017-03-23
Ensure we bind functions using `Function.prototype.bind`

### LINKED ISSUE #1055: can't call underscore functions with ractive 0.5 (was working with 0.4)
I'm trying to call underscore functions from ractive, it was working with 0.4, not anymore with 0.5

ractive 0.5.x (exception thrown: Uncaught TypeError)
http://jsfiddle.net/ZBa4A/

ractive 0.4
http://jsfiddle.net/e43qL/

It seems like ractive is trying to call function.bind() but calls _.bind instead.

I don't know if i should blame underscore or ractive, but it should work in ractive, doesn't it? :)

### LINKED ISSUE #2915: Can't pass Underscore as part of template data in Ractive 0.8.12
Where `_` is Underscore, this produces a `TypeError`:

```js
new Ractive({
    el: 'container',
    template: '#template',
    data: { 
        underscore: _,
        object: {}
    }
});
```

JSFiddle: http://jsfiddle.net/pe5zgk8d/.

This is a regression from https://github.com/ractivejs/ractive/issues/1055. It broke sometime after the next version after 0.5, I'm not sure exactly when since the file structure has changed a lot.

Thankfully it is broken for the exact same reason as #1055. I'll put up a PR in a minute.

wearhere:

The function itself may have overridden `bind`, for instance if it
is Underscore: https://github.com/jashkenas/underscore/blob/8fc7032295d60aff3620ef85d4aa6549a55688a0/underscore.js#L765

Fixes #2915.

A similar fix was put in place in 6ce7082f67c963ec078c4e7b6f561c2cbbd5300c
to fix #1055, but that was lost somewhere along the way. This defeated the
regression test for #1055 because `src/utils/bind.js` only calls `bind` on
_functions_ that _used `this`_ and the regression test for #1055 simulated
Underscore using a POJO.

## Is breaking:

No (at least all previous tests continue to pass, this is my first PR against Ractive so there may be something I'm missing :).

### DIFF
--- a/src/utils/bind.js
+++ b/src/utils/bind.js
@@ -1,7 +1,9 @@
+const fnBind = Function.prototype.bind;
+
 export default function bind ( fn, context ) {
 	if ( !/this/.test( fn.toString() ) ) return fn;
 
-	const bound = fn.bind( context );
+	const bound = fnBind.call( fn, context );
 	for ( const prop in fn ) bound[ prop ] = fn[ prop ];
 
 	return bound;
--- a/tests/browser/misc.js
+++ b/tests/browser/misc.js
@@ -1062,6 +1062,37 @@ export default function() {
 		t.htmlEqual( fixture.innerHTML, 'FOO' );
 	});
 
+	test( 'Regression test for #2915', t => {
+		// This is how Underscore defines `_`:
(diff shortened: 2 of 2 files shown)


evs-chris:
Thanks! I'll cherry pick this over to 0.8 as well.

wearhere:
Cool, thanks for that + plus the quick turnaround!

PR #2943, opened 2017-04-23
Source map all the things

ghost:

Configure source maps to map all the way back to the source for both library and tests.

The only real hurdle was the module-skipping code. It needed to return a source map for the replacement code in order for the mapping to be a through and through.

## TODO:

- [x] Source maps for tests
- [x] Source maps for src.
- [x] Make the source map not print gobble paths
- [x] Make the tests not print gobble paths

Annoying debugging of code using the bundled version as reference.

## Is breaking:

No, i think.

## Reviewers:

Everyone

## Dependencies

Requires https://github.com/ractivejs/ractive/pull/2912 to go in first. It depends on several gobblefile changes there.


evs-chris (approved):
LGTM

ghost:
Source maps should now go all the way back to source. Hopefully this helps in debugging. 👍

evs-chris:
Does this cause gobble to loop around the hardlink step for anyone else? I haven't gotten too far into it, but it looks like maybe creating a hard link in linux also triggers an fsevent on the file?

ghost:
Looks like I missed checking the watch. PR inbound.

PR #2952, opened 2017-05-06
fixes #2945 - oncomplete won't fire if instance is torndown

### LINKED ISSUE #2945: oncomplete called after onteardown
## Description:
In this fiddle, http://jsfiddle.net/sabob/mygdmrzv/, the oncomplete lifecycle is called after onteardown. 

In my app I have a menu to switch between pages with a fading transition.
I generally use oncomplete to setup external libraries such as validations which need access to the form on the page. 

When switching pages I teardown the current page, then render the new page.

If I quickly switch away from a page that is still fading in, I get an error because the current page dom has been removed by the time oncomplete is called.

I expected the teardown call to jump to the end of the transition, call oncomplete, unrender the dom and call the rest of the lifecycle callbacks.

Is this a bug?

## Versions affected:
0.9, 0.8.x

sabob:
This PR adds a check to only fire oncomplete if the instance is not already torndown. Some of the oncomplete lifecycle tests failed where we  create a Ractive instance followed immediately by a teardown. Because the oncomplete hook is fired asynchronously the teardown call is made before oncomplete is invoked.

As a workaround I call the teardown in the next tick by wrapping it in a setTimeout,

#2945 

Is breaking:
Perhaps in tests, where creating a Ractive instance is immediately followed by a teardown call, and one expects the oncomplete hook to still fire.

Reviewers:
Yes

### DIFF
--- a/src/Ractive/render.js
+++ b/src/Ractive/render.js
@@ -47,5 +47,9 @@ export default function render ( ractive, target, anchor, occupants ) {
 	runloop.end();
 	ractive.rendering = false;
 
-	return promise.then( () => completeHook.fire( ractive ) );
+	return promise.then( () => {
+		if (ractive.torndown) return;
+
+		completeHook.fire( ractive );
+	});
 }
--- a/tests/browser/init/hooks/misc.js
+++ b/tests/browser/init/hooks/misc.js
@@ -169,6 +169,27 @@ export default function() {
 		});
 	}
 
+	test( 'oncomplete should not be fired if teardown is invoked while the rendering phase is still in progress (#2945)', t => {
+		t.expect( 1 );
+
+		const done = t.async();
+
+		const ractive = new Ractive({
+			template: '<p>foo</p>',
+
+			oncomplete () {
+				t.ok( false );
+			}
+		});
+
+		ractive.render( fixture );
+
+		ractive.teardown().then ( () => {
+			t.ok( true );
+			done();
+		} );
+	});
+
 	test( 'hooks include the source ractive instance as the last argument', t => {
 		const done = t.async();
 		t.expect( 3 );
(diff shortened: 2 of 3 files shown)


src/Ractive/render.js
```diff
@@ -47,5 +47,9 @@ export default function render ( ractive, target, anchor, occupants ) {
 	runloop.end();
 	ractive.rendering = false;
 
-	return promise.then( () => completeHook.fire( ractive ) );
+	return promise.then( () => {
+		if (!ractive.torndown) {
+			completeHook.fire( ractive );
+		}
```
ghost: Purely preference, but one less indent 😄 
```
if( ractive.torndown ) return;
completeHook.fire( ractive );
```

ghost:
Looks good to me 👍 Might need to squash the commits tho.

evs-chris:
Looks good to me too. I'm :+1: with a squash and forced push.

sabob:
err, squash is something I need to do from my side? I've no idea what that is...

evs-chris:
Basically reset to `origin/dev` (leaving all of your changed files sitting around), commit them again, and force push. I _think_ this will work, though I've never tried resetting to a remote branch...

```sh
git remote add ractive git@github.com:ractivejs/ractive
git fetch ractive dev
git reset ractive/dev
git status
# do the files look right?
git add .
git commit -m "Add a check not to fire oncomplete if ractive instance is torndown - fixes #2945"
git push -f origin dev
```

evs-chris:
Forgot to say what that should do. That should put all of your changes into one commit and then overwrite this PR's change set with that single commit. It's mostly to keep the main commit log a little cleaner, as we use it for changelogging :smile:

sabob:
Thanks @evs-chris, managed to run those commands. Should be ok now.

evs-chris:
:+1: Thanks!

PR #2956, opened 2017-05-11
Add windows build instruction. Fix #2951

### LINKED ISSUE #2951: Build Ractive on Windows
## Description:

Ractive's build experience on Windows (at least Windows 10) seems to have improved over the years. Following the Contributing.md page, I can do most of the npm commands in package.json without downloading the dependencies on that page such as, a "real shell" and Python.

Then the page notes the command: 

> npm run build:local

 which depends on a real script environment for Windows. 

However, as far as I can tell, the two most important commands for hacking Ractive is:

>npm run dev:browser

 for development (is this the command  everyone uses to hack on Ractive? It runs the tests at location:4567/tests) and

> npm run bundle:release

 for a release.

Both ran without any issue or external dependencies.

Perhaps contributing.md should put less emphasis on the external dependencies which raises the barrier to entry. It seems unnecessary for the primary use cases which is dev and build.

Instead we can highlight the above mentioned commands to develop and build Ractive.

On the other hand, it is quite possible I could be missing something.

## Versions affected:
edge

## Platforms affected:
Windows 10

sabob:
Add Windows instructions to CONTRIBUTING.MD

#2951

## Is breaking:
No

## Reviewers:
Yep


ghost:
So normally, you'll only need to run the `build:local` ones locally. Curious why you need to run `bundle:release` for Windows. You normally only do that to verify how the bundle looks like for release.

Here's the breakdown of the scripts.

- `bundle:test` - Bundles only Ractive and tests. Used by `build:*` since it makes no sense bundling everything when we're just testing.
- `bundle:release` - Bundles Ractive for publishing (lib, bin, polyfill, manifests). The bundle emitted is what gets sent to NPM.
- `build:ci:pr` - Travis runs this one on PRs (lint, tests, coverage).
- `build:ci:dev` - Travis runs this one on dev commits and after merges. It's `build:ci:pr` with `bundle:release` in preparation for publishing. Intentionally separate from `build:ci:pr` to avoid the possibility of publishing on submitted PRs.
- `build:local` - Similar to `build:ci:pr` but doesn't run Sauce Labs tests nor report to coveralls. Meant for local build, the one you want to run locally before submitting your PR.
- `build:local:fast` - A trimmed down version of `build:local`. Skips linting and node tests for fast iterations.

sabob:
> Curious why you need to run bundle:release for Windows. 

bundle:release builds a Ractive release on Windows without a dependency on external shells, gitbash and what not. I don't see a reason why shells should be needed to build Ractive. It raises the barrier to entry unnecessarily.

ghost:
The `build:*` scripts are just shorthands for multiple operations which are stored in shell scripts. Other than that, there's actually no need for a shell, everything's just an npm script. For instance, `build:local` is just the following npm scripts, which you can run manually without the need of a shell.

```
npm run lint:src
npm run lint:tests:browser
npm run lint:tests:node
npm run bundle:test
npm run test:phantom
npm run test:electron || true # Allow to fail for now
npm run test:coverage
```


Also, `bundle:release` just bundles Ractive up. It doesn't run any sort of testing.

I often do `npm run bundle:test && npm run test:phantom` myself. I guess this is a better option than running `bundle:*`.

sabob:
Agreed, we just need to highlight the commands on the contributing.md page because Windows users who want to contribute need to look at shell scripts to figure out which commands to run.

Also, any command that uses lint fails on Windows with:
 Expected linebreaks to be 'LF' but found 'CRLF'  linebreak-style

This line in .eslintrc is the cause:
"linebreak-style": [ "error", "unix" ],

Is that line needed?

ghost:
> "linebreak-style": [ "error", "unix" ],

This is to [enforce consistent line endings](http://stackoverflow.com/a/170968/575527). The expectation is that your editor uses LF and Git keeps it LF. If you're encountering line ending errors on lint, then you probably either:

- Use a text editor that doesn't understand LF (i.e. Notepad). 
- Use Git on Windows, which is usually pre-configured to convert LF -> CRLF on checkout.

Windows understands unix line endings just fine. It's just that some editors (again, notepad) don't know LF (saves as CRLF anyways), cannot represent LF properly (text ends up as a single line of text), or know but have their newline settings set to CRLF by default because they're on Windows. Git on Windows by default also converts LF to CRLF on checkout and vice versa on check-in, which is only useful if you're using an editor that doesn't understand unix line endings.

The root of the problem is just the editor. It's suggested that you use a better text editor - one that knows LF or one with [editorconfig](http://editorconfig.org/) support. [We have an editorconfig file](https://github.com/ractivejs/ractive/blob/dev/.editorconfig) in place that enforces LF. Supporting text editors will automatically detect this file and use LF. With that, there won't be a need to have Git do auto conversion. The following will tell Git to checkout as is, and yell if it sees CRLF:

    git config --global core.autocrlf input

sabob:
@fskreuz thanks it was git. This command did fix it
git config --global core.autocrlf input

We should also mention that in the contribution.md for Windows users.

PR #2971, opened 2017-05-23
Fix infinite hardlink on linux

ghost:

The hardlink gobble transform is causing infinite watch cycle on linux. This reverts it back to the copy version.

https://github.com/ractivejs/ractive/pull/2943#issuecomment-303273645

## Is breaking:

Nope

## Reviewers:

@evs-chris

### DIFF
--- a/gobblefile.js
+++ b/gobblefile.js
@@ -109,7 +109,7 @@ function buildBrowserTests() {
 		browserTests,
 		browserTests.transform(buildTestEntryPoint, { dir: 'browser' })
 	])
-		.transform('hardlink')
+		.transform(copy)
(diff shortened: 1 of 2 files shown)


evs-chris:
With this, I'm not getting the loop, but I also am not seeing the tests, which is really weird, cause they're clearly included. I think I'm missing something...

evs-chris:
Ah! Moved to `qunit`?

ghost:
Yep. Moved it so that sourcemap `tests` doesn't merge with the served `tests`.

evs-chris:
Gotcha. LGTM 😄

Issue #2988, opened 2017-05-31
0.9 - copy data object, wrong values 


kouts:
Here is the link in the playground: https://goo.gl/zLGf4F
When I'm copying the data object with ```JSON.parse(JSON.stringify(this.get()));```
I get the previous selected values instead of the current values.

Change a dropdown select and then check the console.
This does not happen with 0.8.
Any ideas?

ghost:
I vaguely remember having a discussion about this before, whether or not to update something before or after the change event. Lemme see if I can find it.

ghost:
Aha! Found it! https://github.com/ractivejs/ractive/issues/1371 Not sure if this was reintroduced recently. Worth debugging tho! 😄

kouts:
If data isn't copied it works as expected though, check it here: https://goo.gl/TiBCHE
Not sure if this is exactly like #1371

evs-chris:
This is caused by twoway bindings not being delegation-aware. I'll see if I can extract an internal listening interface so that they are, which will have the bonus of allowing custom events and decorators to use delegation as well. In the meantime, you can throw a `no-delegation` directive on the `div.row` element to prevent it from being a delegation target or just turn off delegation entirely with `Ractive.defaults.delegate = false` if you don't have any iterations with more than several events.

kouts:
But why does it work with 0.8 https://goo.gl/BXOYPa ?
I'm not sure what delegation-aware is :blush: would putting ```Ractive.defaults.delegate = false```
revert to the 0.8 behavior?

ghost:
It's an optimization feature that was added just recently. It's similar to `jQuery.on`, where the handler is placed on an ancestor instead of the element itself. Helps with dynamic elements, and avoids adding repeated handlers when you can let bubbling do its thing.

- https://github.com/ractivejs/ractive/pull/2841
- https://github.com/ractivejs/ractive/pull/2846

By default, [it's enabled](https://ractive.js.org/api/initialization-options/#delegate). As with any initialization option, modifying `Ractive.defaults` will globally change the setting. `Ractive.defaults.delegate = false` globally disables it.

kouts:
Ahh I see! Ok so Ractive tries to delegate the events "automagically", nice!
Hope this gets fixed so we can retain the performance benefits added by this feature.
Thank you @fskreuz  and @evs-chris

evs-chris:
[...]

PR #2992, opened 2017-06-02
Adjust internal event handling to allow delegation for any event including bindings - #2988

evs-chris:

This moves dom event handling to the element vdom item, so that delegation can be used with any subscribed events.including twoway bindings. Handlers are collected into a map on the element and only a single listener is installed for each event type, and delegation is handled as a special reference counted handler on the delegation target.

#2988

## Is breaking:
Maybe.

## Reviewers:
Yep

### DIFF
--- a/src/shared/Context.js
+++ b/src/shared/Context.js
@@ -89,6 +89,14 @@ export default class Context {
 		return promise;
 	}
 
+	listen ( event, handler ) {
+		const el = this.element;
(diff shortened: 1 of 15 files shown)


evs-chris:
This has pretty decent test coverage, and I'm _pretty_ sure it won't break anything. I'mma kick it out on edge for a bit to make sure though, before I move it to 0.9.

Issue #2986, opened 2017-05-31
Style-attributed set to outdated value after transition


LeJared:
## Description:
When an element is transitioned-in (e.g. fade-in) any style-*-directive or style-attribute which was updated after the transition had started, will be reset to its initial value it had right before the transition began.

## Versions affected:
0.8.14

## Reproduction:
https://jsfiddle.net/lejared/v7m4jxwb/
Click on "Toggle Visible" to start fade-in animation
Click on "Change Color" before animation ends

The Color will change instantly but will be reset to its initial value after the transition has completed.

Note that the colors text-value in the box is not affected.

evs-chris:
The issue is that transitions take a snapshot of the `style` property when they start and then restore it when they finish. Anything that happens in the middle is not accounted for. We can address this by tracking individual property values on transitions and only restoring those that have a value that we expect, but there's a considerable amount of cruft in the transition code that makes it a bit painful.

LeJared:
Naive approach without knowing anything about the internals: Wouldn't it be possible to just reapply the current model on the style-attribute or re-evaluate any style-* expression after the transition has completed?

PR #2995, opened 2017-06-03
Keep transition complete from wiping out external styles - #2986

evs-chris:

When a transition starts, it records the value of the `style` attribute, and when it completes, it sets the `style` attribute back to its original value. The problem with that is that if something else changes the value in the interim, it gets wiped out.

This adds tracking to individual style keys as they are set from the transition. The initial and target values are recorded, and then upon completion, if the current value matches the target value, it will be reset to the initial value.

There are still some corner cases that can't really be helped here:

* If you happen to set the style to the target value, it will get wiped out. There's not really a way that I know of to detect that.
* I think there may be some async race conditions that could arise in certain _very_ tight timing situations, but generally, it should work out.

#2986

## Is breaking:

Maybe, but probably not. Transitions are a finicky construct.

## Reviewers:
Yep.


evs-chris:
Going to test this a bit on edge for a little while and see if any problems shake out.

PR #2997, opened 2017-06-05
Use prettier for formatting - RFC

evs-chris:
This adds a `prettier` script entry in `package.json` and adds a call to it to the `build:local` script.

Ractive has a slightly unusual formatting standard for 2017, and it has been a bit of an issue for contributors in the past. Rich recently switched svelte to prettier at sveltejs/svelte#618, so I figure it's probably time for ractive to get automatic formatting too.


ghost:
Nice! Was thinking of using adding [Standard](https://standardjs.com/) to replace eslint, get rid of the configs, and just adhere to their rules instead. Haven't got to it yet tho.

tommarchi:
@fskreuz I absolutely dislike Standards configuration. Eslint + Airbnb config has become my new favorite. With small tweaks for personal use. There are a few issues I find with standard. A way to sum up standard in my eyes is it's a formater for lazy devs. Eslint + AirBNB guide is closer to clean and highly readable code.

PR #3000, opened 2017-06-07
Add an eval option that can disable parsing of expressions

evs-chris:
This adds an `eval` option for instances and parsing that when set to `false` causes the parser to skip any expression parsing functions and just use the reference parsing function. The end effect is that any non-reference mustaches throw a parsing error when `eval: false`.

Per comments in ractivejs/ractivejs.github.io#68, if you have untrusted content in your templates and you're using SSR, there's not really a way to completely ensure that terrible things wouldn't happen to the server when rendering. With expressions turned off, you can at least be sure that the template has no access to the environment.

### Question
Is `eval` the best name for this option? `allowExpressions` seems a bit more descriptive of what will actually happen. For instance, you could pre-parse a template that allowed expressions, and they would still execute in an instance with `eval: false`.

Security. Please, think of the children.

## Is breaking:
Only if you use an `eval` property on your instances.

## Reviewers:
@MartinKolarik @fskreuz

### DIFF
--- a/src/Ractive/config/defaults.js
+++ b/src/Ractive/config/defaults.js
@@ -8,6 +8,7 @@ export default {
 	template:               null,
 
 	// parse:
+	allowExpressions:       true,
 	delimiters:             [ '{{', '}}' ],
 	tripleDelimiters:       [ '{{{', '}}}' ],
 	staticDelimiters:       [ '[[', ']]' ],
--- a/src/Ractive/config/runtime-parser.js
+++ b/src/Ractive/config/runtime-parser.js
@@ -14,7 +14,9 @@ const parseOptions = [
 	'preserveWhitespace',
 	'sanitize',
 	'stripComments',
-	'contextLines'
+	'contextLines',
+	'parserTransforms',
+	'allowExpressions'
 ];
 
 const TEMPLATE_INSTRUCTIONS = `Either preparse or use a ractive runtime source that includes the parser. `;
--- a/src/parse/_parse.js
+++ b/src/parse/_parse.js
@@ -70,6 +70,7 @@ const StandardParser = Parser.extend({
 		this.includeLinePositions = options.includeLinePositions;
 		this.textOnlyMode = options.textOnlyMode;
 		this.csp = options.csp;
+		this.allowExpressions = options.allowExpressions;
 
 		this.transforms = options.transforms || options.parserTransforms;
 		if ( this.transforms ) {
--- a/src/parse/converters/readExpression.js
+++ b/src/parse/converters/readExpression.js
@@ -1,6 +1,14 @@
 import readConditional from './expressions/readConditional';
+import readReference from './expressions/primary/readReference';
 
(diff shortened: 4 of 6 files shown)


MartinKolarik:
>For instance, you could pre-parse a template that allowed expressions, and they would still execute in an instance with eval: false.

That's bad IMHO. The option should be also checked in runtime.

ghost:
Pull request # 3000. Dun dun duuun! 😄

ghost:
I'd go with the `allowExpressions` name, and `true` by default. `false` would:

- Have the parser ignore expressions/omit them from the parsed result.
- Disable all forms of string evaluation on runtime, like computed props.
    - Log warning during initialization if Ractive spots something that causes string evaluation.
    - References to expressions would render nothing or compute to `undefined`.

Just throwing in ideas. May be missing a few marks.

evs-chris:
I've updated the PR:

* `eval` is now `allowExpressions`
* `allowExpressions: false` on an instance will cause the expression proxy to use `noop` rather than getting the actual expression fn to evaluate - effectively disabling runtime evaluation of expressions

So if you want to pre-parse certain trusted templates with expressions supported, you can and still have others pre-parsed without expressions as long as the instance `allowExpressions !== false`. If you trust _no_ templates, then `allowExpressions: false` on the instance has you covered. (right?)

I think computeds are probably safe _because_ if the malicious entity can access your ractive instance to modify your computeds, it can already do pretty much whatever anyway, no?

Should there also be a runtime option to disable the runtime parser on a particular instance? It wouldn't be hard to add and would allow you to force a particular instance into using pre-parsed templates. Seems like it might go well with anchored instances. (`runtimeParser: false`).

MartinKolarik:
👍 

>Should there also be a runtime option to disable the runtime parser on a particular instance?

I don't see how that would be useful.

ghost:
Looks good. 👍 

Thinking about it, it looks like the option is better as a global config, a full lockdown of the library (i.e. `Ractive.allowExpressions = false`). I haven't tried Ractive on a multi-tenant setup though. Not sure if locking down one subtree and not the other is a thing.

MartinKolarik:
You can just use `Ractive.defaults.allowExpressions = false` if you want it to be global.

ghost:
Reading on CSP further, it looks like CSP is a page-wide security feature. Doesn't seem to make sense to have a per-instance config. Why would we allow an  instance to use expressions on a page that doesn't allow it? Or have one instance forcefully not be allowed to use expressions while on a page that doesn't CSP?

Is there a use-case I'm not aware of? 🤔

MartinKolarik:
@fskreuz this is about scenarios discussed in https://github.com/ractivejs/ractivejs.github.io/pull/68 (SSR and/or *parts* of templates being from untrusted sources) and not strictly related to CSP.

ghost:
Ahh, gotcha. 👍

MartinKolarik (approved):

Issue #3026, opened 2017-06-30
Promise error on IE 11


dagnelies:
Hi,

Currently, ractive latest/edge breaks on IE 11 due to missing `Promise` object.

https://caniuse.com/#search=Promise

Probably introduced through #2801

I'm amazed this did not pop up earlier. 😲

kouts:
Yes, you can solve this by adding the polyfill file and after loading Ractive insert
```Ractive.Promise = Promise;```

My opinion though is to support at least IE11 without polyfills, just my 2c.

ghost:
0.9 doesn't ship with a legacy build (one with built-in polyfills). Instead, it ships the polyfills on a separate file. Load it before Ractive for legacy browser support. Otherwise, if you only intend to support newer browsers or already have another polyfill in place (i.e. ES5 shim, ES6 shim, etc.), you can just use Ractive as is.

If plugins rely on `Ractive.Promise` (which was also removed), you can just stick the global promise (polyfilled or not) to as `Ractive.Promise` as @kouts mentioned.

dagnelies:
Considering IE 11 to be "legacy" is a bit harsh, don't you think?

ghost:
I just like to call IE legacy 😛 

But IE has [ceased feature development](https://en.wikipedia.org/wiki/Internet_Explorer#End_of_life). Pinning a library to a browser that's half-dead would only make it hard to move forward. IE11 is 99% ES5 compliant and for ES6, it will only need 3 APIs polyfilled - which isn't that bad really.

Technically, the polyfill + Ractive = legacy build, sort of. The only difference is that the polyfills are moved out of the library and into a separate file. It also makes it easy to use another polyfill. I'm not a fan of packaging own polyfills when other polyfills are better tested.

sabob:
I think the argument is if everyone is going to include polyfill.js anyway to support IE11, then maybe Ractive should just include it by default? Batteries is included...

MartinKolarik:
I have to agree with @sabob on this. IE 11 should be supported by default.

dagnelies:
Me too, I don't care about IE <= 10, but IE 11 is still widespread and should be supported for a couple of years IMHO.

In debug mode, a check could be performed when starting and mention something like "Error: this legacy browser is not supported, please use polyfills at `<some-url>`.

evs-chris:
This should now be fixed on edge, and will also be addressed for 0.9.3 when it is cut.

PR #3027, opened 2017-07-01
ES6 APIs pre-polyfilled

ghost:
## Description:

- `Promise`, `array.find` and `Object.assign` now pre-polyfilled on main build.
- `polyfills.js` still being shipped for older, pre-ES5 browsers.

Many, mostly comments on various post-removal issues.

## Is breaking:

Shouldn't be.

## Reviewers

@dagnelies @MartinKolarik @sabob @evs-chris and others who I cannot tag because Github.


MartinKolarik:
IMHO the point of #3026 was to have only polyfills needed for IE 11 directly in the regular build, which was the case before #2801. I personally don't care about older IE versions, which are officially marked as unsupported by MS, and don't want to load polyfills for ES5 features (e.g. `Array#forEach` and friends), but I think ES6+ polyfills should be included by default for some time.

For all polyfills you can just use https://cdn.jsdelivr.net/combine/npm/ractive/polyfills.min.js,npm/ractive/ractive.min.js, which is also linked in the docs.

dagnelies:
Thanks!

dagnelies:
Ah, I was a bit quick in my answer since I was in a hurry. I completely agree with what @MartinKolarik said, IE 11 should be supported in the default for some time. The adequate polyfills for it should be in the standart release.

Btw, IE 11 is not EOL as stated in #2801. It's older IE < 11 that are EOL since 2016.

ghost:
Updated PR. Build updated with `Promise`, `Object.assign` and `array.find` polyfilled out of the box. Ractive still requires at least ES5 to work, but the polyfills file is still there should there be a dire need to support really old platforms. 🎉  

Aaand re-removed the all-in-one legacy.js. Unless someone really wants a fully-polyfilled build, lemme know. 😄 

---

Totally unrelated aside: I read that WinXP usage increased 1.20-something percent over the weekend. Long live IE8! :trollface:

evs-chris:
nice!

PR #3034, opened 2017-07-06
handle fire with context as plain object - fixes #3033

### LINKED ISSUE #3033: Manually fired events don't seem to be able to extend Event's context
## Description:

From [the docs](https://ractive.js.org/api/#ractivefire):
> [context] (context|object): A context object to reuse for the event or an object with properties to assign to a new context object. **If you need to pass arguments but don't need to provide context, pass an empty object ({}) before the additional arguments.**

However, extending event's context doesn't seem to work.
See example below.

## Versions affected:
0.9.2

## Platforms affected:
reproduced on Chrome & Safari (macOS)

## Reproduction:
https://jsfiddle.net/4ervr8kv/3/

```javascript
const r = window.r = new Ractive({
  el: '#main',
  template: '#template',
  data: {
  },

  oninit() {
    this.on('myEvent', (...args) => {
      console.info(`Number of arguments: ${args.length}`, args);
    });

    // this.on('myEvent'...) should have its context extended by { wontExtendContext: 'this will be argument 1' }, no other argument.
    // will actually have a sort-of clear Context, and an additional argument: { wontExtendContext: 'this will be argument 1' }
    // SHOULD RESULT IN 1 ARGUMENT, currently 2
    this.fire('myEvent', { wontExtendContext: 'this will be argument 1' });

    // this.on('myEvent'...) should have its context extended by {} and an additional argument
    // will actually have a sort-of clear Context, and 2 additional arguments: {}, { someData: 'something' }
    // SHOULD RESULT IN 2 ARGUMENTS, currently 3
    this.fire('myEvent', {}, { someData: 'something' });
  }
});
```

marcalexiei:
## Description:
[Fire docs](https://ractive.js.org/api/#ractivefire) says that if context param is an object it should be used as a new context. Or if is an empty object no context param should be passed to event handler.

#3033

## Is breaking:
No.

## Additional information
I've modified "method-calls" test on events test suite. I've also added another test which controls that context can be changed in another object.

### DIFF
--- a/src/Ractive/prototype/fire.js
+++ b/src/Ractive/prototype/fire.js
@@ -8,6 +8,16 @@ export default function Ractive$fire ( eventName, ...args ) {
 		const ctx = Object.create( proto );
 		Object.assign( ctx, proto );
 		return fireEvent( this, eventName, ctx, args );
+	} else if ( args[0] && typeof args[0] === 'object' && args[0].constructor === Object ) {
+		if ( Object.keys(args[0]).length ) { // if first param is a plain object use it as context object
+			const proto = args.shift();
+			const ctx = Object.create( proto );
+			Object.assign( ctx, proto );
+			return fireEvent( this, eventName, ctx, args );
+		} else { // if object has no key that means that event handler will be called without scope
+			args.shift();
+			return fireEvent( this, eventName, null, args );
+		}
 	} else {
 		return fireEvent( this, eventName, Context.forRactive( this ), args );
 	}
--- a/tests/browser/events/method-calls.js
+++ b/tests/browser/events/method-calls.js
@@ -152,7 +152,7 @@ export default function() {
 
 		const component = ractive.findComponent( 'Component' );
 		fire( component.find( '#test' ), 'click' );
-		component.fire( 'bar', 'bar' );
+		component.fire( 'bar', {}, 'bar' );
 	});
 
 	test( 'component "on-" with ...arguments', t => {
@@ -307,5 +307,27 @@ export default function() {
 
 		fire( ractive.find( '#return_zero' ), 'click' );
 		t.ok( !preventedDefault && !stoppedPropagation );
+	});
+
+	test( 'Extend a context event', t => {
+		let customContext = false;
+
+		const ractive = new Ractive({
+			el: fixture,
+			template: `<button on-click='@this.fire("customContextEvent", { customContext: true }, 1)'>{{foo}}</button>`,
+			data: { foo: 0 },
+			on: {
+				customContextEvent: (context, value) => {
+					ractive.set({ foo: value });
+					if (context && context.customContext === true) {
+						customContext = true;
+					}
+				},
+			}
+		});
+
+		fire( ractive.find( 'button' ), 'click' );
+		t.equal( ractive.get( 'foo' ), 1 );
+		t.ok(customContext);
 	});
 }
--- a/src/events/fireEvent.js
+++ b/src/events/fireEvent.js
@@ -47,8 +47,10 @@ function variants ( name, initial ) {
 export default function fireEvent ( ractive, eventName, context, args = [] ) {
 	if ( !eventName ) { return; }
 
-	context.name = eventName;
-	args.unshift( context );
+	if (context) {
+		context.name = eventName;
+		args.unshift( context );
+	}
 
 	const eventNames = ractive._nsSubs ? variants( eventName, true ) : [ '*', eventName ];
 


tests/browser/events/method-calls.js
```diff
@@ -308,4 +308,26 @@ export default function() {
...
+						customContext = true;
+					}
+				},
+			}
+		});
+
+		fire( ractive.find( 'button' ), 'click' );
+		t.equal( ractive.get( 'foo' ), 1 );
+		t.ok(customContext);
+	});
```
ghost, line 332: Might also want to test how this works when the source is a proxy and the event is re-fired in a component, when the re-fire is from the template or from code, and propagation (fire from very bottom and second level, receive at top level via template `on-*` or code `on`).

The following is done in a rush and may not work, but you probably get the idea.

```
const Component = Ractive.extend({
  template: `
    <button on-click="myevent">Click Me!</button>
  `
});

const AnotherComponent = Ractive.extend({
  components: { Component },
  template: `
    <Component on-myevent="@this.fire('yetanotherevent', { custom: 'context' }, 'arg1')" />
    <Component on-myevent="@this.fire('yetanotherevent', {}, 'noContext')" />
    <Component on-myevent="anotherevent" />
  `,
  on: {
    anotherevent: function(context){
      this.fire('yetanotherevent', { custom: 'context' }, 'arg1');
      this.fire('yetanotherevent', {}, 'arg1');
    }
  }
})

const App = Ractive.extend({
  components: { AnotherComponent },
  template: `
    <AnotherComponent on-yetanotherevent="receiver" on-myevent="receiver" />
  `,
  on: {
    receiver: function(context){},
    '*.yetanotherevent': function(context){},
    '*.myevent' : function(context){},
  }
});
```

ghost:
Might also want to check with others. This _might_ be a docs issue and the feature might have been scrapped without docs update. I could be wrong, I never used `ractive.fire()` this way.

src/Ractive/prototype/fire.js
```diff
@@ -8,6 +8,16 @@ export default function Ractive$fire ( eventName, ...args ) {
 		const ctx = Object.create( proto );
 		Object.assign( ctx, proto );
 		return fireEvent( this, eventName, ctx, args );
+	} else if ( args[0] && typeof args[0] === 'object' && args[0].constructor === Object ) {
```
ghost, line 11: There's an `isObject` in the utils https://github.com/ractivejs/ractive/blob/dev/src/utils/is.js#L25

marcalexiei:
Since this could be a doc issue I'll wait further review feedback before edit code.

Thanks for the additional info about object checks and test!

evs-chris:
FWIW, here's the patch that I was going to apply:

```patch
From 0a907e411ad0ef8b6bc93ed872847aa728cd182b Mon Sep 17 00:00:00 2001
From: Chris Reeves <chris@ev-soft.net>
Date: Wed, 5 Jul 2017 17:54:38 -0400
Subject: [PATCH] allow plain objects to extend context when firing - fixes
 #3033

---
 src/Ractive/prototype/fire.js |  2 +-
 tests/browser/events/basic.js | 23 +++++++++++++++++++++++
 2 files changed, 24 insertions(+), 1 deletion(-)

diff --git a/src/Ractive/prototype/fire.js b/src/Ractive/prototype/fire.js
index 9c98827c..94c9203c 100644
--- a/src/Ractive/prototype/fire.js
+++ b/src/Ractive/prototype/fire.js
@@ -3,7 +3,7 @@ import Context from '../../shared/Context';

 export default function Ractive$fire ( eventName, ...args ) {
        // watch for reproxy
-       if ( args[0] instanceof Context ) {
+       if ( args[0] instanceof Context || ( typeof args[0] === 'object' && args[0].constructor === Object ) ) {
                const proto = args.shift();
                const ctx = Object.create( proto );
                Object.assign( ctx, proto );
diff --git a/tests/browser/events/basic.js b/tests/browser/events/basic.js
index fbdba992..823165ee 100644
--- a/tests/browser/events/basic.js
+++ b/tests/browser/events/basic.js
@@ -311,4 +311,27 @@ export default function() {

                t.equal( count, 10 );
        });
+
+       test( `firing a pojo first arg extends context (#3033)`, t => {
+               t.expect(3);
+
+               const r = new Ractive();
+               class Foo {}
+
+               r.on('foo', (ctx, foo) => {
+                       t.ok(foo instanceof Foo);
+               });
+
+               r.on('merge', ctx => {
+                       t.ok(ctx.foo instanceof Foo);
+               });
+
+               r.on('str', (ctx, foo) => {
+                       t.ok(foo === 'foo');
+               });
+
+               r.fire('foo', new Foo());
+               r.fire('merge', { foo: new Foo() });
+               r.fire('str', 'foo');
+       });
 }
--
2.13.1
```

Basically, if the first param is an object that has no special prototype, use it to extend per the docs. Extending a context is used internally by the event proxying mechanisms, and since context is now required in the callback, it is also exposed to `fire`.

Issue #3038, opened 2017-07-10
Array polyfills are incorrectly enumerable


paulie4:
## Description:
The array*.js polyfills in https://github.com/ractivejs/ractive/tree/dev/src/polyfills are all added directly to Array.prototype without using Object.defineProperty(), which makes them enumerable. That makes `for (var idx in arr)` include the polyfilled properties, when it shouldn't.

Since we dropped support for IE8 and under (see https://github.com/ractivejs/ractive/issues/2597), we can safely use Object.defineProperty() to add the properties so they are not enumerable.

## Versions affected:
All

## Platforms affected:
IE9, IE10, IE11, and any other browser that uses the polyfills.

## Reproduction:
```
var arr = [];
for (var idx in arr)
    console.log(idx);
```

evs-chris:
I fixed array.find in edge because that is the only array polyfill required for IE10+ (I think). When I publish 0.9.3, it will include the changes that include the promise, array.find, and object.assign polyfills in the default build.

I'll get 'round to the rest of the polyfills, but if anyone wants an easy intro PR... 😄.

evs-chris:
I've added the `node.contains` polyfill to the default build and verified that IE10+ pass the test suite sans polyfill. There are still some `if (window...`s and direct assignments to prototypes that need to be addressed in the polyfills.

paulie4:
I checked MDN and also tested on Node.js, and all of these exist in IE9+ and in Node.js, so we can probably remove the polyfills, since we don't support anything less than IE9:
- Array.isArray
- Date.now
- Object.create
- Object.defineProperties
- Object.defineProperty
- Object.freeze
- Object.keys
- addEventListener (it's not in Node.js, but I doubt we need it there. MDN says IE9 supports it, but they do show [a few caveats](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener))
- Array.prototype.every
- Array.prototype.filter
- Array.prototype.forEach
- Array.prototype.indexOf
- Array.prototype.map
- Array.prototype.reduce
- Function.prototype.bind
- getComputedStyle (it's not in Node.js, but I doubt we need it there)
- Node.prototype.contains (it's not in Node.js, but I doubt we need it there)
- String.prototype.trim

[...]

PR #3039, opened 2017-07-11
remove old polyfills that should not be needed for IE9+ - #3038

paulie4:
for https://github.com/ractivejs/ractive/issues/3038


evs-chris:
Looking at the [kangax ES5 table](http://kangax.github.io/compat-table/es5/) and checking [caniuse](http://caniuse.com), I'd say it's safe to remove all of these. After that, there're only two polyfills left out of the main build - `performance.now` and `requestAnimationFrame`, which aren't particularly big, so it'd probably make sense to just include them in the main build as well?

ghost:
Yup. If we're only supporting IE11, pretty much all ES5 and web polyfills can be removed.

evs-chris:
According to the kangax chart, we would still support IE9+. We might lose some very old versions of safari or the android browser, but it looks like even those are partially supported enough to get by.

paulie4:
What features won't work for "some very old versions of safari or the android browser?" I see in kangax that a few JS engines fail on "Enumerable properties can be shadowed by non-enumerables" (which I think should be worded in reverse), and "Array.prototype.sort: compareFn must be function or undefined," both of which do seem like very edge case problems.

ghost:
I think "android browser" means pre-"Chrome WebView" Androids, versions below 4.4 that don't use Chrome as internal browser. Apple keeps iOS fairly updated so this shouldn't be an issue, unless you still use an iPhone 5 (which I think Apple ditched recently because they're not 64-bit) or older (which don't get iOS updates anymore).

Not familiar with the API support for those browsers.

paulie4:
@evs-chris, in [this comment](https://github.com/ractivejs/ractive/issues/3038#issuecomment-314262442) you mentioned the "polyfills [will be] in the default build", and in a comment in this PR, you said, "it'd probably make sense to just include them in the main build as well." What did you mean by "default build" versus "main build"? Also, are you thinking of putting the few remaining polyfills within built `ractive.js`, or would the build still have a separate `polyfills.js` file?

evs-chris:
default build, main build, and ractive.js are all the same thing in my ramblings.

It didn't really make sense to have a polyfill for two tiny functions, so I pulled them into ractive.js. I didn't remove the polyfill file, since that might cause errors for people that have a merged CDN link. It now just prints a message to the console that it is no longer needed, and it will be completely removed for 1.0.

Issue #3001, opened 2017-06-09
Default CSP doesn't like Ractive's CSS


ghost:
## Description:

While avoiding/pre-compiling expressions avoids the eval-related CSP warnings, CSP does not like inline CSS as well. While you'd think not providing `css` would avoid this, the initial setup does some inline CSS stuff, tripping warnings.

Proposed fix:
- Docs to inform developer that `style-src 'unsafe-inline'` can be used to bypass the whole thing.
- Not providing `css` shouldn't trip any warnings (fix that CSS setup code).
- Maybe an `allowCSS` that if set to `false`, totally ignores any CSS provided and doesn't print it too.
- To have component-scoped CSS, either
    - Write the CSS externally to boot. Use BEM or some other naming convention to scope.
    - For SSR, `css` can be extracted to a file using `Ractive.getCSS()` or `ractive.toCSS()`. Few problems:
        - `ractive.toHTML()` doesn't emit scoping ids nor emit the CSS if rendering a document. Related issue: https://github.com/ractivejs/ractive/issues/2706

Thoughts?

## Versions affected:

All?

## Platforms affected:

All?

## Reproduction:

http://plnkr.co/edit/uci3CFljlSJVVoPOIic5?p=info

## Code:

```html
<!DOCTYPE html>
<html>

  <head>
    <meta http-equiv="Content-Security-Policy" content="default-src 'self'">
  </head>
  <body>
    <script src="ractive.js"></script>
    <script src="script.js"></script>
  </body>

</html>
```

```js
Ractive({
  el: 'body',
  append: true,
  data: {
    msg: "Hello, World!"
  },
  template: '{{ msg }}'
})
```

evs-chris:
`toHTML` will emit scoping ids if there's css that needs to render. If you're gonna use it though, you probably need to set `cssId` so that it's something stable.

I think just doing a deferred init on the style element will suffice for getting rid of the csp error. It looks like setting up a nonce would be the best way to skirt the unsafe-inline requirement for ractive's css.

I'm not exactly how to approach the injected styles for progressive enhancement. Maybe throw an attribute on the style tag that tells ractive to take it over and then do `indexOf` checks for each chunk of css that would be added to it to see if it's already there?

evs-chris:
As of edge now, the style injection tag will be lazy loaded rather than created immediately during ractive lib init.

PR #3041, opened 2017-07-13
Add support for css functions for customizable component styles - RFC

evs-chris:
## Description:

I think the component situation is improving fairly rapidly with webpack support for the bin and rollup support coming when I sit still long enough to figure out some of the fringe details (loading external files). I'm in the process of reworking the playground into a (hopefully) good example of how to structure an app with components of both the widget-y and view-y flavors, and I'm trying to make sure the widget-y components are nice and usable elsewhere.

So far, that's all coming together nicely except for the reusability aspect of the widget-y components. They work great for the project for which they're developed, but unless other projects need the same look and/or feel, then you either end up adding a big external stylesheet tweaking the default styles (and that requires `noCssTransform`), you require an additional class for every instance and try to override from there, or you move all of the relevant styles inline and set them from data. None of those are particularly appealing for more than just a few instances.

I was originally going to add a build step for processing styles as a text-only template, but @fskreuz suggested letting the `css` option be a function. That makes a lot of sense, especially when combined with a data registry for use in the css function.

### Code

This:

* adds a check for `css` as a function during `extend`, and if it's a function, it will be called with the component's static `cssData` wrapped in a safe accessor function and in the context of the component.
* adds `cssData` that uses any parent `cssData` as a prototype and gets the object passed as the option assigned to it - so child `cssData` inherits parent `cssData`.
* makes components keep track of any extensions so that changing `cssData` and reapplying css can also cascade to children.
* adds ractive-like helpers for managing `cssData`
  * the `css` function is passed a function to safely retrieve keypaths from `cssData`. It also has `cssData` as its `__proto__`, so you can access the data directly if that's what floats your boat.
  * the component constructor gets a `styleSet` method that will safely set one or more keypaths
* makes the css apply function that manages ractive's component styles flag which styles have been applied. If `styleSet` affects a style that has already been applied, it will trigger a reapply.
* adds support to the bin component processor for css functions in a `script` element by using `rel="css"`, where the script is used as the body of the css function. The name of the param defaults to `data`, but can be specified on the script tag with `data-name="someOtherName"`
  * there's also support for using a ractive template for css with `<style type="text/css+ractive">` and `<link rel="ractive" type="text/css+ractive" href="other-file.css">`
  * if there are no computed css sections, then the `$CSS` var will be a string - otherwise, it will be a function that sticks all of the various css segments together
* exposes the component `cssData` as `@style` - because some style-y things could very well affect other parts of the component, so binding to the style var would be quite handy

and the mostly unrelated changes:

* `cssId`s are tracked at extension time so that they can easily be applied to top-level component elements
* the ractive style element isn't attached to the document until it's needed

### Example

```html
<span class-toggle class-on=".value" class-off="!.value" class-disabled=".disabled" {{#unless .disabled}}on-click="@.toggle('.value')"{{/unless}}>
  <span class="rail" />
  <span class="nubbin"/>
</span>

<script>
  import Ractive from 'ractive';
  const Toggle = Ractive.extend({
    template: $TEMPLATE,
    css: $CSS,
    cssId: 'toggle',
    cssData: {
      color: {
        on: {
          rail: 'rgba(0, 0, 0, 0.2)',
          nubbin: '#fefefe'
        },
        off: {
          rail: 'rgba(0, 0, 0, 0.3)',
          nubbin: '#eee'
        },
        disabled: {
          rail: '#888',
          nubbin: '#888'
        }
      }
    }
  });
  export default Toggle;
</script>

<script rel="css" data-name="d">
  return `
    .toggle {
      display: inline-block;
      position: relative;
      width: 1.7em;
      height: 1em;
      vertical-align: middle;
    }
    .rail {
      position: absolute;
      display: inline-block;
      top: 0.15em;
      height: 0.7em;
      width: 100%;
      border-radius: ${d.square ? '2px' : '0.7em'};
      background-color: ${d('color.off.rail')};
      transition: background-color 0.3s ease-in-out;
    }
    .nubbin {
      position: absolute;
      top: 0;
      left: 0;
      width: 1em;
      height: 1em;
      border-radius: ${d.square ? '2px' : '1em'};
      display: inline-block;
      transition: left 0.3s ease-in-out;
      background-color: ${d('color.off.nubbin')};
      /* this would be handy as an inherited variable for material component */
      box-shadow: 0 2px 2px 0 rgba(0, 0, 0, 0.14), 0 1px 5px 0 rgba(0, 0, 0, 0.12), 0 3px 5px -2px rgba(0, 0, 0, 0.2);
      transition: background-color 0.3s ease-in-out;
    }
    .on .rail {
      background-color: ${d('color.on.rail')};
    }
    .on .nubbin {
      background-color: ${d('color.on.nubbin')};
      left: 0.7em;
    }
    .disabled .rail {
      background-color: ${d('color.disabled.rail')};
    }
    .disabled .nubbin {
      background-color: ${d('color.disabled.nubbin')};
    }
  `;
</script>
```

Later at the use site, we decide we want red/green rails for off/on states rather than shaded parent background color, so:

```js
Toggle.styleSet({
  'color.on.rail': 'green',
  'color.off.rail': 'red'
});
```

### TODO

* [x] Tests.

#3001, and `cssIds` for `Component.extend` not having both the `Component` and extension ids.

## Is breaking:
Don't think so.


lib/component.js
```diff
@@ -68,11 +69,29 @@ function build ( string, opts, readFile ) {
 						promises.push( readFile( src ).then( str => partials[id] = str ) );
 					}
 				} else if ( !type || type === 'text/javascript' || type === 'application/javascript' ) {
+					const rel = getAttr( 'rel', item);
```
ghost: `<script rel>`?

- Preferring we keep the template at 3 major sections. Otherwise, it gets confusing.
- Taking rcu as reference implementation, [it enforces the single-`<script>`](https://github.com/ractivejs/rcu/blob/master/src/parse.js#L46) logic.
- `<script>` for styles is strange.
- Text editor highlighting will yell at you. 😄 

How about `<style type="text/ractive-fn-css">`? [Ractiveify](https://github.com/norcalli/ractiveify) uses `type` to hint pre-processing. We could use this to hint the bin for special treatment of CSS while being within the gray area of the spec - where there is no defined behavior if we provided different `type`. 😄 

We might soon exploit `type="text/ractive-*"` moving forward.

lib/component.js
```diff
@@ -68,11 +69,29 @@ function build ( string, opts, readFile ) {
...
+						cssFn = true;
+					}
+
 					if ( !src ) {
-						script.unshift( item.f );
+						if ( rel === 'css' ) {
+							style.unshift( makeCssFn( item.f.join('') ) );
+						} else {
+							script.unshift( item.f );
+						}
```
ghost: For this one, `<link rel="text/ractive-fn-css" href="...">`. Would make sense since this is how external CSS is loaded, plus `<link>` fits since it's a multi-purpose element for loading external things (favicon, etc.).

Spec is fairly specific of its us on components anyways `<link rel="ractive">`. 😁

lib/component.js
evs-chris: I was mostly just going for highlighting support with the script tag, I don't think there's any reason it couldn't be in a style tag beyond that. I have a few ideas for components that would generate chunks of styles based on config e.g. a grid framework with customizable units (I _need_ 14ths on this project because _reasons_). In that case the style fn would look much less like css as in the example and more like js.

tests/browser/render/css.js
```diff
@@ -419,4 +419,171 @@ export default function() {
...
+		new cmp2({ target: fixture });
+
+		const div = fixture.querySelector( 'div' );
+		const style = getComputedStyle( div );
+
+		t.equal( div.getAttribute( 'data-ractive-css' ), '{cmp} {cmp2}' );
+		t.equal( style.color, 'rgb(0, 128, 0)' );
+		t.equal( style.display, 'inline-block' );
+	});
 }
```
ghost: Might want to include tests for `toCSS` and `getCSS` for SSR purposes, and to pin down order-related things.

lib/component.js
```diff
@@ -68,11 +69,29 @@ function build ( string, opts, readFile ) {
 						promises.push( readFile( src ).then( str => partials[id] = str ) );
 					}
 				} else if ( !type || type === 'text/javascript' || type === 'application/javascript' ) {
+					const rel = getAttr( 'rel', item);
+
+					if ( rel === 'css' ) {
+						if ( cssFn ) return Promise.reject( `Only one css script may be specified per component.` );
+						else if ( style.length ) return Promise.reject( `<style> elements and css script cannot be mixed.` );
```
ghost: They could probably mix. They could concat into one big string in the order they appear, and the end result is the function version so that the functions would still work.

```
<style>
  .foo{ color: red }
</style>
<style type="text/ractive-css-fn">
  .bar{ color: ${ this.barColor }
</style>

// Becomes
css: function(){
  return `
    .foo{ color: red }
    .bar{ color: ${ this.barColor }
  `
}
```

But it shouldn't work if a `type` is present and contains something other than the function CSS type. That way, the tool can warn if say Sass was present.

```
<style type="text/scss">
  .foo{ color: $fooColor }
</style>
<style type="text/ractive-css-fn">
  .bar{ color: ${ this.barColor }
</style>
```

My 2c anyways. Alternatively, I've always been a fan of a single-`<style>` like the single-`<script>`. I find the spec strange that it allows for multiple `<style>` in a component file. The bin could start enforcing this.

giovannipiller:
What about avoiding any confusion by using `this.getCSS(...)` rather than `this.get(...)` from the <style>?

```css
.on .rail {
      background-color: ${this.getCSS('color.on.rail')};
    }
```

Given that we'll be already setting CSS data via `this.setCSS(...)`, it would seem logical to me to use `this.getCSS(...)` from the <style> too.

ghost:
Wondering if we could skip the whole `${ this.getMyAwesomeCssDataPath(...) }` and use Sass-style vars, string replacement, and enforce a flat `cssData`. 🤔

evs-chris:
The original implementation just had the `css` function take an argument that was the `cssData` object. After using it a bit, I realized that I don't like using camel casing for namespacing (`color.on.rail` as `colorOnRail`), and the safe js way to access nested props isn't particularly ergonomic (`color && color.on && color.on.rail`). I like Ractive's way of handling data access, so it seemed fitting to apply it here too, but perhaps putting a `get` on the context _is_ a bit confusing.

I'd guess that 99% of css functions would only care about `cssData` and not whatever else is the constructor. It's also not terrible to pass args to the function so any of these seem fairy reasonable:

1. Call the css function with the data access function as the context so that data refs become `this('some.path')`.
2. Pass the data access function as an arg, which feels pretty natural for direct init (```css(d) { return `div { color: ${ d('some.color.path') }; }`; }```. For bin components, you could specify the arg name as the `data-name` attr on the css script block, as was done originally with the data object. `<script rel="css" data-name="d">`
3. Change `this.get` to `this.getCSS`, though that's getting a bit noisy.

I would rather avoid any parsing in the css, as that either limits you to simple data access, comes with a sizeable parser, or leaves tons of corner cases in what sorts of expressions you can use.

I'm leaning toward 2 right now. I'm also going to add bin support for additional `style` tags that get automatically composed into the css function.

evs-chris:
I've updated this with an implementation of option 2. The css fn is still called in the context of the component, but it is passed an argument that is a function to safely access the data that also has the data as its prototype.

The bin will now also handle multiple `style` tags in addition to the css script. Additional styles are collected before and after the fn script, processed through any configured preprocessors, and wrapped along with the fn into a composite function.

evs-chris:
@JonDum @dagnelies I know you have a fair number of published widget-style components - if you happen to have a few minutes to look this over, I'd love to hear what you think.

evs-chris:
Summary of recent changes:

* style vars are now implemented as ractive models and exposed to templates/instance via `@style.whatever` - this is mostly because you might have some template structure that depends on a style var setting. It also saves a bit of code in the implementation.
* any forms of style in a component to be processed by the bin may be combined - the bin will handle creating the appropriate string or function to compose them.
* a style or link element with `type="text/css+ractive"` will be parsed as a `textOnlyMode` template and have a function created for it that renders using a ractive instance and the component `cssData`

MartinKolarik:
Can we use `rel=ractive` instead of `type="text/css+ractive"` ? Custom `type` is going to break syntax highlighting in some editors and `rel` is also used in other Ractive tools.

evs-chris:
We can. Would it be better to just do detection based on whether or not a parse turns up more than just a single string chunk?

dagnelies:
Hi,

I'm currently on parental leave, taking some vacations, so I'm not really in a "busy" mode. 😎 

Personally, I'm *veeeeeeeeeery* happy about the `{{>extra-attributes}}`. 😀 I think it was a major roadblock for making reusable components, even if people aren't very aware of it. What's more missing is the ecosystem as a whole and making some noise IMHO.

On a similar note, CSS was never a big deal to me. Just import a CSS framework, add a few classes, or your own styles and that's it. No biggy. Therefore, I must confess my interest in this is very limited.

MartinKolarik:
@evs-chris I can't think of any situation where that wouldn't work but still think explicit `rel` is better.

giovannipiller:
@evs-chris explicit rel seems better

evs-chris:
Updated with:

* `<style rel="ractive">` is parsed as a template and plain `<style>` is just text
* `<link>`s are a bit different, as they already required `rel="ractive"` or `rel="component"` to be consumed, so the `type` still differentiates between plain text and a template: `text/css` for plain, `text/css+ractive` for template

@dagnelies parental leave? If that's what I think, congratulations!

As far as ecosystem goes, this will hopefully make growing the ecosystem a bit easier for component authors. I have a number of components that I haven't really published because they're tailored to particular projects, and while they would probably work out well in others, they aren't really generic enough to fit in most places. Up to now, I've used stylus and variable support there to make style somewhat configurable, but that has the downside of requiring anyone who wants to customize rebuild everything. With this, you can build in a fair amount of configurability and users can pull the already compiled and ready to go component from npm/gh/jsdelivr/wherever, set a few vars, and roll on.

CSS vars have the potential to make component styles a bit less complicated eventually, but they lack the power available to templates or functions, namely loops, conditionals, and the ability to throw whole chunks of CSS into a single variable as an extension point that gets automatically scoped.

Issue #2271, opened 2015-11-11
RFC: Plugins monorepo


Rich-Harris:
Plugins are something of a ghetto at the moment. Clicking [Plugins](http://docs.ractivejs.org/latest/plugins) in the nav of any ractivejs.org page takes you to a page on docs.ractivejs.org (with the Docs tab highlighted), containing a bunch of links, many of which are to out-of-date pages (I've been pretty dreadful at keeping the 'official' ones up-to-date).

On top of that, the process for creating plugins is unnecessarily confusing and inconsistent and out-of-date (hello, unmaintained Grunt template).

## Solution: a monorepo

Specifically, [this monorepo](https://github.com/ractivejs/ractive-plugins). There are a few clear advantages:

* We can automate lots of drudge work – building dist files, generating high-quality package.json files, etc
* Much easier to maintain consistency of code style and so forth
* Creating new plugins is as simple as creating a new folder and cribbing from similar plugins. No awkward project setup. Everyone gets ES6 out of the box
* Discoverability is much better – for plugin users, plugin developers (easier to avoid duplication of effort), Ractive developers (at the moment it's hard to keep track of which plugins are in the wild)
* When plugin APIs change (which is inevitable, e.g. we'll probably need to overhaul the adaptor and transition APIs at some point) it's much easier to update plugins en masse
* Easier to have a coherent testing strategy
* It'll make it much easier to generate a plugins.ractivejs.org site, with (for example) automatically generated installation instructions and demos for each plugin

## Downsides

* Some loss of flexibility – if a plugin needed a very specific build process, we'd have to accommodate that somehow
* Third party plugin authors would need to dance to our tune if they wanted to be in the official plugin repo (maybe not a bad thing though – easier to ensure quality across the ecosystem)
* The repo *could* get huge, eventually

Lots still to do but I thought I'd put this out there to gauge initial reactions. Let me know your thoughts

JonDum:
How is it going to work with NPM? How will you make tags and versions if it's all a single repo?

How will we standardize on templates & css? Or is this *only* for transitions, adaptors, events. What about decorators and components then? Rollup doesn't handle anything besides js, right? My "solution" to this has been webpack with the side benefit of being able to point the package.main to the `src/` file and completely avoid the additional umd/loader bloat (but you'd have to be using webpack also :|).  

[...]

PR #3052, opened 2017-08-10
WIP: Monorepo

ghost:
## Warning

I occasionally force-push to this repo to fix existing commits (fix-ups) or when I rebase it against dev (to update the branch). This branch is not final, and is a work in progress.

## Description

TL;DR: Move everything Ractive into one meta-repo. 

Reserve this thread for implementation review only. Route all conceptual/theoretical discussions to the monorepo issue (https://github.com/ractivejs/ractive/issues/2271) to avoid discussion in multiple places (I like clean, historical documentation 🤓 ).

### On repo

```
- package.json
- [other files]
- packages/
  - [package name]/
    - package.json
    - README.md
    - src/
      - rollup.config.js
      - index.js
      - [other files]
    - test/
      - rollup.config.js
      - karma.conf.js
      - index.js
      - [other files]
```

### When published (for each package)

```
- package.json
- README.md
- dist
    - lib.umd.js
    - lib.es.js
```

ES build will ship _pre-transpiled_ so that end-users don't have to futz around with stage support.

## Things to do

- [x] Make a Ractive npm org.
- [x] Linting rules.
- [x] Naming.
- [x] Dev setup.
- [x] Test setup.
- [x] Rewrite each plugin to conform to new setup.
- [ ] Unify utility APIs (essentially just updating and simplifying rcu).
    - [x] Parse component file to parts.
    - [ ] Allow pre-processing tools to manipulate parts (i.e. transpilation).
    - [x] Construct ES module from parts.
    - [ ] Hi-res source maps when disassembling the component file to parts.
    - [ ] Hi-res source maps when assembling the parts to a module.
- [ ] Loaders
    - [ ] Rollup
    - [ ] Webpack
    - [ ] Gobble
    - [ ] Gulp
    - [ ] Broccoli
- [x] Automated testing.
- [ ] Automated publishing.
- [ ] Phasing out the old projects (deprecation notices everywhere).

- https://github.com/ractivejs/ractive/issues/2271
- Outdated dependencies.
- Maintenance.
- Incompatibilities.

## Is breaking:

Yes

### DIFF
--- a/src/view/items/Partial.js
+++ b/src/view/items/Partial.js
@@ -249,7 +249,7 @@ function partialFromValue ( self, value, okToParse ) {
 
 	if ( isArray( tpl ) ) {
 		self.partial = tpl;
-	} else if ( isObjectType( tpl ) ) {
+	} else if ( tpl && isObjectType( tpl ) ) {
 		if ( isArray( tpl.t ) ) self.partial = tpl.t;
 		else if ( isString( tpl.template ) ) self.partial = parsePartial( tpl.template, tpl.template, self.ractive ).t;
 	} else if ( isFunction( tpl ) && tpl.styleSet ) {
--- a/tests/browser/partials.js
+++ b/tests/browser/partials.js
@@ -1183,4 +1183,16 @@ export default function() {
 		r.set( 'foo', 'baz' );
 		t.htmlEqual( fixture.innerHTML, 'bar' );
 	});
+	
+	test( `null variable with a name of a missing partial doesn't break anything (#3154)`, t => {
+		new Ractive({
+			target: fixture,
+			template: '1{{>part}}2',
+			data: {
+				part: null
+			}
+		});
+		
+		t.htmlEqual( fixture.innerHTML, '12' );
+	});
 }


evs-chris:
I'm ill at the moment, so I'm a little fuzzy - I'd say that having everything integrated such that everything gets tested as changes are made to ractive would be a good thing. Having docs in the main repo would also be cool.

I'd hesitate to split the cli out as a separate project though. Having the component/partial/template parsing code sitting in the same package as the actual parser has been tremendously beneficial. I would rather pull rvc into the lib and expose its functionality as an export - which reminds me... I really need to get some tests in place for lib functionality and make it browser friendly.

I'll ponder a bit more when the haze clears 😄

.editorconfig
```diff
@@ -5,11 +5,8 @@ charset = utf-8
 end_of_line = lf
 insert_final_newline = true
 trim_trailing_whitespace = true
-
-[*.js]
-indent_style = tab
+indent_style = space
```
evs-chris: Don't let @Rich-Harris see that. Them's fightin' words. 😆

ghost:
> I'm ill at the moment, so I'm a little fuzzy

😱  Get well soon!

> Having docs in the main repo would also be cool.

Mkdocs is actually designed for that. It was part of the plan all along. 😉 

> I'd hesitate to split the cli out as a separate project though.

Yep, just trying this out to see how it works. It may not end up separated in the end. But then (assuming I understand Lerna correctly), having them in the same version line should make it just work as if it was shipped with Ractive, albeit in a separate package.

Still figuring out how inter-package development works tho. 😵 

> I would rather pull rvc into the lib and expose its functionality as an export 

Yep. There's a significant overlap between the cli, rcu, rcu-builders and the loaders. Planning to move that out to a utility library and have loaders/builders just be function-specific wrappers to those functionality. The `cli` pretty much does that already with bin+lib.

ghost:
Wild idea: What if, instead of publishing separate packages/repos, just ship everything with `ractive`? If you want a compatible plugin, tool, or utility, look no further than the `ractive` package itself? Everything comes compatible with each other, bundled to the same level. Kinda like lodash functions, you can either import the entire thing or the bits and pieces version.

What are the benefits of shipping them separately as opposed to having them together?

There's also not a whole lot of them. And some, especially the events, are just under a hundred lines.

capture

.editorconfig
Rich-Harris: HULK SMASH

auvipy:
just dont use bower as its depricated and the bower maintainer suggested to use yarn or webpack

ghost:
@auvipy It's in the list because I still use it in some projects (omg, old PHP and Java websites). The idea of dropping Bower haunts me, especially when times comes when Ractive is _the_ tool for the job.

But then this is still a proof-of-concept. Probably by the time this is done, Bower is long gone. Also, repo publishing is the last on my checklist since it involves some Bash and Git sorcery. That can easily be dropped. 😁

sarendipitee:
How are you going to deal with people who have differing tastes in how they publish ractive components?

ghost:
It shouldn't affect anyone. People can still publish their integrations however they want, wherever they want. The monorepo just pulls in everything already in the Ractive org into one repo, that's about it. It's not forcing everyone to join.

I might be missing a use case tho. What do you have in mind?

---

Is this the appropriate time for the `#courage` hashtag? _(Disclaimer: Clearly not an Apple fan)_ 😁

sarendipitee:
@fskreuz It's actually getting really annoying maintaining all my ractive components as ractive changes. I think I may actually be on board to have one repo with Lerna so I can standardize builds/publishes better. What would it take to bring one of my components into your monorepo as a PoC?

ghost:
@JonDum Nothing's final yet, everything's very volatile on this branch. 💥 

But so far, the build only expects a package directory in `packages` with at least a `README.md`, `package.json`, `src/index.js` (lib entrypoint), `tests/index.js` (tests entrypoint). Source and tests should be written in ES6+. Tests are anticipated to be QUnit on Karma, similar to Ractive. The goal is to have packages just be sources and tests, with all heavy-lifting done centrally in one place.

sarendipitee:
@fskreuz What about importing other assets like styles and images (for components)?

My components currently use Webpack to embed Stylus into the javascript as css strings that get automatically added to the page on import.

If someone tries to `import RactiveFoo from 'ractive/packages/ractive-components-foo'` that 'ractive-components-foo' needs to be prebuilt unless they are using webpack with the correct loaders.

sarendipitee:
Or, is the point of this just to have one central place for all the packages, but they'll still be published separately under their own npm packages?

evs-chris:
I think if the package is in this repo, then it should be published in the `@ractivejs` scope. No reason it couldn't also be published elsewhere too, though. That probably means we need to settle on a base build method, and it looks like webpack is a pretty good option. I'm currently leaning towards publishing component src and pre-built modules together so you can go either way - direct include as a script tag for the pre-built, no-deps module (other than Ractive), or use webpack yourself and get all the benefits of using webpack.

I haven't explored pre-built modules yet, as I have all of my components packaged to be consumed from a build tool. I'm also not using external style processing, though I do use a js transpiler for es20-whatever.

(Also, I'd like apologize for breaking shit. Breakage should be very minimal henceforth, as we're nearly to 1.0. Everything surrounding 0.8 was a little painful.)

ghost:
>  is the point of this just to have one central place for all the packages, but they'll still be published separately under their own npm packages?

Yup, that's the plan - host, build, and test everything in one place but publish them individually. The packages will all be published under an org this time (i.e. `@ractivejs`). All packages will be published pre-built (transpiled, bundled, minified, sourcemap-ified) as ES and UMD modules so you can choose your weapon/poison.

From the consumer point of view, the only immediately visible difference will be a fancy `@ractivejs` prefix on the dependency and probably a minor renaming.

>  Breakage should be very minimal

No problem. I'm working on `packages` exclusively and the packages in there are just for reference. When the build is set, I'll be updating whatever needs update. Breaking Ractive shouldn't be a problem.

paulocoghi:
I know I'm getting late in the matter, but would not [github.com/lerna/lerna](https://github.com/lerna/lerna) be of any help in this new organization?

ghost:
Will be working on a fork to try out travis in isolation... and not flood gitter or the gh feed with every push I make.

Issue #3051, opened 2017-08-09
Not closed tag inside if blocks are not highlighted by parser


marcalexiei:
## Description:
If you include a component inside an _if_ block without closing tag the parser compile template wihtout error and when the condition is _true_ the if content is rendered correctly in the page.
However when condition is _false_  the content inside _else_ block is not rendered in the page.
Content after both blocks is rendered correctly in both case.

Another issue appears with not closed HTML tags: if condition is _true_ both _if_ and _else_ blocks are displayed.

Should parser throw an error on those situations?
 
## Versions affected:
- 0.9.x
- edge

## Platforms affected:
- All

## Reproduction:
*Component scenario*
1. Open https://jsfiddle.net/hjgmp41r/
2. When _showMe_ is _false_ "I'm inside the else" is not displayed.

```html
<div id="main"></div>

<script type="text" id="template">
    <h1>Hello {{who}}!</h1>
    <p>We are ready for some bugs!</p>

    showMe: {{showMe}}

    <button on-click="@.toggle('showMe')">Toggle</button>

    <div class="comp">
      {{#if showMe}}
        <MyComp>
      {{else}}
        I'm inside the else
      {{/if}}
    </div>

    <p>other content</p>
</script>
```

```javascript
Ractive.components.MyComp = Ractive.extend({
    template: 'Hi, I\'m a template!!!!!',
});

const r = window.r = new Ractive({
    el: '#main',
    template: '#template',
    data: {
        showMe: false,
    }
});
```

*HTML scenario*
1. Open https://jsfiddle.net/axxw71xp/
2. When _showMe_ is _true_ both blocks are displayed.

```html
<div id="main"></div>

<script type="text" id="template">
 <h1>Hello {{who}}!</h1>
    <p>We are ready for some bugs!</p>

    showMe: {{showMe}}

    <button on-click="@.toggle('showMe')">Toggle</button>

    <div class="comp">
      {{#if showMe}}
       <div>showMe is true
      {{else}}
        I'm inside the else
      {{/if}}
    </div>

    <p>other content</p>
</script>
```

```javascript
Ractive.components.MyComp = Ractive.extend({
	template: 'Hi, I\'m a template!!!!!',
});

const r = window.r = new Ractive({
  el: '#main',
  template: '#template',
  data: {
  	showMe: false,
  }
});
```

evs-chris:
TL;DR the `{{else}}` is being eaten as a plain interpolator because the component context owns the parser when it appears.

[...]

PR #3054, opened 2017-08-12
Manage unclosed tag in if-elseif-else section

marcalexiei:
## Description:
When user leaves _element_ or _component_ tag opened add `elseif` and `else` to implicit closing section tag logic to avoid wrong template render.

At the beginning I was intended to thrown an error for this scenario, however I have found that  a logic for not closed tag is already in place so I've added _elseif_ and _else_ readers to avoid breaking change.

#3051

## Is breaking:
No.


evs-chris:
Nice!

I think this is _technically_ breaking, but i also think it doesn't matter in practice, as the odds of using `{{else}}` as an interpolator or outside of an if block are pretty slim.

PR #3065, opened 2017-08-25
Try out a bit of minified shrinkage with a custom build of buble

evs-chris:
Looks like it will shrink the minified build by ~5k.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 
 # Dependencies
 node_modules/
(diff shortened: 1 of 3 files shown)


evs-chris:
Great scott! The package lock thingy for npm 5 is ginormous... even more so than playground links!

ghost:
I usually ignore the lock file unless some strange deps issue happens (which has not happened... yet) 😁

Issue #3067, opened 2017-08-27
Is there an easy way to get the keypath of a component's event?


ceremcem:
# Orig: Is there an easy way to get the keypath of an event source?

## Description:

Consider the following template: 

```html 
<ul>
{{#each curr.myArray:i}}
    <li on-click="['aHandler', i]">{{.foo}}</li>
{{/each}}
</ul>
```

```js
...
on: {
    'aHandler': function(ctx, i){
        curr = this.get('curr');
        item = curr.myArray[i];
        item.foo = "hello!";
        this.set('curr', curr);
    }
}
```

When `x` is clicked an action is taken by taken `i` into account in order to calculate `curr.myArray[i]`.

Is there an easy way to get the clicked keypath, like `"curr.myArray.3"` ?
 
This would be handy when we have a button in a deeply nested document, for example, which will add or remove an item to the array. 

## Versions affected:

0.9.3

ghost:
I think that would be [`context.resolve()`](https://ractive.js.org/api/#contextresolve) without args.

evs-chris:
Yep, that's what `context` is made for. You can event store a ref to the context and use it to modify the array with `context.push('../', { newItem })`.

ceremcem:
[...]

PR #3070, opened 2017-08-29
Allow components to request additional mappings during construct

evs-chris:
## Description:

This adds an extra step during component setup that checks for `this.mappings`, which is accessible from `this.component.mappings` in the `construct` event, and if present and an array or a string, will install additional mappings before the init of the component. That _sounds_ pretty janky and like a violation of encapsulation, and it is the latter, but it's quite useful when you're processing the component template to split out partials and related attributes for use in the main component template. For instance, in a tab component that allows you to specify a `disabled` attribute that can be a mapping on a particular tab:

```html
<tabs>
  <tab title="tab 1" disabled="{{tab1disabled}}">tab1 content</tab>
  <tab title="tab 2">tab2 content</tab>
</tabs>
```

The template is processed such that during init, the two tabs are turned into an array with inline partials for the content and properties for attributes. In the case of the `title`s, they're static strings, so there's no need to do any mapping or yielding (in the actual component, the titles are _also_ turned into partials and yielded, but that's not relevant to this PR). The `disabled` attribute needs to be used in a conditional though, but there's no way to do so without trying to reverse engineer the reference passed in the interpolator, which is possible with a simple reference but gets very difficult if it happens to be an expression or reference expression. It's also not possible to use a yielded partial as a value, as that doesn't really make sense, so `{{#each tabs}}...{{#if yield .disabled}}...` can't happen.

Creating an additional specifically named mapping can solve this by taking the interpolator passed to the `disabled` attribute, giving it a generated name, requesting that the component create a mapping with the name and interpolator, storing the name on the tab item, and referencing _that_ in the component template using a reference expression e.g. `{{#each tabs}}...{{#if ~/_generated[.disabled]}}...`. The parent instance still controls the expression inside the interpolator, and the child component can use it in its own template.

This also has other uses, such as in #3067, where the event generated inside a component around a yield is trying to also exist within the yielded context. Simply creating a mapping for the scope and wrapping the component template in a `{{#with scope}}` resolves the mismatch of expectations. There's a hacky way to achieve that now that assumes you're familiar with the internals of how components work, but it seems like it would be a useful thing to expose in a slightly less foot-gun-prone way.

I have dogfooded the hacky way quite a bit, and it definitely makes components _much_ more used friendly from the template side. It also helps avoid whole nested component-within-component weirdness that is typically recommended for things like tab components in ractive. I'll hopefully be able to start polishing up and publishing some of these components soon.

### Usage

```js
Ractive.components.aaa = Ractive.extend({
  template: '{{#with scope}}<span on-click="click">{{yield}}</span>{{/with}}',
  on: {
    construct() {
      if (this.component) this.component.mappings = [{ t: 13, n: 'scope', f: [{ t: 2, r: '.' }] }];
    }
  }
});
```

which is the replacement for the example from #3067:

```js
Ractive.components.aaa = Ractive.extend({
	template: '{{#with scope}}<span on-click="click">{{yield}}</span>{{/with}}',
	on: {
		construct() { // this has to be in construct because at config, the mappings are alread in place
			const cmp = this.component; // reference to the component item containing this instance
			if (!cmp) return; // can't mangle the outer template for mappings if not a component
			const tpl = cmp.template.f; // the fragment array for the component template
			const attrs = cmp.template.m ? cmp.template.m.slice() : []; // copy the attrs array so we can safely mangle it
			cmp.template = {
				e: cmp.template.e, // copy the component name
				f: tpl, // copy the fragment
				t: cmp.template.t, // copy the vdom type
				m: attrs // swap in our own attrs array
			} // the shallow copy prevents future renders of the template from being pre-mangled
			
			// create a new mapping on the template
			attrs.push({
				t: 13, // it's an attribute
				n: 'scope', // named scope
				f: [{ // and the content is an interpolator {{.}}
					t: 2,
					r: '.'
				}]});
				
				// now there is effectively a new attr on the component 'scope="{{.}}"'
		}
	}
});
```

If you don't mind parsing at runtime, there's also:

```js
Ractive.components.aaa = Ractive.extend({
  template: '{{#with scope}}<span on-click="click">{{yield}}</span>{{/with}}',
  on: {
    construct() {
      if (this.component) this.component.mappings = 'bind-scope="."';
    }
  }
});
```

#3067 

## Is breaking:

No.

### DIFF
--- a/src/Ractive/config/runtime-parser.js
+++ b/src/Ractive/config/runtime-parser.js
@@ -16,7 +16,8 @@ const parseOptions = [
 	'stripComments',
 	'contextLines',
 	'parserTransforms',
-	'allowExpressions'
+	'allowExpressions',
+	'attributes'
 ];
 
 const TEMPLATE_INSTRUCTIONS = `Either preparse or use a ractive runtime source that includes the parser. `;
--- a/src/parse/_parse.js
+++ b/src/parse/_parse.js
@@ -72,6 +72,8 @@ const StandardParser = Parser.extend({
 		this.csp = options.csp;
 		this.allowExpressions = options.allowExpressions;
 
+		if ( options.attributes ) this.inTag = true;
+
 		this.transforms = options.transforms || options.parserTransforms;
 		if ( this.transforms ) {
 			this.transforms = this.transforms.concat( shared.defaults.parserTransforms );
--- a/src/view/items/Component.js
+++ b/src/view/items/Component.js
@@ -12,19 +12,22 @@ import render from 'src/Ractive/render';
 import Item from './shared/Item';
 import ConditionalAttribute from './element/ConditionalAttribute';
 import createItem from './createItem';
+import parser from 'src/Ractive/config/runtime-parser';
 
 export default class Component extends Item {
 	constructor ( options, ComponentConstructor ) {
 		super( options );
-		this.isAnchor = this.template.t === ANCHOR;
+		let template = options.template;
+		this.isAnchor = template.t === ANCHOR;
 		this.type = this.isAnchor ? ANCHOR : COMPONENT; // override ELEMENT from super
+		let attrs = template.m;
 
-		const partials = options.template.p || {};
-		if ( !( 'content' in partials ) ) partials.content = options.template.f || [];
+		const partials = template.p || {};
+		if ( !( 'content' in partials ) ) partials.content = template.f || [];
 		this._partials = partials; // TEMP
 
 		if ( this.isAnchor ) {
-			this.name = options.template.n;
+			this.name = template.n;
 
 			this.addChild = addChild;
 			this.removeChild = removeChild;
@@ -32,7 +35,7 @@ export default class Component extends Item {
 			const instance = Object.create( ComponentConstructor.prototype );
 
 			this.instance = instance;
-			this.name = options.template.e;
+			this.name = template.e;
 
 			if ( instance.el ) {
 				warnIfDebug( `The <${this.name}> component has a default 'el' property; it has been disregarded` );
@@ -58,6 +61,17 @@ export default class Component extends Item {
 
 			construct( this.instance, { partials });
 
+			// these can be modified during construction
+			template = this.template;
+			attrs = template.m;
+
+			// allow components that are so inclined to add programmatic mappings
+			if ( Array.isArray( this.mappings ) ) {
+				attrs = ( attrs || [] ).concat( this.mappings );
+			} else if ( typeof this.mappings === 'string' ) {
+				attrs = ( attrs || [] ).concat( parser.parse( this.mappings, { attributes: true } ).t );
+			}
+
 			// for hackability, this could be an open option
 			// for any ractive instance, but for now, just
 			// for components and just for ractive...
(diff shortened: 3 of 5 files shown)


ceremcem:
@evs-chris Thank you for such a great support. 

I'm poking with the new feature and yes, we can achieve the same goal **way easier** with that approach, without need to know component internals, which is good for preventing new team members run like hell from Ractive :) 

But still, I want to ask you one more thing: What should we do to make a component's context absolutely compatible with a standard tag's context? 

Sticking to the same example we can set `foo` key in the array like so: 

```js
x: function(context){
     this.set(context.resolve() + '.foo', 'this works! ' + context.resolve());
}
```
and the result is: 
```
clicking this works correctly : this works! curr.myArr.0
clicking this works correctly : this works! curr.myArr.1
clicking this works correctly : this works! curr.myArr.2
```

...which is exactly the expected thing.

On the other hand, replacing the `<span on-click="x" ...` with our component (`<aaa on-click="x" ...`) requires modifying the `x` handler as `context.resolve()` resolves to `scope` string, not `curr.myArr.0`, which requires changing `this.set(context.resolve() + '.foo', ...` to something else. 

Do we have a chance to make our component's context compatible with a standard tag's context, so that we could be able produce a drop-in replacement components when needed?

ceremcem:
My unsuccessful attempt is [here](https://ractive.js.org/playground/?env=docs#N4IgFiBcoE5SBTAJgcwSAvgGhAF3gDxICWAbgATFIC8AOiAPYCuuADi-QHwED0JpnWgDthBAM4BjGMVa5yuAJ6sEdPAgAeuHjACGE3GQT1KNesSHtcXYbVwEAtgoC0EhvdYMhCIXIlMYMKrAwH4BGBjWInLkMQRMADaCUTEpKcEAxAh6YOShMAB0jgCCYRHJqSkE8cRJthX15AQ6zeSeLtUSANaq6lwSHZ3mKPJgxGLkSAwI40IMcgDuDDCd5AAUAEYsuWAIXSMIuZ5iDPEIAJTkkOTB+QBmDAzhvM06tdH1vNVvDdfAPFkSMDhGzvRo8BKCKxRXiOFxuDxeHxvEEOZyudyeby+fyBejBPLAkDfSoQkE-DIAnJ5QoKEowYHlD5fMk-SpiVg6IStITtYhdHp9AZDEZjciLZbjVwBXa4eIKS6-O4PJ48dmc4kfHjMxkVYL-bIMuqVcGJWwoniw9EIrFJXiSaSyTggHBieAAJT0BlICHyVsxPjE+Re5Go5A9+kM+Q0uG8SFWwBBMfc8R0MauAHIMvNiLgco5JAxlE81Vy2v0+d16AB9ctdLjBBTEBDxJAqkucPXZ3PhdNYEGeK4JqK2VxCMS4GBMfSrC7AcjCerEW5rXNjX3w-24C6rwN+xG4Qo6VisIbjUPp9bmJBOfOuZSqfL0dMAbhBtmwC5SNYGq1HMc0s6fjEI64OoIYioGaC4AAwp4-64L+cHRvksxIOcr5cikI5HCcPrxAwKCrPQtYrH+0ZXPQWC5KB+QwNMJzejOZwYak2E+MhdHHPE3rgbcTBCBGngzkOrFWBOCC4P4XL6OotH0dxCAzixWG4GUok7ncxB0as6Ykb2hzsQBykxGU77CBgZw2EI4Zej6e5YmIADa6aWhu+7pgAuuBNmRtGsbxomCDJqmCAZsEnBkT4PZ9lEA7XG+oFXHxAkGEJZEAUOZpiaOXF4QRREgJxDHINR6gUSA5AANQGfBclcYxZzMQltgaWIEmIYZB5FQpM6VemSoMPp6Y7mKSydGIACE5DplVNUcfJDVNcOqkgmUFlWV48xhp6hgBVEzYZukzBsCwvaBcFabTek5iWGdURIKmOjQAleSDs1uDFAEVyOe9tjAB+y1Zf9MVZUDplibYHkJeDa1nJgQA)

sabob:
To use this users have to know the AST to add mappings?
> t: 2
> f: [{ 

I probably misunderstand something.

evs-chris:
@sabob No, you can also use a string as in the third block if you know that parsing will be available. Most people seem not to ship the runtime-only version of ractive, so it's probably reasonable. That drops it to `this.component.mappings = 'bind-mapping1=some.ref bind-mapping2=other.ref'` etc. This is intended as a pretty low-level hatch made slightly less painful for those who find themselves in a weird situation that need it.

Inlining the ast will always be faster though, and it will work no matter the runtime flavor of ractive. The only downsides are requiring a _bit_ deeper knowledge of internals, which if you're writing components that require this sort of thing probably isn't too big an imposition, and it's not absolutely guaranteed to be insulated from template format changes. That said, the template format for mappings hasn't changed since before I started working with Ractive.

sabob:
thanks Chris, so no way to hide that in an API?
addMapping(atttr, something)

evs-chris:
Probably could, but it would seriously limit what you can process through it e.g. expressions and reference expressions. That would kick the pass-through case on processing sub-templates in a complex container component.

evs-chris:
@ceremcem that's not currently possible, as the component will always inject it's own context at the component boundary. It really sounds like something that should be possible, so I'll poke at it a bit and see what shakes out.

PR #3083, opened 2017-09-08
Add !Array.isArray() checks to partial lookups (fixes #3082)

### LINKED ISSUE #3082: Conflict between partials and line positions, missing hasOwnProperty check
First, a demo:

```js
new Ractive({
    el: 'body',
    template: Ractive.parse('<div>{{>sort}}</div>', { includeLinePositions: true }),
});

// Uncaught TypeError: Cannot read property 't' of undefined
```

https://jsfiddle.net/4cLtuk81/2/

A `p` property is used to store inline partials as well as line positions, which can lead to some hard to find bugs. [Here](https://github.com/ractivejs/ractive/blob/76e49bd6ba0ef982dd8e7fd9e3e89adf779a29f8/src/view/items/partial/getPartialTemplate.js#L92) Ractive checks for `p[name]`, but `p` can also be an array of line positions that have nothing to do with partials.

Because Ractive uses `p[name]` rather than `p.hasOwnProperty(name)`, this check returns `true` for all methods on `Object.prototype`, and when `p` contains line positions, also for `Array.prototype` methods (which causes the above bug).

To fix the demonstrated bug, adding `.hasOwnProperty()` would be enough, but I think line positions should also move to a different property, as there might be other edge cases this conflict causes.

MartinKolarik:
#3082

## Is breaking:
No.

### DIFF
--- a/src/parse/_parse.js
+++ b/src/parse/_parse.js
@@ -125,7 +125,7 @@ const StandardParser = Parser.extend({
 						}
 
 						// watch for partials
-						if ( node.p ) {
+						if ( node.p && !Array.isArray( node.p ) ) {
 							for ( const k in node.p ) walk( node.p[k] );
 						}
 					}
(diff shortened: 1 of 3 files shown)


evs-chris (approved):
:+1:

Issue #3099, opened 2017-09-29
Question: Async/lazy/dynamic components/elements


evs-chris:
I've been pondering a few different things primarily related to decorators, async loading of components, and other dynamic-ish template things. Everything relates to dynamic/lazy loading of components and also to dynamic lightweight component-like elements.

**WARNING**: What follows is a long and rambly. The TL;DR is that I think a new simple template item that is a sort of component/partial/decorator/parser transform hybrid would be tremendously useful, especially for implementing future constructs as external plugins e.g. dynamic components (`<Component component-name="{{.type}}" ... />`) and async components (`<Async async-src="some/path" ... />`).

## Async components
Loading a component that loads and caches another component to use as a pass-thru should definitely be possible. Something like `<Async async-src="some/path/or/something" mapping1="{{.foo}}" etc="you get the idea"><loading>Not loaded yet...</loading>Stuff outside of the loading tag there is going to be supplied to the actual component as content...</Async>`. I think all it would take to implement that would be to add a low-level API that allowed a component to cause its host VDOM node to unrender and render again - like maybe `this.component.reload()`.

The placeholder component would need to be supplied with a way to turn a name/path string into a Promise for a component constructor, which with webpack would just be fronting `System.imports(path')`. It could then check its internal registry upon `config` and if the component was already loaded, skip displaying the placeholder, update the component VDOM name and maybe template a little, and trigger a deferred reload. If not loaded, go ahead and set up the placeholder render, start the async load, and wait until the load finishes to trigger the reload. All-in-all not terrible, though there is some mild abuse that would need to take place at the template AST level, but for a low-level component I think that's not unreasonable.

## Dynamic element constructs/lightweight component-ish things
Pseudo-related to that - 0.9 introduced experimental parser transforms, which allow you to supply a hooks that process templates on their way out of the parser and transform them at the element/component level. This is a very convenient process for things like turning `<card><title>stuff</title>some stuff</card>` into `<div class="ractive-material-card"><div class="ractive-material-card-title">stuff</div><div class="ractive-material-card-body">some stuff</div></div>`. Using that and a context alias to `@local`, you can achieve something like a lightweight component with no performance overhead. Obviously, you don't get things like complete data isolation, full instance event support, and independent custom css on your "component", but there is a pretty large class of widget-y components that don't call for those sorts of things.

[...]

Issue #3089, opened 2017-09-15
Async components - feature request


kouts:
It would be nice if we could have async loaded components and use webpack for code splitting.
This basically is a continuation from here: #2830 
I thought I'd open this to see if anyone else think this should come handy.

evs-chris:
I'm using webpack with the bin loader to do async loading of different chunks of an app via dynamic import and splitting. The async chunks tend to be more view-style things that are loaded in anchors, but widget-style components work as well. Are you looking for something beyond that?

ceremcem:
Note: I think this is highly related with #3086 

[...]

PR #3106, opened 2017-10-10
Proxy elements - or a marriage of decorators, parser transforms, and partials - covers lightweight, dynamic, and async components - RFC

evs-chris:
## Description:

This adds a new VDOM construct tentatively called a proxy. A proxy is simply a placeholder that gets a context handle that is extended with a method, `refresh` that triggers the proxy to unrender and render again. Proxies look like regular elements in the template, kinda like components look like regular elements. They even automatically get access to a content partial that is the content of the proxied element. Since the template on the proxy is effectively dynamic, you can achieve two common requests for components - dynamic and async - with a proxy.

Also, parser transforms are removed.

### Details

__This is outdated... see 4th post below__

For this PR Ractive has grown a new registry called `proxies` (open for suggestions on the name). It has also grown a new static method designed to create a proxy element, conveniently named `proxy`, which takes an init hash like a component would with `extend`. The only _required_ param in the init has is a `proxy` function, which receives a handle object that is a context with a few bonus methods, notably `refresh`, which causes the proxy element to rerender. The `proxy` method is expected to return an intermediary instance that is used for communication between the proxy element and the proxy class, kinda like a decorator.

```js
const thing = Ractive.proxy({
  proxy(handle) {
    return { template: '{{>content}}' };
  }
});

new Ractive({
  proxies: { thing },
  template: '<thing>I will just render this string</thing>'
});
```

That is a proxy that does nothing but render the content passed to the proxy. Not particularly useful, but it's about the simplest example of a proxy that could possibly exist. Since components are checked before proxies, here's what a simple async component proxy would look like:

```js
const Async = Ractive.proxy({
  proxy(handle) {
    const middle = { template: 'loading...' };
    loadAComponentSomehow(handle.name).then(cmp => {
      handle.ractive.components[handle.name] = cmp;
      middle.template = handle.proxy.template;
      handle.refresh();
    });
    return middle;
  }
});

new Ractive({
  proxies: {
    ComponentA: Async,
    ComponentB: Async
  },
  template: '<ComponentA /><ComponentB>content...</ComponentB>'
});
```

### CSS

Since I also want to target lightweight components as proxies, they also support Ractive's scoped CSS, including CSS functions. This means that you can use the ractive-bin to create lightweight components that are roughly as expensive as partials with scoped CSS in the single-file format. This is actually the main reason that there is a constructor function on the Ractive constructor - so that the styling can get set up and to grant access to `thing.styleSet(...)`.

```html
<div class="material-card"><!-- probably some more structure to this, but you get the point -->
  {{>content}}
</div>

<script>
  const template = $TEMPLATE;
  export default Ractive.proxy({
    noCssTransform: true,
    css: $CSS,
    proxy(handle) {
      return { template };
    }
  });
</script>

<style>
  .material-card {
    /* all of the material card styles... */
  }
</style>
```

### Questions

__Question 1:__ I was pondering an `update` method for the intermediary, kinda like decorators get. There's already an `invalidate`, which if supplied, is called any time the bit of VDOM containing the proxy is told to update. In order for `update` to work, the proxy would have to define `attributes` that it wants to keep to itself e.g. `<Async name="ComponentA" />` could say `attributes: ['name']` and get an additional attrs object in the `proxy` method and also any time those attributes updated, an optionally supplied `update` method in the intermediary would be called, allowing `<Dynamic name="{{.type}}" />`.

Note, since the handle is a context object, you can already set up context-aware observers and the like. You can't, however, use Ractive's nice expression support to allow just about any computation to be passed to the proxy e.g. `<Something foo="{{.bar * 10 + 2}}" />`.

So, should I add `attributes` and support for an `update` method?

__Question 2:__ Should any unclaimed attributes be supplied in an additional `{{>extra-attributes}}` partial like the one provided to components?

__Question 3:__ Along those same lines, the content of the proxy tag is supplied as the `{{>content}}` partial (and yielding means nothing in this context, because it's already in the same context as the surrounding template). Should the entire proxy template also be supplied as maybe `{{>initial}}` or `{{>outer}}` or `{{>outer-content}}` or something like that?

__Question 4__: When added to `attributes` and `update` support, any attributes claimed by the proxy would be removed from the outer template partial. Does that sound useful, or at that point are you going to be messing about with template AST anyway, so no real benefit?

__Question 5:__ Any suggestions for better names? I've also considered placeholders, transformers, lightweights, and probably a few more, but I don't know that any were better. Proxies feels a bit... strange, I guess.

#3099 #3089 and probably some others.

## Is breaking:

Yes, in that `proxies` and `proxy` are added to Ractive constructors. Parser transforms are also removed in this because proxies provide the same approximate service in a more flexible, runtime-friendly way.

## TODO

* [ ] Tests
* [ ] Finalize the names
* [ ] Advanced attribute support?
* [ ] Allow for proxy registries?
* [ ] Allow for proxy inheritance hierarchy?


evs-chris:
Oh yeah... the `template` on the intermediate uses partial resolution, so you can specify script tags by id or strings to be parsed as templates if you have runtime parsing - or you can directly reference parsed templates. It's generally best to avoid runtime parsing, and the ractive bin is currently the best way to do that in general - and the only way to do that with this construct, as I believe rcu/rvc and friends are tied to components only.

ceremcem:
@evs-chris Is there any way to test it out via Playground or something?

evs-chris:
After sleeping on it, I have another question: should proxies also have the ability to contribute to registries e.g. provide decorators, transitions, and friends? My gut says yes, and we may even want to be able to create an inheritance hierarchy with proxies in a similar manner to components at some point.

To allow that at some point, I've adjusted the `Ractive.proxy` function to take a function as a first arg followed by an optional options hash. The simple example from the initial post then becomes:

```js
const thing = Ractive.proxy(handle => ({ template: '{{>content}}' }));

new Ractive({
  proxies: { thing },
  template: '<thing>I will just render this string</thing>'
});
```

Any options like `css` would be supplied in a hash after the function e.g.

```js
Ractive.proxy(
  handle => {
    // do stuff here
  },
  {
    cssId: 'my-proxy',
    css: '.my-styles { color: pistachio; }'
  }
);
```

Back around to registries, that would also open up the possibility of blocking proxy names from within the proxy to prevent infinite recursion e.g. with a `a` proxy that renders an `a` element. Having `Ractive.proxy(..., { proxies: { a: false } })` would stop Ractive from using the `a` proxy in template contained by an `a` proxy in much the same way that components can prevent recursion.

evs-chris:
If anyone wants to try out the current iteration of this PR in the playground or somewhere similar without having to build it, I've published it as `@evs-chris/ractive@1.0.0-proxy-001`, so you can pull it in from `https://cdn.jsdelivr.net/npm/@evs-chris/ractive@1.0.0-proxy-001/ractive.js`.

Note that while I have tested it in the sandbox, _actual_ tests are still on my TODO list pending feedback.

evs-chris:
[Here's a playground](https://ractive.js.org/playground/#N4IgFiBcoE5QdgVwDbIDQgM5QNrgC74AOmkA9GQMYAm8AdAFabUCmyAlgG4x3wv5l4RALZkAAi06YAtJTAx2mMjACGlfFxZiAjHQAM+6URgB7AB4BPaXr0BmZWo2cWjbAF0AvhnxQQAHjBtAD4AcRY+BUoAAhVMC3hoyhNhIhM+eHwo43MLPzJAoIAdeD9qLij2agBeQpB8FRgAc35aoLyyziL4Yr9MSgUiTPwLIhYaupYzATB8YWRaiura-BYU5BUV1uLC-AAJFhgWAHJMGLiEqIBhZNT0-ABBSG38P3vz6Nj4yml4FWEx2rXFJpcIPBYAMxMJnGwGAdAARg0PB4tt18DsACpgRQVU74MAsKJJDKgqKQmBRfGEpLAu50Z47PzIEwqMrwRpBACSUUwYBMAHcovzschCczWex2XRpXlxWyOc88m8vl08n0BvggiAMNhICBiZhModGop8DALFEqlFgB4ANzbBJpQ1RZUXK0AJUcmjo2UsAAowCp4NRRZagtbngbMr6LVabfa0VGoiZECsKVbA8HRT7TJY6Cs1hsWAmdknfv9LcnUwc6MI6ODJdQ-SowzFeJaqlajp8Ej8-scAJT1kvotGi6PrSgsPnIVjpqJ+lNp+tRAA+q6iODcQ4bwb9hKq4ZcB87USOcsljSOA5H7HBC6Ik+nJlnBwHWSfM7nlcfamfr54cERzYTBCV-KcvwOSscHPFl5WlOgjjcW0ogoKJWHBFQUAnP9IJgBlEydIZH2g4AohYSAonLFg0EpSilxrfBaOEejqx4OsG2QNNm1bFR2wAQlPbt3j7f5r1o8FWOXcF63YLiDn3VtjyiQSuwvdlryiDxkIInY7wXT11G9XdqAABWQRATXgP1qNoo4aVuUFMCOWjMxDFxVCM5wB3fYAGWIXMLHzVZfxWStiGQEcURWZBQIjNEdhjYLCzCq1wP-OcRx2fy5SBRyMhs-shypazKBSVs-IS9F8SDdy6E8pwXAckEMkwHBqLcSsyqILLqtYcdCQYngiDoYkVgyXrEsC5LQoPSlH0mmqsw8lhwUOXk-RvfyPC2tEUQStFDnwRAYHgLJAoTLxrR2kt4HBRAEg0NIolym4WvwQr-l854jpOs6+EFUzTGERQWEXABrCr-NA-AMXYf4lz9TaoaqnYTHBv1DMauhJnGptKuy6qdgLWbKIAAwCWwgn2VATDJYGiTeu4zwAEmAaiPCOPIwCpqIAE0Ux5FRKkpQNMmEC0jkhEwjlxKJalhaXkVqSmggAdTFskX2ZflThxCXGZJDJIG5qnYQsdg2GoZEyf8nYdt2wmroAVhsPRHfwG7in24oAaiLHND9AnCAaZp8Eoo4AGJ6iafgXOeEn1hWCPo5CpPjjQZ5fUt0hrRdd4tMztFqA2FRKODnZERgCPuT+KIq5iI0oXwI5ni8SMmac8v-IoPL3sef2vWcHGpnCfHKTTosI4Lvu7nuKJpGtYAlY8Px4RgMggnNy3Z2RWWdtQgAqKIHppf4MlFnF8Dp0DCWFDYokDIhRngPX7ypQ4ThiZBDlZC0W1nqCeeKhODC3WPCUMh8yBt29jeEAHggA) with a moderately featured async component proxy that pulls its loading message from an optionally supplied `loading` content element that is not passed through to the loaded component. Any attributes other than `name` are also passed through to create mappings on the target component.

There's also a synchronous component in place and commented out so you can see how the async proxy behaves if the component is already loaded.

__Note:__ I found a few bugs while playing, so the version is now `1.0.0-proxy-003`.

evs-chris:
I've added claimed attribute support for proxies, which means you can specify that the proxy will consume certain attributes by name, and if they're dynamic, you can supply a callback to be called whenever they change. They are also passed in on the intermediary creation call.

```js
const Proxy = Ractive.proxy((handle, args) => {
  // args is a map of attribute values that are claimed in attributes below
}, {
  attributes: ['name']
});
```
```html
<Proxy name="{{someName}}" class="foo" />
```

There, `name` will be claimed by the proxy, so it won't be present in the template available on the handle or in the now-supplied `extra-attributes` partial. Everything else is though - in this case `class="foo"`.

[Here's](https://ractive.js.org/playground/#N4IgFiBcoE5QdgVwDbIDQgM5QNrgC74AOmkA9GQMYAm8AdAFabUCmyAlgG4x3wv5l4RALZkAAi06YAtJTAx2mMjACGlfFxZiAjHQAM+6URgB7AB4BPaXr0A2ZWo2cWjbAF0AvhnxQQAHjBtAD4AcRY+BUoAAhVMC3hoyhNhIhM+eHwo43MLPzJAoIAdeD9qLij2agBeQpB8FRgAc35aoLyyziL4Yr9MSgUiTPwLIhYaupYzATB8YWRaiura-BYU5BUV1uLC-AAJFhgWAHJMGLiEqL92IURM9YAvCyjOFWREMeBgShSAORVhFgeDxRMhBSDbfB+ACC52isXilGk8H+Y1qn2+RD+AKBCwAZiYTONPnQAEYNHEgLo7HYAFTAigqp3wYBYUSSGXCmXxMCizNZSRSaU5dAhO0+AGIiA0NK8oviTECAJIxYS8llZWKYFjUOUEz5kKUwGXIHHdfA7PzIEwqMrwRpBZWYMAmADuURd9OQrKtNuujToAbyPtt9oheRhCK6eT6A3wQRAGGwkBA7MwmUOjUU+BgTyqUWAHgA3NsEmk01EIxc8wAlRyaOjZSwACibYBU8GoXrQMSamAAlFEqkF8xDU5lG7n80XR2XMiZbgdB1E2x2vXQVmsNixi2aIV7x+tKCxnchWDy8035yseLiogAfO9RHBuPt0XHXahN1lDqIub9VPMjmDP0jj7HcdnYW8myIQ9jxMU8DgHGC1DghDzyyWCTzPN9wJWZAtQwlCsMXPMcCA60QwDOgjjcQsQTIKJWFxFQUAPIj4LPUUzQnOhECIagtyXXFEASDQ0iiQ5gybDZs0wbt3xgNMB2AUV8DHKJkQBJcZMUsj4QSJEURo3CdnU4hkCXHBgF-SANJRbt8Fsq8DnXbthCchceGEeSPOvN833YZBry-Qdhz-KIAEIAKicjfTtUDuyIWyV07FxDWNU4PFo1TVMgqIm1rdR63fDsAAU3kzeAm00lhuyOAVUnSfBMCObsUrXVQiucPtlJy4hTEsddVmQlYl3Mkz8A8HY2AIlTd3NfqciGzdRrzZCj2ImAJupeadmDABhZJGs5aqUVfPkqoxUKR12hb2pcTqnBcBqhQyTAcBqtwlwxbaFp4jcRu-XkYN+nZ7roQ5cUOJ0mzAvqPDh27VKm3b5rypsIoUpTl3bVKIZYKGWBhxGdmnXduIGixeP4rdpN7ByYHeEnzTNQ58EQGB4CySmdy8G6dh09gSQXUgnyOfTERqmjigR8D4GE0T2HEg6jte-BToBXrWf4Dmub4N1StMYRFBYS8AGtrrm6lmv4Gl2ABK8WwHH8rYWnYTDNgq62cOhJhWDsm1d62dgB9YVlsgADAIAGYgn2VATDlI22VVpqYoAEmAGqPCOPIwFjqIAE15yiTAVEqNUNiiYQniOeUjkZKI0WAeUKRjoIAHU2y5eCrRdU4GRrlOOQySA89jz4LHYNhqCBUouCCSfp9PXUFQ8doF4j+GeomvmAFYbD0ZnZZlkt9aiQqnsDiF6iafhbKOcVb+afBWpv4aw5YB+n4-rc34p8w09RbWUrNELwEIBL1FskHfAZIYAP2VP8KIcCYjpgJK-NAqkMRYi-jFQ6gompQiOBCcBZoXpNWAapCg+DjoZChLZS+9Y-bhE-NZUOW4H6wiiDQtWUIojSHzC3Akc8SQwFBEvGeQIG4IxBAAKiiCJAUAIMhqgZPgROWpWQeirm2Igox4AD1vHyQ4JwYjIEODaJ4KhuGp05HwlQLxAoqBJF6KIsiyAkJlmBEAHggA) the async component example updated to make the loaded component dynamic based on the `async-name` attribute. By changing the name in the input, you will cause a different component to be swapped in. The first time a new name appears, it will go through the mock loading cycle, and any subsequent appearances will be synchronous because it's already available.

ghost:
Been off the grid for 3 days 🎉  

Here's an example of the minimalistic API I was thinking of based on the comments from previous discussions (like https://github.com/ractivejs/ractive/issues/3089#issuecomment-333712717). The gist is that an async component should just be like any other component, albeit loaded and rendered differently. Underlying machinations may be different, but at least authoring experience shouldn't cause a drastic change.

```js
// AsyncComponent.js - write in vanilla or component file format
export default Ractive.extend({ ... })

// RegularComponent.js - write in vanilla or component file format
export default Ractive.extend({ ... })

// App.js
import RegularComponent from './RegularComponent'
import Loader from 'some-loader'

// Loader method that returns a function that
// 0. Is called by Ractive when it needs the component.
// 1. Returns a promise.
// 2. Loads the component (Ractive doesn't care how).
// 3. Does some pre-processing (Ractive doesn't care if it does).
// 4. Resolves the promise with the constructor.
// 5. Resolution causes any existing usages to refresh/toggle/magically render
// 6. Ractive caches the constructor, as if it were defined in `components`.
const AsyncComponent = Loader.load('AsyncComponent.js')

const App = Ractive.extend({
  components: { RegularComponent },
  asyncComponents: { AsyncComponent },
  template: `
    <RegularComponent />
    <AsyncComponent/> <!-- this would re-render if it was already rendered -->
  `
})

App({ target: '#app' })
```

> should proxies also have the ability to contribute to registries e.g. provide decorators, transitions, and friends?

In the above case, since the async component is written just like any other component, I would expect that plugins would still be possible and follow existing rules (i.e. plugins defined in components are component-scoped), CSS would be re-evaluated (i.e. async component CSS appended), that sort of stuff.

My 2c

evs-chris:
My main issue with async components as a built-in like that is that there is no way to show a loading placeholder while you wait on the promise to resolve. With `<Async component="MyComponent" other="params">...</Async>`, you get the opportunity to throw in whatever you want for a placeholder - including nothing if that works for you. A cherry on top - having css scoped for the proxy means that you can also style your placeholder independent of the loaded component :).

That's also ignoring the impl details, which would effectively be building an async proxy plugin directly into ractive to handle the impedance mismatch between the sync render and async load.

ghost:
I was thinking of using the component innerHTML but I guess we have plans for using it for yielded content. Hmm... conditional based on the registry and the resolution? 😁 

```
{{async MyAsyncComponent }}
  <AsyncComponent />
{{else}}
  <MyLoadingAnimationComponentThatIsAlreadyThere />
{{/async}}
```

(elegance just left the building, hahaha 😁 )

monoblaine:
What I'm going to say is probably almost off-topic, but I'm concerned about the naming. Will using the name "proxy" create confusion regarding the *proxy events*?

evs-chris:
Definitely not off topic - that's one of the main questions to resolve before merging this 😀

ghost:
Referencing related implementation of stuff:
- https://github.com/sveltejs/svelte/issues/640
- https://github.com/sveltejs/svelte/issues/742

PR #3132, opened 2017-11-09
Computation children should adapt too - fixes #3130

### LINKED ISSUE #3130: Adaptors not called on items of computed arrays
Hello,

I'm currently upgrading an application from Ractive 0.7 to Ractive 0.9.9.  
I've noticed a behavior change in the way Ractive handles arrays of external objects returned by a computed property.

## Description:

When an computed property returned an array, Ractive *used* to adapt every item of the array. 
If the array returned by the computed property contained external objects, the corresponding adaptor were used, and the external objects could be user in the template.

This behavior has changed in newer versions of Ractive: when a computed property returns an array, Ractive doesn't check wether the arrays contains external objects. The external objects in the arrays are handled as if they were POJO. 

This change beaks several of my Ractive components. How can I workaround the problem?

Thanks :-)

## Versions affected:

It affects Ractive 0.8, 0.9, and the "Edge" version.

## Reproduction:

Here is an example of code reproducing the problem ([Playground](https://ractive.js.org/playground/?env=docs#N4IgFiBcoE5SBTAJgcwSAvgGhAF3gDxICWAbgATFIC8AOngIYxq70B8BA9CaW7QHYCCAZwDGMYgAdclGvVwIAtpIA2DBfXK4AnpIR08CAB65OMBqNxkE7AeXIFRAeyQI+-e-f5PcAYSfKAK4KSACCMObakOTAwN5+AZLByOGRAHQADGn8DIoIGBgEAEYw5JzunuTOQSGpDFExwNVJtRH1mdm5+YUlZRWezclIAMqiDGow0bGDIaPjTJ15BXZczq7uXGIS0mwgOMLwnJzkoeTCxMoqCOQAUgykDKPbMqJqwsICrwzv5AAi6gwYnYqk5+MJcDBApYnDAABQ5PIASiBHk8uDAxGEi2u1HICIQAG5gRhgbgnMMIcR+ChYcjgMD7DAELhAjAPAADf64QHAPFdaIAEmA6Mx2Iw5Aw7KJqJJ-FlAiOJw8DCQDGkMPIADMNejrlzAV93p9QeC-gDQqr1aVcfTUZriCoFJNyA8VIEcWwUZUmSy2S7xu7yAAyIP+t3XKnghj8UQIJyas3c6X2bDAgDu5kk0VhzVBCH4uCw5FV3OR1E9tsqzjBTiuaVcRUCNM5APIKrVZNKYxUVyQ5FB5CFJYYksRyc8PtZHkrlXILFpXtnjOZU8XS68-OLAOx5AA1ORNLD29JkIj6AzZxhx5VU4JUd6EMJmQuZ0vJ37NeMn9fPLeL-Yn1wF9byXBQmCQJw034YCLyvYkBDgu9FVOAAlCwrFIa5c34fMZECc5qS0MBrmPTtjTBGRzEsaxyFxHC03INDqMw2EZwQFRonoABiblmGZegsFJJRVHUBBOJAHjhLUDQ9mBUjogAbX1C0OxhABdQTUWHKYL3ifwahSNoGgU-88QQBj9Vheg9MSIY6m0cgAEZ6ERTSl3oxMGCskAbIMsIjPIAAmFy3NnDzLOsnx9JaQzInIABmFyLzU4lQpmZAdPvEE-Ps7My09Eysq8czPO89L-Li5yQFc0zwoBMrbNaOLguq0LKjq7kGpygLEuq5K2uymKRm7Jg8toz0Oq8+hyrmCYkplBCx0wIA)):

```html
<div id="target"></div>

<script id="template" type="text/ractive">
  <code>
    notComputedArray: {{notComputedArray.0.name}}<br />
    computedArray: {{computedArray.0.name}}<br />
    computedScalar: {{computedScalar.name}}
 </code>
</script>
```

```js
// A simple JavaScript class
class Data {
  constructor(name) {
    this.name = name;
  }
  toString() {
    return `Data { name: ${this.name} }`;
  }
}

// An adaptor for the Data class
const DataAdaptor = {
  filter: value => {
    return value && value instanceof Data;
  },
  wrap: (component, data) => {
    console.debug(`Data adaptor called on ${data}`);
    return {
      get() {
        return {
          name: data.name + " (adapted)"
        };
      },

      reset() {
        return false;
      },

      set() {},
      teardown() {}
    };
  }
};

// A Ractive component using the adaptor
const ractive = new Ractive({
  el: "#target",
  template: "#template",
  adapt: [DataAdaptor],
  data: {
    notComputedArray: [
      new Data("notComputedArray 1"),
      new Data("notComputedArray 2"),
      new Data("notComputedArray 3")
    ]
  },
  computed: {
    computedArray: () => [
      new Data("computedArray 1"),
      new Data("computedArray 2"),
      new Data("computedArray 3")
    ],
    computedScalar: () => new Data("computedScalar")
  }
});
```

Using Ractive 0.7, the example renders the following output:

```
notComputedArray: notComputedArray 1 (adapted)
computedArray: computedArray 1 (adapted)
computedScalar: computedScalar (adapted)
```

With the latest version of Ractive, it renders the following output:

```
notComputedArray: notComputedArray 1 (adapted)
computedArray: computedArray 1
computedScalar: computedScalar (adapted)
```

evs-chris:
## Description:
This puts the adapt step back into refreshing a computation child value, as it probably got lost somewhere in the massive model shuffle between 0.7 and 0.8.

@fskreuz I _think_ this is the correct way to handle this, but it's been a while since I've spelunked that part of the code.

#3130

## Is breaking:
Shouldn't be.

### DIFF
--- a/src/model/ComputationChild.js
+++ b/src/model/ComputationChild.js
@@ -40,6 +40,8 @@ export default class ComputationChild extends Model {
 			this.dirty = false;
 			const parentValue = this.parent.get();
 			this.value = parentValue ? parentValue[ this.key ] : undefined;
+			if ( this.wrapper ) this.newWrapperValue = this.value;
+			this.adapt();
 		}
 
 		return this.value;
--- a/tests/browser/plugins/adaptors/basic.js
+++ b/tests/browser/plugins/adaptors/basic.js
@@ -903,4 +903,37 @@ export default function() {
 
 	});
 
+	test( `children of computations are also adapted (#3130)`, t => {
+		const arr = [ 'hello' ];
+		let count = 0;
+
+		const Adaptor = {
+			filter ( v ) { return typeof v === 'string'; },
+			wrap ( ractive, v ) {
+				return {
+					get () { return v + ' adapted'; },
+					teardown () { count++; }
+				};
+			}
+		};
+
+		const r = new Ractive({
+			target: fixture,
+			template: '{{#each list}}{{.}}{{/each}}',
+			adapt: [ Adaptor ],
+			computed: {
+				list: {
+					get () { return arr; },
+				}
+			}
+		});
+
+		t.htmlEqual( fixture.innerHTML, 'hello adapted' );
+
+		arr[0] = 'still';
+		r.update( 'list' );
+
+		t.htmlEqual( fixture.innerHTML, 'still adapted' );
+		t.equal( count, 1 );
+	});
 }


src/model/ComputationChild.js
```diff
@@ -40,6 +40,8 @@ export default class ComputationChild extends Model {
 			this.dirty = false;
 			const parentValue = this.parent.get();
 			this.value = parentValue ? parentValue[ this.key ] : undefined;
+			if ( this.wrapper ) this.newWrapperValue = this.value;
+			this.adapt();
```
ghost, line 44: As far as I know, adaption happens on set. Not sure how computations work yet. 😕

tests/browser/plugins/adaptors/basic.js
```diff
@@ -903,4 +903,37 @@ export default function() {
...
+			}
+		};
+
+		const r = new Ractive({
+			target: fixture,
+			template: '{{#each list}}{{.}}{{/each}}',
+			adapt: [ Adaptor ],
+			computed: {
+				list: {
+					get () { return arr; },
```
ghost, line 926: I also find this a strange case, where a computed gets data outside the instance. If the data was set on the instance, they'd probably have adapted values by then.

But then again, I'm not familiar with computation internals. 😁

src/model/ComputationChild.js
evs-chris: For computations and their children, they get notified any time they change, including on `set`. They then set a dirty flag so that next time a dep asks for the value they know to pull a fresh one. The `Computation` is handling the adapt correctly on`get`, which is when it first knows its actual value, but `ComputationChild` was missing that.

tests/browser/plugins/adaptors/basic.js
evs-chris: Just a dumb case to fake a computation :smile:. It doesn't matter _how_ it's computed for the test, just _that_ it's computed.

src/model/ComputationChild.js
evs-chris: I left out a crucial bit: all computations are lazy, so they don't know their value until they are asked for it.

That means that `get` acts kinda like `applyValue` from a regular model too, and `set` calls the setter, invalidates the model, and waits for a dep to ask for the new value before actually updating.

ghost: > all computations are lazy

Aha! That makes sense now. 😁

ghost (approved):

m-r-r:
Hello,
I downloaded the latest version and now it works :-)
Thank you for your help!

Issue #3137, opened 2017-11-14
Children of computed properties should be unwrapped by `ractive.get()`


m-r-r:
Hello,

I've found another breaking change between Ractive 0.7 and the latest version.
I forked the repository and I corrected it. I plan to submit a pull request ASAP.

## Description:

With the latest version of Ractive, calling `ractive.get()` on a child of a computed property returns the wrapped value if the property is using an adaptor.

The method `ractive.get()` in Ractive 0.7 used to return the unwrapped value, even when it was called on computed properties.

Additionally, I noticed that the `unwrap` option (added in Ractive 0.9) is ignored when the method `get()` is called on computed properties.

## Versions affected:

All versions from 0.8 to 1.X.

## Reproduction:

[...]

PR #3138, opened 2017-11-14
Children of computation should be unwrapped by `ractive.get()`

m-r-r:
Hello,

This pull request fixes the issue #3137.

## Description:

- The value of computed properties is now correctly unwrapped
- The `unwrap` option is no more ignored for computed property. I don't know why it was ignored before, I guess it was a error.

- Issue #3137.

## Is breaking:

I hope it doesn't be breaks anything :crossed_fingers:

### DIFF
--- a/src/model/Computation.js
+++ b/src/model/Computation.js
@@ -40,7 +40,7 @@ export default class Computation extends Model {
 		if ( this.signature.setter ) return this;
 	}
 
-	get ( shouldCapture ) {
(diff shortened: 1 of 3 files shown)


ghost:
👍 Looks good to me

evs-chris:
Thanks for the report and fix! I'm looking to cut 0.9.10 in the next day or two, and this will go out with it. Until then, it should be available on edge and v0.9-dev as soon as Travis does his thing.

Issue #3120, opened 2017-10-27
Supported API to add computed dynamically


Jeff17Robbins:
## Description:
We are close to general release of a healthcare app that uses Ractive.  We're looking forward to a Ractive 1.0 official release, and are hoping to remove the **one** undocumented feature we are using:

```ractive.viewmodel.compute()```

We need to add computeds that on-the-fly to an already existing Ractive instance.  It would be great if there were a supported API to do this!

## Versions affected:
1.0

**Comment by ghost** [comment]
It's so undocumented, I wasn't even aware this exists. 😁

**Comment by Jeff17Robbins** [comment]
Yeah, @paulie4  and I were trying to remember how we learned about it.  I think there is an old conversation about it from someone else asking for the same feature.   

If anyone is curious, our use-case is that we want to sort some data that arrives from the server.  We started out using a helper function to do the sort -- this works great in a template.  But...not so great in JavaScript `ractive.get()`.   Whereas a computed works great **both** in templates and in `ractive.get()`.

**Comment by evs-chris** [comment]
You can actually get a handle to a template expression using a context pretty easily, and it's very nearly the same as a computed. Not that adding computeds dynamically isn't something we should support. It just wasn't possible until 0.8, I think.

I'll see what it would take to make an official API for 1.0. I don't think it's quite as simple as exposing the viewmodel compute, as that wouldn't handle transferring any existing deps to the new model.

**Comment by Jeff17Robbins** [comment]
I'm unfamiliar with getting a handle to a template expression.  Happy to
move to a different feature if it does what we want and is official!

As for "transferring any existing deps to the new model"...well...if I knew
what that meant I bet I would feel smarter!  :-)

**Comment by evs-chris** [comment]
For the context handle, if you have something like

```html
{{#with some(computation)}}
<div id="mah-expr">...</div>
{{/with}}
```

you can `get` stuff from the expression by getting the `div`s context with `var ctx = Ractive.getContext('#mah-expr')` and using `ctx.get('.exprProp')`. This works because the context for the `div` is the expression model that's created in the background and owned by the template.

Jeff17Robbins:
Made a fiddle to try to understand http://jsfiddle.net/vt76xm54/

HTML
```
<div id='target'>

</div>
```


JAVASCRIPT
```
var ractive = new Ractive({
  template:'{{sort(items)}}',
  target: '#target',
  data: {
    items: [ 2, 10, 200, 3, 1, 4],
    sort: function(arr) {
[...]

Issue #2748, opened 2016-11-06
Could the Computed Properties Compact Syntax be improved?


Jeff17Robbins:
## Description:
The syntax of Ractive is getting very clean and regular.  The Compact Syntax for Computeds looks increasingly lonely as something irregular to learn. 

As an enhancement idea, could the Compact Syntax for computeds be upgraded to support an "expression", more like how Event Methods are now event expressions?

As an example, could 
```
  computed: {
    area: '${width} * ${height}'
  }
```

be changed to

```
  computed: {
    area: 'width * height'
  }
```

I know that this analogy isn't 100%, but, in spirit, a template can do this (from [doc](http://docs.ractivejs.org/edge/method-calls)):

```
{{#each items :i}}
  <button on-click='@this.select(this,i)'>select this item</button>
{{/each}}
```

I'm thinking that if Ractive can "know" that `i` is part of the evaluation context, so too it could know in the case of a Computed that `width` and `height` are part of the data context, without recourse to the special `${...}` markers.

## Versions affected:
edge

## Platforms affected:
all

## Reproduction:
N/A

## **Code:**
N/A

evs-chris:
That's probably a reasonable thing to do, especially if it saves space on some special-cased parsing. It would also make template strings tiny bit less confusing when they eventually make it in to the parser :smile:

dagnelies:
+1 from me. The `${...}` always seemed out of place to me.

And as usual, first only a warning ;)

-----

Besides of this, I'd like to point out a design flaw in computed variables: they are "lonely".
Let's say you have a list of boxes: 
```js
data: {
   boxes:[ {width: 1, height: 2}, ...]
}
```
In this case, computed variables become useless.

Extending it to keypaths would be rather nice in the future:
```js
data: {
   boxes:[ {width: 1, height: 2}, ...]
},
computed: {
  'boxes.*.area': 'width * height * @index' // (scope would be 'boxes.*')
}
```
...but that would be a whole other topic, where care should be taken that it does not clash with observers and components. IMHO this would be a very low priority feature though.

aharpervc:
What about mustache (or mustache-lite) syntax, the same way you would in a template?

dagnelies:
@aharpervc  I had this thought too. However, there is subtlety in it:

[...]

Issue #1583, opened 2014-12-16
Expression Functions and Computed Properties


dagnelies:
EDIT: TL;DR

Goals:

- allow computed properties keypaths like `foo.*.bar`
- provide current scope in the computed callback
- allow functions inside of `computed` or `expressions` (to avoid placing them in `data`)

``` javascript
computed: {
  'foo.*.bar': function(scope) {
     // cool, all "foo" items have a computed property "bar" 
     // and can access the data of the current foo!
  },
expressions: {
  pkg: {
    myFunction: function(arg) {
      // cool, an expression function to be used like this: {{ pkg.myFunction('abc') }}
      // That's a better place than in "data"! :)
    }
  }
}
```

----

Hi,

As you know, these are _relatively similar_. At least, they serve the same purpose. It's to compute something a little more fancy than trivial expressions.

However, they differ in some ways, which have their pros and cons. The cons are mainly:

**Expression functions**
- the functions are mixed with the data

_Call me old-fashioned but I like to separate my data and my processing functions ...especially in our AJAX world where raw json data tends to be the currency. It's rather annoying to "boost" your received json with additional functions just to render trivial expressions, and strip it away when saving._
- no getters, observers or setters

**Computed Properties**
- you cannot give them parameters

_which means you can't use it in lists of items!!!_
As shown even in the doc:

``` javascript
new Ractive({
  template: '{{#boxes}}{{area}}{{/}}',
  data: { boxes: [ { width: 20, height: 40 }, ...] },
  // there's no way to specify this for "each" box :(
  computed: { area: '${width} * ${height}' }
});
```

The question is: can we merge these in order to have the best of both worlds? Ideally, in a way that is compatible with the existing stuff.

**Option A**
Move expression functions out of `data` into a better place. You would still not have getters/observers/setters but in would already be a great step forward IMHO.

**Option B**
Be able to declare computed values for arbitrary keypathes _and_ provide the getter and setter a local scope object. Or at least a keypath. So that declaring properties would become:

``` javascript
computed: {
    'boxes.*.area': function(scopeOrKeypath) {
        // yeah, each box can have a computed area property!
    }
}
```

There are probably ways to make it nicer and more practical, but at least, it would enable to add properties to list items.

ceremcem:
@dagnelies How about the recursive use-cases? For example, boxes might contain some other boxes and we might want to display the `width`, `height` and the `area` properties of the boxes in a tree-view, like so:

```js
...
data: {
  boxes: [
[...]

Issue #956, opened 2014-07-09
Functions outside data


ghost:
[In this example](http://learn.ractivejs.org/list-sections/5/), the function that alters the array is inside `data`.

``` js
var ractive = new Ractive({
  el: output,
  template: template,
  data: {
    superheroes: xmen,
    sort: function ( array, column ) {
      array = array.slice(); // clone, so we don't modify the underlying data

      return array.sort( function ( a, b ) {
        return a[ column ] < b[ column ] ? -1 : 1;
      });
    },
    sortColumn: 'name'
  }
});
```

However, if I did something that would alter `sort`, it will run over `sort` in the data and would break the template. [See this demo](http://jsbin.com/cinimuha/1/edit). Same happens [when using reset](http://jsbin.com/riqewopi/1/edit), which I often use to clear off stuff, especially when reusing modals (clear, dump data, show).

Would it be better if these "helper" functions were separate from the data? Like a `helpers` per instance, and `helpers` in global, similar to partials.

``` js
var test = new Ractive({
  helpers : {
    foo : function(){...}
  }
});

Ractive.helpers.foo = function(){...}
```

PR #3141, opened 2017-11-15
All the computed changes - WIP - RFC

evs-chris:
## Description:

This addresses a handful of requests involving computeds starting from at least two years back. It also introduces a number of breaking changes, so it should probably land in a version greater than 0.9.

This PR is not complete. There are a number of open questions, mostly related to nested computations that need to be answered before the test suite can be filled out around all of these changes.

### Helpers

As outlined in a number of issues, many devs simply don't like having functions in their data. I personally find it liberating to be able to call functions that I throw in there and pass around, but I can see where it would be unpleasant for some. To address that, this introduces another registry that starts with `Ractive.defaults.helpers` and is inherited by subsequent extensions and instances. This registry can technically contain anything, but it's meant to hold helper functions, like `upper(str)` and `format(date, 'YYYY/mm/dd')`. The helpers registry is added into the resolution process after prefixed references and before context traversal, so if you have a reference with a single unprefixed key, it will resolve in the helpers registry before looking through data and contexts.

This is only breaking if you happen to have a `helpers` key on one of your components, but otherwise, you can only break stuff by adding to the `helpers` intentionally.

### Nested computations

This feature adds the ability to have computed members at an arbitrary keypath depth - not just at the root. It does so by breaking the current behavior of computed paths e.g. that `foo.bar` as a computed is available as `foo\.bar` in the model and makes it available as the nested `bar` key on the `foo` object - if that object exists.

As it stands currently, the computed registry is only checked the first time a keypath is requested. If there's a match, subsequent requests will turn up the computed and if not, they'll turn up the plain model that maps to the actual data, which may not exists. I think there should probably be a knob to adjust that behavior a bit, like a computed that only applies if there's no underlying value and a computed that _always_ shadows the underlying value.

Building on nested keypaths, this also introduces wildcard computeds, which are similar to pattern observers but with fewer limitations. A single `*` in a computed path will match _one_ key and a `**` will match multiple keys, so `foo.*.bar` will match `foo.1.bar` not `foo.1.baz.bar` and `foo.**.bar` will match `foo.1.bar` and `foo.1.baz.bar`. Multiple wildcards can be used in a single path, including `**.bar` to make `bar` available on _any_ path. __Note__ that matching wildcards is not _terribly_ expensive, but it's also not as cheap as just matching underlying data directly. If you have more than a handful of these, it's going to start having some performance overhead.

Nested computations receive their parent value as the first argument to their getter, followed by their keypath and any wildcard matches. I _think_ it would be useful to give them a context object somewhere in there too, but I'm not sure where or how. I leave that as an open question.

Also open is the question of how to call the setter for a nested computation. Should it be the same signature as the getter, but with the target value first? Should setting a nested value also invalidate the parent model, as that's likely what would be mutated?

### Runtime addition of computeds

You can already add to the `computed` registry of a Ractive instance, but any existing references to the same keypath as the new computed won't use it until they are re-rendered. This adds a new method `compute` to the Ractive prototype that allows you to add a new computed at runtime and have any existing models that match the keypath transfer their deps to the new computed.

The answers to the questions for nested computations will also heavily affect how runtime computations behave. For instance, should there be a flag for wildcard computations that override existing models that match the path with a new computed instance at those paths?

### Stringy computed syntax

The current string syntax for computeds looks something like a string template, but it's definitely not the same thing. Instead the interpolation points in the computed string are turned into `__ractive.get('...')` and a `var __ractive = this;` is added at the beginning. That string is then set as the body of a `new Function`.

This PR drops the manual conversion of the computed string and instead leverages Ractive's expression parser directly. This means that references no longer need to be wrapped in interpolation tags `${ ... }`. A function is still created at runtime.

### Other stuff

As of 0.7, models would walk their constructor chain up to `Object`, `Function`, or `Array` looking for prototype property matches for child keys. This is problematic for the fairly common pattern of using `Object.create(null)` as the root of a prototype hierarchy, where there is no constructor. `Object.getPrototypeOf` is also sufficiently unsupported that it's not really a practical solution, so it's much simpler to just ad an `in` check and let the root prototypes contribute whatever they will too. This is breaking.

## Try it out

A build of this branch is available on npm as `@evs-chris/ractive@1.0.0-wild-compute-001` and https://cdn.jsdelivr.net/npm/@evs-chris/ractive@1.0.0-wild-compute-001/ractive.js.

#3120, #2748, #1583

## Is breaking:
Yep.

`computed`s with a `.` require escaping if they're supposed to have a `.` in the keypath. This is more consistent with the rest of the data anyway.

_All_ prototype properties are considered when checking to see if a model contains a value. This means that scoped refs are _much_ more important if you do things like `{{#each items}}{{length}}{{/each}}` where some items have a `length` and some don't. For those that don't, the `length` of the array will be returned because the parent context `items` _does_ have a _length_.

The string version of `computed` syntax changes to match regular old mustache expressions e.g. rather than `${width} * ${height}` you would write `width * height` for an `area` computed. This also opens up template string use in computeds e.g. ```'`the area is ${width * height}`'```. Notice the template string nested in a regular single quoted string. Ractive already supports template strings internally by having the parser transpile them, so the variable references are turned into ractive refs.

### DIFF
--- a/src/Ractive/config/config.js
+++ b/src/Ractive/config/config.js
@@ -22,6 +22,7 @@ const custom = {
 	computed: config,
 	css: cssConfigurator,
 	data: dataConfigurator,
+	helpers: config,
 	template: templateConfigurator
 };
 
--- a/src/Ractive/config/config.js
+++ b/src/Ractive/config/config.js
@@ -11,8 +11,15 @@ import RactiveProto from '../prototype';
 import { hasOwn, keys } from 'utils/object';
 import { isFunction } from 'utils/is';
 
+const config = {
+	extend: ( Parent, proto, options, Child ) => configure( 'extend', Parent, proto, options, Child ),
+	init: ( Parent, ractive, options ) => configure( 'init', Parent, ractive, options ),
+	reset: ractive => order.filter( c => c.reset && c.reset( ractive ) ).map( c => c.name )
+};
+
 const custom = {
 	adapt: adaptConfigurator,
+	computed: config,
 	css: cssConfigurator,
 	data: dataConfigurator,
 	template: templateConfigurator
@@ -32,12 +39,6 @@ const order = [].concat(
 	custom.template,
 	custom.css
 );
-
-const config = {
-	extend: ( Parent, proto, options, Child ) => configure( 'extend', Parent, proto, options, Child ),
-	init: ( Parent, ractive, options ) => configure( 'init', Parent, ractive, options ),
-	reset: ractive => order.filter( c => c.reset && c.reset( ractive ) ).map( c => c.name )
-};
 
 function configure ( method, Parent, target, options, Child ) {
 	deprecate( options );
--- a/src/Ractive/config/defaults.js
+++ b/src/Ractive/config/defaults.js
@@ -1,3 +1,5 @@
+import { create } from 'utils/object';
+
 export default {
 	// render placement:
 	el:                     void 0,
@@ -21,8 +23,8 @@ export default {
 	contextLines:           0,
 
 	// data & binding:
-	data:                   {},
-	computed:               {},
+	data:                   create( null ),
+	computed:               create( null ),
 	syncComputedChildren:   false,
 	resolveInstanceMembers: true,
 	warnAboutAmbiguity:     false,
--- a/src/Ractive/config/defaults.js
+++ b/src/Ractive/config/defaults.js
@@ -24,6 +24,7 @@ export default {
 
 	// data & binding:
 	data:                   create( null ),
+	helpers:                create( null ),
 	computed:               create( null ),
 	syncComputedChildren:   false,
 	resolveInstanceMembers: true,
--- a/src/Ractive/config/registries.js
+++ b/src/Ractive/config/registries.js
@@ -7,13 +7,15 @@ const registryNames = [
 	'decorators',
 	'easing',
 	'events',
+	'helpers',
 	'interpolators',
 	'partials',
 	'transitions'
 ];
 
 const registriesOnDefaults = [
-	'computed'
+	'computed',
+	'helpers'
 ];
 
 class Registry {
--- a/src/Ractive/config/runtime-parser.js
+++ b/src/Ractive/config/runtime-parser.js
@@ -1,4 +1,4 @@
-import { fromExpression, fromComputationString } from 'parse/utils/createFunction';
+import { fromExpression } from 'parse/utils/createFunction';
 import { doc } from 'config/environment';
 import { fatal } from 'utils/log';
 import { addFunctions } from 'shared/getFunction';
@@ -24,7 +24,6 @@ const TEMPLATE_INSTRUCTIONS = `Either preparse or use a ractive runtime source t
 
 const COMPUTATION_INSTRUCTIONS = `Either include a version of Ractive that can parse or convert your computation strings to functions.`;
 
-
 function throwNoParse ( method, error, instructions ) {
 	if ( !method ) {
 		fatal( `Missing Ractive.parse - cannot parse ${error}. ${instructions}` );
@@ -37,8 +36,11 @@ export function createFunction ( body, length ) {
 }
 
 export function createFunctionFromString ( str, bindTo ) {
-	throwNoParse( fromComputationString, 'compution string "${str}"', COMPUTATION_INSTRUCTIONS );
-	return fromComputationString( str, bindTo );
+	throwNoParse( parse, 'compution string "${str}"', COMPUTATION_INSTRUCTIONS );
+	const tpl = parse( str, { expression: true } );
+	return function () {
+		return tpl.e.apply( bindTo, tpl.r.map( r => bindTo.get( r ) ) );
+	};
 }
 
 const parser = {
--- a/src/Ractive/construct.js
+++ b/src/Ractive/construct.js
@@ -5,12 +5,12 @@ import { findInViewHierarchy } from 'shared/registry';
 import dataConfigurator from './config/custom/data';
 import RootModel from 'src/model/RootModel';
 import Hook from 'src/events/Hook';
-import getComputationSignature from './helpers/getComputationSignature';
 import subscribe from './helpers/subscribe';
 import Ractive from '../Ractive';
 import { ATTRIBUTE, INTERPOLATOR } from 'config/types';
-import { assign, create, hasOwn } from 'utils/object';
-import { isString } from 'utils/is';
+import { assign, create, hasOwn, keys } from 'utils/object';
+import { isString, isFunction } from 'utils/is';
+import { splitKeypath } from 'shared/keypaths';
 
 const constructHook = new Hook( 'construct' );
 
@@ -51,6 +51,8 @@ export default function construct ( ractive, options ) {
 		ractive[ name ] = assign( create( ractive.constructor[ name ] || null ), options[ name ] );
 	}
 
+	const computed = ractive.computed = assign( create( ractive.constructor.prototype.computed ), options.computed );
+
 	if ( ractive._attributePartial ) {
 		ractive.partials['extra-attributes'] = ractive._attributePartial;
 		delete ractive._attributePartial;
@@ -65,14 +67,15 @@ export default function construct ( ractive, options ) {
 
 	ractive.viewmodel = viewmodel;
 
-	// Add computed properties
-	const computed = assign( create( ractive.constructor.prototype.computed ), options.computed );
+	keys( computed ).forEach( k => {
+		if ( isString( computed[k] ) || isFunction( computed[k] ) ) computed[k] = { get: computed[k] };
 
-	for ( const key in computed ) {
-		if ( key === '__proto__' ) continue;
-		const signature = getComputationSignature( ractive, key, computed[ key ] );
-		viewmodel.compute( key, signature );
-	}
+		if ( splitKeypath( k ).length === 1 ) {
+			viewmodel.compute( k, computed[k] );
+		} else {
+			computed[k].pattern = new RegExp( '^' + k.replace( /\*\*/g, '(.+)' ).replace( /\*/g, '((?:\\.|[^\\.])+)' ) + '$' );
+		}
+	});
 }
 
 function getAdaptors ( ractive, protoAdapt, options ) {
--- a/src/Ractive/construct.js
+++ b/src/Ractive/construct.js
@@ -8,9 +8,9 @@ import Hook from 'src/events/Hook';
 import subscribe from './helpers/subscribe';
 import Ractive from '../Ractive';
 import { ATTRIBUTE, INTERPOLATOR } from 'config/types';
-import { assign, create, hasOwn, keys } from 'utils/object';
-import { isString, isFunction } from 'utils/is';
-import { splitKeypath } from 'shared/keypaths';
+import { assign, create, hasOwn } from 'utils/object';
+import { isString } from 'utils/is';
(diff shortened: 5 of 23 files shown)


ghost:
> This addresses a handful of requests involving computeds starting from at least two years back.

Make that _three years_ https://github.com/ractivejs/ractive/issues/956 😁 

I settled with computed props after that tho.

ghost:
I'm also good with a 0.10 release to ease everyone into the new changes. I particularly like the dropping of the computed shorthand in favor of expressions.

ceremcem:
@evs-chris How do we set a data in `@keypath` context with wildcard computeds? For example, I want to set `.w` by setting the `.area` property [here](https://ractive.js.org/playground/?env=docs#N4IgFiBcoE5QdgVwDbIDQgM5QNrgC74AOmkA9GQMYAm8AdAFabUCmyAlgG4x3wv5l4RALZkAAi06YAtJTAx2mMjACGlfFxZiAjHQAM+6QHd2yarID2woonwtpevduVqNnFo2wBdAL4ZskCB8RgAEAEqumgAUwAA68LF2yJAhAOQARhbUAJ6paPGJdtbIKnYpAAYF+InAwADERCowGirIIZkAHgAKMBZELM3sLJg+PlWJADwoAHzj1fi1dSxqYO0WHcOjc5Mc0ybU+GAptXRGo2ghYCzsAOZg+MfAdGDnIU3LKRPsQrYhnK2IFgAXliIBO7xUWxAITI0wmZF22wW9XYADMQnR0lsEvNEjVgNNOj0+gMNMMMek3pg1hsRmMcXjkWQ0djGfiyMs5Kz5vCZnNamRGoNWty8QzkYT1sT+oNNvTEuV8jjqKUVMc5p1hikcEjgEYUtoLkcQgAmC7pSA68Vshb6kIAZiNKQALH4kYy9SkAKxOkIANnN2vduPmnpCAHYnQAOeUhm2+YO+JU2sMATl92j0scZXiqbpxlCsNjs1HV4tSACoK3QIakyzabvwUqjEPB1OwLPAopREDAYABKOLW+YUEIACQsoWoFhCRhYIUb+BCYgA1ixso1DpcBiwAPwhYMwfi9+AhHt904hCtn3s8MAAbiR+ZtmCbIRbbY0nai-3QITXG6lGAg7BqOE5TjOc4LvwITlOoHTlCEFgwMuAGbqsVxHvuwaFvAmAWMgHjIBYNxRKkTQ3IgwgsPA+CkHkbwwJR1G0Zg-aPsO2b4LG9I+P2IA+EAA).

evs-chris:
Turns out I missed adding in the context for the setter, but that's still a bit of an open question. Would it be better to get `parent object, keypath, *keys` as args on the getter and setter or just a context object or some other combination of those things?

ceremcem:
I liked the @martypdx 's [proposal about "optional `scope` variable"](https://github.com/ractivejs/ractive/issues/1583#issuecomment-67503952). So if we use getters and setters, we could do:

```
computed:
  'boxes.*.area': {
    get: function (scope) {
       box = scope.context;
       return box.w * box.h;
    },
    set: function (value, scope) {
      ...
    }
  }
}

**Comment by ceremcem** [comment]
...but, thinking again, if these variables (`context` and `keypath`) would be passed separately; 
1. simple computeds does not need extra lines
2. current getters and setters will not be broken

```
computed:
  'boxes.*.area': {
    get: function (box, ctx) {
      this.get('myParentData');
      return box.width * box.height;
    },
    set: function (value, box, ctx) {
      ctx.set('.width', value / box.height);  // set box width by area property
      box.width = value / box.height; // I'm not sure about this. Is it possible?
      ctx.set('~/changes.' + ctx.getKeypath(), value); // store the changes in a separate branch
      ...
    }
  }
}
```

paulie4:
I found an issue. I'm not sure if I should create a new GitHub issue, since this is using this pull request version of Ractive, but the problem also happens when using the undocumented `ractive.viewmodel.compute()` method.

This is the [wild-compute version](https://ractive.js.org/playground/#N4IgFiBcoE5QdgVwDbIDQgM5QNrgC74AOmkA9GQMYAm8AdAFabUCmyAlgG4x3wv5l4RALZkAAi06YAtJTAx2mMjACGlfFxZiAjHQAM+6QHd2yarID2woonwtpevduVqNnFo2wBdAL4ZskCCUFvCY+AAEAMr4FjAs4QC84XxG4QBKrpoAFMAAOvC5+NQq+CqQ4XkF+IWYViwAairI5QDknE0t+YU++T4AlADc+dGxHsHWtixZLeM2dtQA1kQtaBVd+ADm-OUAZojw6uwhWX0V4XH4iDDw4S0W+GAsMI3ILQPhPfD9QwUHIWHhTjsFhGbSJZIg9KZdw5dZscrUCyURDCFjwfB0Lb4ACiyBYqPRACEAJ4ASWo020LT6aHWIWC8B27A2u32h2Op0qhUKD0UdA48AW0xWtxGcRFwFUh3c5TFLG+6z86zs1mQJRY5QABgAeahccKUNWYGS1VGyI2YBK5EBynAtU0NDpea0APm1ZD1nBdmt6gy6f1CESBIIATOCUlDpVMuXZmuFEciCRisbj8Wj8CTydMQ9TaVV6SEmSzwnsDhoOWsqjywHyBUKWiKWnKJVK3BqojE4gqqkqqrNJtRyjHCsJiQBhKxzFiDkts8vwE6V7nVfAXK43OWY-jTfvzJbUn7Lz7dPM8-FENV2VrAYCjicTeY+HydL6DEAYfBQEC6-XsahWkBtFdd1PRdfIf04cI-wAkNgI9LgXRAHwgA).
This is the [undocumented viewmodel.compute() version](https://ractive.js.org/playground/#N4IgFiBcoE5SBTAJgcwSANCAzlA2gLoC+WukIAxgPYB22ALgAQDK9VMCjAvIzQgO6MASgEMK9AJYA3BAApgAHRoL6SEfRGRGi5fRXYqAWwQA1EQBstAcikWrSlUSVEAlAG4lrdggB0UiQKGVEgI5j7UhgAOAK70clYRMXFIANaRVhjaDvRo9HEwWgBm0TTiErSyLtqMHPTRMDSMVlT0YAgwZuZWboxONK4eyqW0DIz+AgCM3LwCwmKSMvLZoVpIVBTRxjT0PrkAouYIW-QAQgCeAJJIslYTVi4Y2bTUNIUSKEUlZRVVOioqrQk2B85gkNBSNwyTS8HChwBg82kCC0MIQA2yJGycSi5nUyMYAAMADxIaSMCi47DYAC0BmM1IpIipXAUIFReCsdNMdgIrIAfESAPSkqR8gnOdwOYZ0JjjfgAJmmfEEojKiz+cUsjDWG2OuwQ9AORwQ23OVxu8vuj10z1obw+jGKpUkPyyugBYCBILBEKsUKsqLhCLV+NR6N0mN0iViyC0GpUhjOAGEjElY46vi6aJU3f89PRavVGqj9fQbtHkml7oM831HNaAUdIri4tZgMBEymojGkEQiPZ+u5MCB6PASWSJEgWSAJvyhSK+Upx1JGJPp-K58LpHyQEQgA).

I think the problem is that when `view1` is rendered, it links its `ReferenceExpressionProxy` as a dependency on `Store.viewmodel.links[0]`, and then, when `view2`'s computed runs and calls the dynamic computed of `Store`, Ractive is incorrectly trying to use the `runloop`'s `batch` before `runloop.start()` is called. You can put a conditional breakpoint on line 33 of runloop.js (or line 5236 of ractive.js) with a condition of "`!batch`" to break at where the exception is thrown.

evs-chris:
@paulie4 it sounds like maybe the init cycle needs to take place in a run loop in case there are updates triggered by settling computeds. I'll poke at it a bit.

paulie4:
@evs-chris it looks like the problem was exacerbated by the addition of `this.base.register( intermediary )` to the `ReferenceExpressionProxy` constructor in [this commit](https://github.com/ractivejs/ractive/commit/4266a32a69f78e61407c3db5d8ea910893007c25#diff-eb2b86b25094f19146a6286403895bf6R126). Before, a dynamic computed call from an indirectly linked `Store` would not crash, but after that change, `ractive.get()`ing the tangentially related computed value does cause a crash. I created an example where the view is linked to `Store2`, which in turn is linked to `Store1`, and calling `Store1.get('computedkp')` only causes a crash after the change was made.

This is before the change and does not crash: [using v0.9.3](https://ractive.js.org/playground/#N4IgFiBcoE5QdgVwDbIDQgM5QNrgC74AOmkA9GQMYAm8AVpgHSXID2i1AZsgIYwCmzVgFsyPOjwAeZZAEsARpjIwelfLIBu-MgAZGATkYBmZavVbGDEAF0AvhmyQQlVvEz4ABAGV8rAQEYPAF4PeH4Adw8AJTNNfgAKAEoAbgAdeB8-fn9GDVkI4VZqfmQhYSJEfASAchdyyv5qAGsiarQPYHTU-ABzfkJ+GEgPTkR4NVlXJI6PAXxEGHgPatZ8MEGANR5kauSPW3TbFK7x13dvXwEAJmDQiOjYrXjO+G7qHnweYZfu7swRfhbZDDaoabbVLr4A7wI5pDKXfhXRhyeBNeLVNrLTIBTHAFQTLTDbHZWEnFxuTx5CKBEJhSIxAkJH5VYEeaisSiIYT8eD4Rh9fAAUWQ-G5vIAQgBPACS1HR-mqiTQkNc5M4sh6w1G43UU0SHUh3TWsiYKLRGPa1WJV1x+PM-CJCKupNeUOVrqq5V4VWGAANgMBrThqv9uUDqnZbL7DsdXvCsjkBei6hUqs1WikQBh8FAQAAeaiaDyyahBVIgfzlgB8ebIhY0VZAtiAA)
This is after the change and does crash: [using v0.9.4](https://ractive.js.org/playground/#N4IgFiBcoE5QdgVwDbIDQgM5QNrgC74AOmkA9GQMYAm8AVpgHSXID2i1AZsgIYwCmzVgFsyPOjwAeZZAEsARpjIwelfLIBu-MgAZGATkYAWZavVbGDEAF0AvhmyQQlVvEz4ABAGV8rAQEYPAF4PeH4Adw8AJTNNfgAKAEoAbgAdeB8-fn9GDVkI4VZqfmQhYSJEfASAchdyyv5qAGsiarQPYHTU-ABzfkJ+GEgPTkR4NVlXJI6PAXxEGHgPatZ8MEGANR5kauSPW3TbFK7x13dvXwEAJmDQiOjYrXjO+G7qHnweYZfu7swRfhbZDDaoabbVLr4A7wI5pDKXfhXRhyeBNeLVNrLTIBTHAFQTLTDbHZWEnFxuTx5CKBEJhSIxAkJH5VYEeaisSiIYT8eD4Rh9fAAUWQ-G5vIAQgBPACS1HR-mqiTQkNc5M4sh6w1G43UU0SHUh3TWsiYKLRGPa1WJV1x+PM-CJCKupNeUOVrqq5V4VWGAANgMBrThqv9uUDqnZbL7DsdXvCsjkBei6hUqs1WikQBh8FAQAAeaiaDyyahBVIgfzlgB8ebIhY0VZAtiAA)

paulie4:
@evs-chris, thanks for fixing #3169. Does that mean this can get merged back in the mainline branch soon, hopefully before 1.0 is released?

evs-chris:
@paulie4 sorry, been slammed lately, but yes, the plan is to pull this into 0.10 to make sure nothing needs to break for 1.0. It's toward the top of my list as soon as the old day job lets up a bit.

evs-chris:
This is now cherry-picked onto edge.

Issue #3148, opened 2017-11-23
fire method does not work in onrender


ceremcem:
## Description:

(included in the reproduction)

## Versions affected:

above 0.7

## Reproduction:

[Playground](https://ractive.js.org/playground/?env=docs#N4IgFiBcoE5SAbAhgFwKYGcUgL4BoQN4AlJAYxQEsA3NAOjIHsBbAB0YDs0OUM6AzRowAEAXmGkKNemgAe6DgBMAFMAA6HNSnRtk6SMIAGKMJQzCzwwY0N4NWzjG6K0MA-wCuHKZ2UBKdU0ULRMzAUonZQByACMkGCi-exQcDRwkji4AdwlyKlpVZLQEA1jGRQBPKLsgnVY9NANDZK0AHjAAJgA+AFFZVjQKA1aAek6ulpRW1i6AdTRheIW5AakOAHNFjkWEVxRhGMZZYSzKE2E1EH4ItEVhMGKERjoXy+FmTAwkdYWPdm2YF5RjNJtMugAVUzmLKcKL7GEwADWFn4whMCyYbE43H2HgwmAuIEczlcbyQShOjCRGEmWkoqLOwjxBMunEx9TQ6EuwImQTaMQ82k4wk4AFoyAhKGREaJLg8EE9Ll1rk57o9GKMBUKOLytG1rCKOKK4jBZeB1W8Rl0LkFbMliUpXO4vD4OP5AnrtFC6L4PcF-fKns7vFRfAFaf6kLsYChoirbmqFc8XokANwR1J8lJ+dNBTPpXBAA)

ghost:
I think I've narrowed down the issue. The handlers declared on a component (in this case, `on-bar`) is registered _after_ `onrender` is emitted. By `oncomplete`, they are all present. I still think it's a bug tho, given that `onrender` is described to be "after render, before transition complete".

fskreuz:
It appears to be caused by https://github.com/ractivejs/ractive/blob/e1d4a0107cc758f4ff92f682c08a2018db7f0a45/src/view/items/shared/EventDirective.js#L162

PR #3149, opened 2017-11-23
Child events on render and complete should trigger proxy events.

ghost:
## Description:

Moves render after attribute and event handler registrations so that they're registered before the render hook is fired.

#3148 

## Is breaking:

Whatever the test says. 😁


evs-chris:
Travis is cool with it, and it looks legit to me 😄

PR #3154, opened 2017-12-01
Fix isObjectType( null )

MartinKolarik:
Fixes a bug introduced in e1d4a0107cc758f4ff92f682c08a2018db7f0a45

### DIFF
--- a/src/utils/is.js
+++ b/src/utils/is.js
@@ -33,7 +33,7 @@ export function isObjectLike ( thing ) {
 }
 
 export function isObjectType ( thing ) {
-	return typeof thing === 'object';
+	return thing && typeof thing === 'object';
 }
 
 export function isFunction ( thing ) {


evs-chris:
`isObjectType` _should_ be new as of e1d4a01 and only replace `typeof x === 'object'` in the code, so it shouldn't affect either `isObject` or `isObjectLike`. Did I miss a case checking for breakage?

MartinKolarik:
I updated from 0.9.3 to 0.9.10 and I'm getting errors [here](https://github.com/ractivejs/ractive/blob/b6af965d27f636202783f0d729925f7e7d837769/src/view/items/Partial.js#L253) because `tpl` is `null`. I checked that adding a `null` check there fixes it and assumed it was accidentally removed. I reviewed e1d4a0107cc758f4ff92f682c08a2018db7f0a45 again now and it indeed didn't remove any `null` checks.

That said, `tpl` [can be null](https://github.com/ractivejs/ractive/blob/b6af965d27f636202783f0d729925f7e7d837769/src/view/items/Partial.js#L258) so that previous check is definitely wrong, even if it somehow worked before.

I thought checking for `null` in `isObjectType()` would be better as I can't think of any situation where you would want to treat it as an object, but it seems to break some test so not sure now.

MartinKolarik:
Ah, `partialFromValue()` was added just before e1d4a0107cc758f4ff92f682c08a2018db7f0a45. In that case #3155 is probably the right solution.

PR #3170, opened 2018-01-03
RFC: Monorepo take 2

ghost:
# For Review. Do not merge.

Continuation of the work done in https://github.com/ractivejs/ractive/pull/3052

### Package structure

Each package will have a minimum an `src` and `tests` directory, with an `index.js` and a `rollup.config.js`, a `README.md` and a `package.json`. Each package will have a `build` and `test` script for bulk build and test. Build and test scripts may be specific to the package since each may be build differently, (i.e. browser plugins vs cli tools). All packages shall emit a `dist` directory containing the bundled package. UMD and ES formats for browser packages, and CJS for node packages. Bundles are pre-transpiled with Buble and bundled with Rollup but not minified - leaving that part to the consumer.

Packages are namespaced with `@ractivejs` in preparation for release on the `@ractivejs` npm org. This will mean it will not be published on any of the existing releases.

### Component file changes

Component file gains a top-level `<template>`  for aesthetic purposes as well as limiting the possible top-level elements to just 4 (link, template, script, style). This also allows future replacement of `Ractive.parse()`, which I find overkill for a pre-parser, with a simpler parser.

While that changed, the component file syntax remains the same. I believe writing in ES brings little benefit but a whole lot of drawbacks. For one, you need a build setup to transpile it and a parser to properly parse it - both of which require a build step which is not feasible on all setups and not feasible to run on runtime. On the other hand, the existing format (CJS, the `components.export`) runs readily on the browser. The only drawback I see is the `require()` scanning. See `utils-component` test specs and `toParts` implementation.

And oh, `.ractive.html` is now official, along with the old `.html`. APIs in `utils-component` use this extension to filter Ractive component paths in dependencies.

### Parser and builder APIs

`rcu` and `rcu-builders` are now merged under `utils-component`. This package will be the one-stop shop for your parsing and construction needs. Tools like loaders and cli will simply consume these APIs or be wrappers for the APIs in this package. This allows one to update parsing and construction logic without mucking tool-specific things. See `loader-rollup` for an example.

Construction is also separated into parser and builder, allowing pre-processing tools to manipulate the parser results before handing it off to the builder APIs. See `*.parts.json` in `utils-component` for the parser outputs and `loader-rollup` implementation for usage example.

Using a monorepo also gets rid of template version checks. Each package will always be compatible with the accompanying Ractive.

## Questions

- Does it make sense to publish all this as a separate repo or just work on top of the existing repo?
- Am I missing a use case/workflow/pattern?

---

Would like some feedback regarding setup, development and testing, `utils-component` in particular. To get started, simply checkout this branch and do:

```sh
npm run bootstrap # installs and links dependencies
npm run build     # bundles all packages
npm run test      # runs all tests (one test in utils-component will blow up atm)
```


evs-chris:
I like where this is heading, especially all of the tests being able to run at once.

For components, I don't know that swapping out the parser with something simpler would work, as you'd still have to parse it again in the browser and take a significant perf hit. If you're not pre-compiling it wouldn't matter, but I think that's a pretty common optimization. Aside from that, I'm definitely fine with including the rcu stuff in the npm package as long as it doesn't bloat core. I also think that having both bin and rcu available for components is not awful as they each fit very disparate workflows.

I _would_ like to see the plugins (event available as es modules, and the same goes for components if they have any dependencies on other modules. Build systems that dedupe rather than requiring global config feel much better to me. For general consumption, I agree that a build system should not be necessary, but for stuff in the main repo, the build system is already there so may as well take advantage.

1. I think this is fine moving stuff around in dev and then pulling in the modules when it's ready.
2. ???
3. Profit

ghost:
> For components, I don't know that swapping out the parser with something simpler would work, as you'd still have to parse it again in the browser and take a significant perf hit.

I could add back an `rcu.make()`-like function for in-browser loaders (i.e. ractive-load) that would skip the intermediate representation that `toParts` and builders do, and create the constructor immediately. This way, the parser is only ever run once. This would mean using it would mean no pre-processing at runtime (sorry SystemJS users 😛 ).

But then again, the APIs are available. One can make a loader that does pre-processing at runtime if they really want to.

> Aside from that, I'm definitely fine with including the rcu stuff in the npm package as long as it doesn't bloat core.

Each package in `packages` will be published as a separate package in npm under the ractivejs org. Ractive will be `@ractivejs/core` but aside from that, nothing will change. Ractive will still be the Ractive we all know and love.

utils-component is also meant to be embedded and not consumed as a standalone lib. With the right build tool (\*cough\* Rollup \*cough\*), a loader can just import the APIs it needs and tree-shake the rest. For instance:

- A Rollup loader like rollup-plugin-ractive will only need `toParts` and `toES`. When built, it will bundle `toParts` and `toES`, and shakes off the rest.
- A Browserify loader like ractiveify will only need `toParts` and `toCJS`. When built, it will bundle `toParts` and `toCJS`, and shakes off the rest.
- An AMD loader like rvc will only need `toParts` and `toAMD`. When built, it will bundle `toParts` and `toAMD`, and shakes off the rest.
- An in-browser loader like ractive-load will only need the `make` function (mentioned above). When built, it will only bundle `make`, and shakes off the rest.

Still working on `loader-rollup` and `loader-browser` for a sample implementation of how build-time and run-time loaders should be built and how it utilizes utils-component APIs. Testing these things and coming up with test cases is really tricky.

> I also think that having both bin and rcu available for components is not awful as they each fit very disparate workflows.

Yep, that's the plan. rcu, rcu-builders and the ractive lib that powers the bin is merged into utils-component. It's rewritten into composable APIs that address specific phases of building components. This way, authors of tools can just import what they need, insert their secret sauce, and build away their tool.

> I would like to see the plugins (event available as es modules, and the same goes for components if they have any dependencies on other modules. Build systems that dedupe rather than requiring global config feel much better to me.

Everything will be pre-bundled, pre-transpiled (and in the case of components, pre-converted to UMD and ES). This way, the consumer doesn't have to worry about build config and setup. But there are cases where plugins require a library API (like in adaptor-backbone). In that case, that dependency will be defined as a peer dependency, dependent on what the consumer has.

Issue #3178, opened 2018-01-18
RFC: Plugins


PaulMaly:
Hi guys!

I want to suggest an additional method for the Ractive's core which is very useful in the modular approach. I won't hide it's just a tracing-paper of Vue's plugins you can find [here](https://vuejs.org/v2/guide/plugins.html).

## Use

In Vue this method is basically needed to extend Vue.prorotype conveniently and transparently with new properties and methods from 3rd party modules (called plugins). And also to control that the plugin will be applied only once. But I suggest making this method available also to instances of Ractive.

In my solution, plugin is just a function which called in context of caller with additional params:

```javascript
/**
 * This callback fired when all waitings resolved.
 * @callback pluginCallback
 * @param {...(*)} args - Bunch of arguments
 * @this (Ractive|instance)
 */

/**
 * Apply external function with args to Ractive's constructor or Ractive's instance
 *
 * @param {pluginCallback} plugin - Plugin function which will be exetuted in context of coller
 * @param {...(*)} args - Bunch of arguments
 */
function use(plugin, ...args) {};

Ractive.prototype.use = use;
Ractive.use = use;
```

```javascript
// ./plugins/myplugin.js

// plugins files itself doesn't need to import Ractive or something. it's just a function.
export default function (options) {
       this.newMethod = function() {
              console.log(options.some);
       };
}
```
```javascript
import Ractive from 'ractive';
import MyPlugin from './plugins/myplugin';

// assign new method(s) or prop(s) to Ractive constructor
Ractive.use(MyPlugin, {some: 'option 1'});

// next call of the same plugin will be ignored (do nothing)
Ractive.use(MyPlugin, {some: 'option 1'});

Ractive.newMethod(); // => 'option 1'

// the same behaviour is for instances
const app = new Ractive({});

app.use(MyPlugin, {some: 'option 2'});

app.newMethod(); // => 'option 2'
```

If this suggestion will be accepted, I can provide the pull request for these features. Actually, I already have that code as a separate module.

Expect your comments, thanks!

ghost:
I'm not exactly sure what the goal of this request is. How different is it from Ractive's registry approach? The way I see it, the proposal is just extra indirection with setters and wrapper objects (which equates to more tutorial explaining and more things to document down the road). Ractive's approach is much simpler - all plugins are assignments to their respective registries. No setters, no wrapper objects, no inconsistent syntax. Just pop it in there, done.

[...]

PR #3182, opened 2018-01-21
Proposal for plugin extension point - RFC - #3178

evs-chris:
## Description:
This is my take on the `use` API from #3178. It's super simple and has the following characteristics:

1. The plugin author and user are responsible for config, similar to express middleware e.g. `use(myPlugin(config))` rather than Ractive trying to be the middleman for config.
2. Ractive/Components and instances have their own separate handling so that you can target either with the same plugin without any special casing. The object passed to the plugin currently contains a `target` property, which is the place where any extra methods or properties should go (the prototype or instance), and a `registry` property, which is where any decorators, components, etc should be registered (the constructor or the instance). So you can `target.newMethod = function() {}` and `register.decorators.foo = function(node) {...}`.
3. Multiple plugins can be registered in one go, because in my experience, `Component.use(fade, aceEditor, tabs, table)` is a bit more ergonomic than individual `use`s, though there are some benefits to having explicitly listed individual components.
4. At some point in the future, the object passed in to the plugin function could supply additional info, like whether or not this is server environment. This would also give us a good place to move some of the core APIs, like the transition and adaptor helpers, into core modules that are optional. The only things that use those are plugins, so that may even make a source module build tree-shake-able in the right build env when ractive is part of the bundle.
5. The target of `use` is returned from the function, cause I like the ability to chain methods.

### TODO

* [ ] Tests. I've played with it a bit, and everything looks good. If this looks good, I'll fill out the test suite.

#3178

## Is breaking:
No.

### DIFF
--- a/src/Ractive.js
+++ b/src/Ractive.js
@@ -15,6 +15,7 @@ import macro from './extend/_macro';
 import parse from './parse/_parse';
 import getContext, { getNodeInfo } from './Ractive/static/getContext';
 import isInstance from './Ractive/static/isInstance';
+import use from './Ractive/static/use';
 import construct from './Ractive/construct';
 import initialise from './Ractive/initialise';
 import { getCSS } from './global/css';
@@ -84,6 +85,7 @@ defineProperties( Ractive, {
 	splitKeypath:     { value: splitKeypath },
 	// sharedSet and styleSet are in _extend because circular refs
 	unescapeKey:      { value: unescapeKey },
+	use:              { value: use },
 
 	// support
 	enhance:          { writable: true, value: false },
--- a/src/Ractive/prototype.js
+++ b/src/Ractive/prototype.js
@@ -43,6 +43,7 @@ import unrender from './prototype/unrender';
 import unshift from './prototype/unshift';
 import update from './prototype/update';
 import updateModel from './prototype/updateModel';
+import use from './prototype/use';
 
 import { defineProperty } from 'utils/object';
 
@@ -94,7 +95,8 @@ const proto = {
 	unrender,
 	unshift,
 	update,
-	updateModel
+	updateModel,
+	use
 };
 
 defineProperty( proto, 'target', {
--- a/src/Ractive/prototype/use.js
+++ b/src/Ractive/prototype/use.js
@@ -1,0 +1,10 @@
+export default function use ( ...plugins ) {
+	plugins.forEach( p => {
+		p({
+			proto: this,
+			Ractive: this.constructor.Ractive,
+			instance: this
+		});
+	});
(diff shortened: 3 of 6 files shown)


PaulMaly:
@evs-chris Very nice, but I think you missed one very important thing - we need to prevent multiple applying of the same plugin to the same ``register``. It's not so hard but will allow us to use of plugins by safer, especially in a case with an instance:

```javascript

export default function use ( ...plugins ) {
    const installedPlugins = (this._plugins || (this._plugins = []));
    plugins.forEach( p => {
        if (typeof p === 'function' && installedPlugins.indexOf(p) < 0) {
            p({
                target: this.constructor.prototype,
                register: this
            });
            installedPlugins.push(p);
        }
    });
    return this;
}
```
Also, I don't remember any other Ractive's api where we use arguments in this way. The ``target`` reminds me HTML target option and ``register`` is a brand-new thing. Maybe we could make it more consistent?

PaulMaly:
@evs-chris I suppose, the good solution of the last remark will be:

```javascript

module.exports = function(Ractive, target) {

    // Ractive constructor to assign static stuff
    Ractive.transitions.fade = function() {};
    
    // Ractive prototype to assign class stuff
    Ractive.prototype.newMethod = function() {};

    // And context object (or target/caller) to assing specific stuff 
    target.foo = false;

    // Also I able to check if context an instance or not
    if (Ractive.isInstance(target)) {
        // specific behavior 
    }
    // An other way to check
    if (Ractive === target) {
        // specific behavior      
    } 
};
```
I think I want to have with guarantee ``Ractive`` constructor (like in Vue plugins), and also I want to have ``target`` (or caller), who call ``use()`` method (replacement of a context, seems nobody loves it).

It looks like much cleaner when using arguments inside plugin. It also much more consistent with other Ractive's plugins, like decorators/adaptors/etc.

evs-chris:
For the args, I would specifically like to avoid positional args because changing them in the future is problematic. As it stands, it's no big deal to throw in a `Ractive` prop on the `opts` arg. You can also do `function({ Ractive, target }) {}` just about everywhere now, so it's less painful than it used to be.

For deduping, out of curiosity, what do you need that for?

The problem with deduping is that it's nigh impossible to do if the config returns a fresh function for each call e.g. `r.use(fade({ duration: 1000, name: 'slowFade' }))`. You'd have the same issue if you allowed ractive to manage the opts, because there are legitimate cases where you'd want the same plugin with different names. One solution to that could be to allow the plugin function to return an object with an id string and use the id for deduping.

PaulMaly:
@evs-chris how do you think when this PR will be available in dev?

PR #3189, opened 2018-02-03
First pass at bringing ractive typings up to date - RFC

evs-chris:
## Description:
This updates the typings defs for typescript to (_mostly_?) match the current API. The only notable thing that's definitely missing is the various `on${whatever}` init options, which I should probably remedy, but it feels like those could _probably_ go away.

I've only been typescripting for a couple of days, so there may be a better generic way to handle some of this. Notably, I feel like there's probably a way to resolve some of weirdness around Static vs Ractive with generics, but it's super weird resolving where and what to do with `T` vs `typeof T` in my head right now.

This also adds a `typings` entry in the package manifest so that typescript will pick up the distributed d.ts file.

Way out of date typings.

## Is breaking:
No, unless you like the out-of-date typings for some reason.

### DIFF
--- a/manifests/package.json
+++ b/manifests/package.json
@@ -7,6 +7,7 @@
   "main": "ractive.js",
   "jsnext:main": "ractive.mjs",
   "module": "ractive.mjs",
+  "typings": "typings/ractive.d.ts",
   "bin": {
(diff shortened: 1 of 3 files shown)


ghost:
lgtm 👍

PR #3191, opened 2018-02-04
RFC: Monorepo experiment, take 3

ghost:
# PR FOR REVIEW/COMMENTS ONLY. NOT FOR MERGE.

## Description:

Take 3 of that same monorepo PR as before. Just for review, things still need finalizing.

<sub>Github does not show diffs of things inside directories named `node_modules` regardless if they're from npm or if you created them yourself. This is best viewed [if you view the branch directly.](https://github.com/fskreuz/ractive/tree/monorepo-experiment)</sub>

### Drops most of the existing tooling

What went away:

- Gobble, gone. Bash scripts and Rollup FTW!
- Sandbox, gone. We have playground which is more accessible than sandbox.
- Code coverage, gone. Not sure how it makes sense on a monorepo.
- Qunit, sort of gone. The files in the repo now replaced by karma-qunit and qunit cli.
- Bower, sort of gone. You can still use CDNs which some Bower workflows support.
- Linting... not yet implemented. Probably last.

For a monorepo, Lerna is NOT used. It uses the [Alle package structure](https://gist.github.com/nolanlawson/457cdb309c9ec5b39f0d420266a9faa4). I found out that I only needed Lerna mostly for linking sibling packages. The rest were mostly bash scripts. This setup takes advantage of node's resolution algorithm to do module discovery for sibling and dev dependencies.

### Unified control, per-package scripting

To allow flexibility on builds, the top-level expect only 3 scripts from each package: `make`, `test` and `publish` (not yet implemented). This allows a consistent interface into each package, while allowing each package to customize their build process. For the most part, most packages use Rollup. However, browser plugins/tools use karma-qunit for testing, while node tests use QUnit cli. A notable oddity is adaptor-ampersand whose build includes building Ampersand with Browserify because Ampersand is written and published purely in CJS.

### Zero deps!

Speaking of zero deps, most if not all packages have zero dependencies. The only dependencies they have are peer dependencies (i.e. depend on user-installed core, or because the package is a plugin of another lib). Runtime dependencies are _rolled into_ the package on build if applicable. Dev dependencies makes no sense since, if you're developing an official package, you should be in the monorepo anyways (dev deps defined on the top level package.json).

### CLI

Comes with a cli version of `Ractive.parse()`. The command previously known as `component` is split into `parts` and `module` + `sourcemap`, allowing you to pre-process the parsed output before handing it to `module` or `sourcemap`. Now also comes with scaffolding tools to emit skeleton forms of Ractive plugins.

```
 Usage: ractive [options] [command]


  Options:

    -h, --help  output usage information


  Commands:

    parse [options]      Takes templates and emits its JSON/JS AST form.
    parts                Takes component files and emits its JSON parts form.
    module               Takes the output of the parts command and emits a module.
    sourcemap [options]  Takes the output of the parts command and emits a v3 source map.
    scaffold <type>      Emit a scaffold of given type. Valid values are: constructor, component, adaptor, decorator, easing, interpolator, event, transition.`
```

One difference from the previous CLI is the sole use of stdin and stdout. This keeps the commands simple, consistent, and allows it to inter-operate with already-available unix commands on scripts.

### utils-component

utils-component is the go-to package for all component file construction needs. It's a combination of rcu, rcu-builders and the cli. This way, everything related to building off from component files reside inside one package (parsing, construction, source mangling, path utilities, etc.) If you look at cli and the loaders in the PR, they're all just thin wrappers to utils-component APIs.

The builder APIs also split parsing and building. `toParts` takes in SFC source and emits JSON representing that component. The consumer can manipulate this JSON however they want (i.e. pre-process styles, transpile script, compile template, change import extensions, etc.) before passing it to either `toConstructor ` or `toModule` for building.

### Official loaders

Officially-maintained loaders! There's currently 5 of them.

- loader-browser - Will replace ractive-load on the browser.
- loader-node - Will replace ractive-load on node
- loader-system - SystemJS SFC transform.
- loader-rollup - Rollup SFC transform.
- loader-webpack - Webpack SFC transform.

ractive-load was split into loader-browser and loader-node so that the browser won't have to bring in node-specific logic (and vice versa) which keeps both light.

SystemJS was selected for a more robust runtime loader (most Angular demos in Plunkr use SystemJS). Rollup was selected primarily for good tree-shaking (and they now have splitting/lazy-loading, yes?). Webpack was included for hipsters :trollface: . Dropped all else from consideration since one or more of the above can already do it (i.e. If you use Gulp/Grunt/Gobble, you potentially just use Rollup/Webpack loaders). Still thinking of including Browserify since...

### CJS FTW!

Yes, still CJS. That means you'll still get `component.exports` and `require`. After pondering about it, AMD is a fading authoring and target format due to the rise of ESM and ubiquity of CJS. And ESM still requires transpilation/bundling to work on the browser. You would not want to transpile in the browser, nor force the user to have a compile step.

Only CJS fits the description of a format that 1) the browser can immediately run 2), node can definitely run and 3) does not require a lot of source mangling (at best, utils-component APIs only wrap, prepend or append to SFC code, not mangle).

<sub>For Rollup, one caveat is that you'll need to use `rollup-plugin-commonjs` since the builder APIs emit a CJS module. Wrapping the SFC in ESM, then having Rollup unpack it again results in code that's not very pretty.</sub>

### Breaking changes to the SFC syntax

One breaking addition is the use of `<template>` to surround an SFC's template. This is to cleanly separate it from the other top-level tags. Works well with IDE code folding, folding away the entire template block. Note that this top-level `<template>` is NOT the `<template>` HTML tag. It's just a wrapper, and will not actually appear in the component's template.

In addition, there's a limit of only one `<template>`, `<style>` and `<script>` in an SFC because I'm not really sure why you'd need more than one of each. It's also better to have a very strict standard then relax it later, than start relaxed and then take into account all unanticipated usages.

### Official plugins

Official Ractive plugins are now in this repo, rewritten to ES and now come with a battery of unit tests (except for transitions, can't find a way to properly test them). This means if they break due to changes on Ractive, we'll know (and vice versa). Also, their unit tests will bolster Ractive's unit tests with more real-world scenarios. 

<sub>Just yesterday, I found out BaconJS updated to 2.0 last December with a small breaking change. If it weren't for the monorepo, I wouldn't have noticed that at all and the BaconJS adaptor would have been outdated since then.</sub>

### Still not complete

The core package (the actual Ractive lib) is not yet updated (it's still early 0.9) because it requires some significant restructuring (module paths, test setup) to get it to work with this setup. This I'll do last.

The CI and auto-publish piece is still not in place. I'm wary of publishing to NPM, especially when they have that no-delete-after-a-day rule. I wish they had some sort of "temporary package" feature, where you can publish and delete and the consumer has to acknowledge that they are using a potentially disposable package.

The PR is also still open to more packages. I still have one more package to add to this PR that's still in the works.

- https://github.com/ractivejs/ractive/issues/2271
- Plugin and tooling maintenance

## Is breaking:

Yes.

- Most of the plugins don't expose the same export.
- The builder APIs have either been renamed, restructured or dropped.
- Loaders have been rewritten.
- Component file syntax has changed.
- The cli commands have changed.
- Bower is gone. This will publish mostly to NPM and made available by CDNs.
- Probably be on a separate repo, to preserve things pointing to Github tags directly.
- Releases will now be in "waves" and version bumps need to be scheduled/planned out carefully.


.editorconfig
```diff
@@ -1,15 +1,13 @@
 root = true
 
 [*]
+indent_style = space
```
evs-chris, line 4: 😁

packages/node_modules/@ractivejs/cli/src/index.js
```diff
@@ -0,0 +1,86 @@
+/* eslint-env node */
+/* eslint-disable no-console */
+import serialize from 'serialize-javascript'
+import program from 'commander'
```
evs-chris, line 4: Does this mean that the cli gets a dependency on commander? It seems to be missing from the package?

packages/node_modules/@ractivejs/cli/src/index.js
```diff
@@ -0,0 +1,86 @@
...
+    readFromStream(process.stdin).then(component => {
+      return writeToStream(process.stdout, serialize(toParts(component)))
+    }).catch(err => {
+      console.error(err.message)
+      process.exit(1)
+    })
+  })
+
+program
+  .command('module')
```
evs-chris, line 47: A new mode that I'm finding quite useful with typescript is to take the template, partials, and css and emit all of those that are available as `export const css = $CSS; export const template = $TEMPLATE; export const partials = $PARTIALS;`, though the partials part is a bit dubious, since they're already in the template at that point. It might make sense to have a command that does that too, so that you can have the script portion be separate in whatever compile-to-js you like as long as you support esm somewhere in your build.

packages/node_modules/@ractivejs/cli/src/scaffolds.js
```diff
@@ -0,0 +1,103 @@
+export const component = `
+<link rel="ractive" href="">
+
+<template>
```
evs-chris, line 4: The current CLI uses `<template id="somePartial">...</template>` for partials added to the template. Is that still the case here?

.editorconfig
ghost: I use standard as linter, one less config to worry about. 😁

packages/node_modules/@ractivejs/cli/src/index.js
ghost: The cli is built just like the other packages. It builds to `bin/ractive` and bundles commander, which conveniently has no dependencies other than native modules. So once published, technically, commander is _in_ the cli. 😁

.editorconfig
evs-chris: But, mah semi-colons! 😄 (I actually like semis)

evs-chris:
Initial thoughts and questions:

1. The sandbox and being able to qunit in-local-browser are both quite handy for debugging. As long as there's a way to build core and the test script to a dir, it shouldn't be a big deal to `http-server` it or something. It should definitely be built in though. I suppose maybe there's a way to tell puppeteer to use a real chrome instance?

2. Publishing is going to happen lock-step all at once? That seems like it wouldn't be awful for core, but if publishing a plugin requires a new core version, that sounds a bit unpleasant.

3. I like the alle layout much better than the requires-additional-tooling method. I'm not even mad about having to load diffs individually, cause it makes focusing on each change easier, especially with big changes.

4. I think the cli and loaders should _have_ an esm mode, even if it's just the abbreviated auto export template/css version. If you already have to have a build process, it's nice to be able to take advantage of it. Having the non-quoted JSON-ish format on templates also saves quite a bit of space and is the only way to support csp mode.

5. A reason to allow multiple scripts/styles in SFC is if you want to have some styles that use vars, which requires a script tag, and some plain-old-css in a style tag.

6. Coverage on core is super useful to me, because it lets me know if I've done something particularly boneheaded with tests in a pretty easy-to-spot way. "Coverage dropped 6.2%"... hmmm... lemme look at those tests again. Doesn't seem like it would be too hard to have the core test script handle coverage too?

7. Definitely go ahead and throw automatic formatting in there too, as the only reason I've held back is that it will make cherry-picking across version branches painful. This already stops cherry-picking from edge to any current major-minors, so no reason left not to. I like prettier better than standard, because again, mah semis :grin:.

8. ???

9. Profit!

10. Probably more, but it's dinner time!

packages/node_modules/@ractivejs/cli/src/index.js
ghost: Hmm... currently, everything is very "SFC-in, CJS-out". Currently it's like:

```js
const module = toModule(transpilePartsWithYourFunction(toParts(componentInJadeScssEsOrTs)))
const constructor = toConstructor(transpilePartsWithYourFunction(toParts(componentInJadeScssEsOrTs)))
```

or in cli

```sh
# component.ractive.html written in Jade, Scss, Ts or whatever
cat component.ractive.html | ractive parse | transpilePartsWithYourCommand | ractive module > component.js
```

If that design works, great. Otherwise, I'll build a command to compile template, styles, script and partials to an ES module (and maybe a CJS variant of this as well). 👍

packages/node_modules/@ractivejs/cli/src/scaffolds.js
ghost: The things like dynamic CSS and other things the current CLI does is not yet included. Will add those next.

ghost:
1. Once added sandbox as its own package (and repeatedly added/removed it). I can definitely add it back. Considering making sandbox a local instance of playground. Hmm... and add playground as another package. Hmm... 🤔 

2. Yep. It's not so bad if you think about releases in terms of regular releases cycles instead of one-off, per-package releases. Releases get scheduled, and everyone that needs to release new code must be there on time 🚌 🛑  ... or wait another 6 weeks for the next. The bumped digit depends on what the updates contain. No updates, wait another 6 weeks for the next.

    If a plugin needs a new core, it gets released together with the updated core (because monorepo, and 6 weeks is probably enough to realize it needed that core). Otherwise, it gets released in the next cycle as a feature update together with other updates.

3. I only realized that Github didn't show diffs when I made the PR. Some impromptu acting happened. 😁

4. utils-component once had the APIs to mangle ES (and even AMD), but removed them with the intent of having the component files written in a universal manner (can traverse across tools and environments, write once run anywhere). Another reason was to keep our end of the API light and offload all things that mangle code to third-party tools that are more capable of mangling JS.

    I can definitely add the APIs back to handle the other formats, but the plugin author/consumer must be aware that if you start writing ES, the component file cannot be used at runtime without compilation. Also, if ES is the target format, it must be written in ES to begin with (ES to CJS/AMD sounds fine, but CJS/AMD to ES sounds really strange). 

5. This is something I still have to look into. I'm behind in 0.9 progress. 😁 

6. Adding core coverage is definitely doable. Was just thinking of adding it to all. Still looking into how other libs handle their coverage with monorepos. Still very magical.

7. Semis are optional :trollface: (although I've come across Buble bugs that are fixed by adding semis). 

8. (wait for iiit)

9. Profit!

10. 🏈 time!

Definitely the feedback I wanted. Some things missed, some things need improvement, some things new. Back to the drawing board. 👍

.editorconfig
ghost: I think there's a "semistandard", standard but with semis. I'll look into that, as well as prettier. 👍 As long as it's one less thing in the repo, it's totally fine by me.

ghost:
Checkpoint!

- Source-level coverage is now available. https://coveralls.io/github/fskreuz/ractive-monorepo?branch=monorepo-experiment
- Re-added sandbox. All packages (built) are accessible.

Next in the list:

- Re-adding a sandbox-y QUnit.
- Re-adding ESM APIs, adjust cli, loader-rollup and loader-webpack to consume SFCs written in ES.
- Moar unit tests! 93% across all packages is a good sign.
    - Core isn't even included yet. Adding its 97% should bump it up even higher.

ETA: 2-ish weeks?

ghost:
Closing. Working on scripts. Might accidentally run on main repo.

PR #3199, opened 2018-02-26
Custom event plugin init args support

evs-chris:
## Description:
This adds init param support for event plugins - because outside of having a bunch of variations on an event, there's no way to tell the plugin how to behave since the only thing you can pass to the event is what it should do when it fires. This means you can't easily do things like tell a swipe event the threshold distance or bounding coordinate range for firing the event, specify which keys a key event should respond to, tell an event it should never bubble, or specify a custom delay on a hover event for a popover.

Granted some of those aren't too onerous as inputs to a factory method to create event plugins, but the more complex scenarios (swipe events) are where it starts to become painful because for things like a menu pull out, where you only want a small chunk of pixels on the side to be a valid start point for the swipe, and you may also want the plugin to update a keypath with the current relative distance before the event reaches the firing threshold for distance, you start to build up a good handful of variations on the same event installed with different cumbersome names in the instance/component.

As it stands now, this is a net ~18 line change, so it's not particularly heavy, but it also doesn't support things like updating params after the event initializes. It's strictly a one-pass resolution at init, where the args are gathered to pass on to the event plugin. It wouldn't be terribly hard to extend support for arg updates, including a teardown/recreate cycle for plugins that don't supply an update function, but it seems a bit overkill at the moment and can be added without _really_ breaking changes later on.

### Syntax?

The syntax added to the template is a not-particularly-pretty-but-to-the-point `on-event(args,go,here)="@.stillTheAction, called.onFire()"`, where the parens and arg list are entirely optional. If the arg list _is_ preset, they are collected into an array and applied to the custom event constructor after the fire param e.g. `function customEvent(node, fire, args, go, here) { ... }`.

Other considered syntax was `on-event[args,here]="fired"` and `on-event{args,go:'here'}="fired"`, where the former is an array and the latter an object. Of course, whitespace is allowed in there at any point, but editors may or may not get very confused by that. On the plus side, for those emacs users out there, you can always just `C-M-S-t M-S-x sgml-attrs-with-spaces<ret> C-x re-highlight<ret>` 😆

If the parser sees an args list on an event, it collects an array expression into the attribute's new `a` member, which if present, will cause the `EventDirective` VDOM node to evaluate the expression and pass the resulting array on to the plugin.

### Barrier to merge

If there are no objections, I'll probably go ahead and merge this in a day or three because it's fairly light and allows me to handle some of the more complex events that I'm working on right now a bit more gracefully.

A few of my own numerous issues 😀, though I suspect anyone wanting to write/use more advanced composite events, like a double tap or swipe events or use many different key events would appreciate this as well.

## Is breaking:
Nope.

### DIFF
--- a/src/parse/converters/element/readAttribute.js
+++ b/src/parse/converters/element/readAttribute.js
@@ -8,7 +8,7 @@ import flattenExpression from 'parse/utils/flattenExpression';
 import refineExpression from 'parse/utils/refineExpression';
 import { isString } from 'utils/is';
 
-const attributeNamePattern = /^[^\s"'>\/=]+/;
+const attributeNamePattern = /^[^\s"'>\/=(]+/;
 const onPattern = /^on/;
 const eventPattern = /^on-([a-zA-Z\*\.$_]((?:[a-zA-Z\*\.$_0-9\-]|\\-)+))$/;
 const reservedEventNames = /^(?:change|reset|teardown|update|construct|config|init|render|complete|unrender|detach|insert|destruct|attachchild|detachchild)$/;
@@ -234,6 +234,11 @@ export function readAttributeOrDirective ( parser ) {
 	else if ( match = eventPattern.exec( attribute.n ) ) {
 		attribute.n = splitEvent( match[1] );
 		attribute.t = EVENT;
+
+		if ( parser.matchString( '(' ) ) {
+			attribute.a = flattenExpression({ t: ARRAY_LITERAL, m: readExpressionList( parser ) });
+			if ( !parser.matchString( ')' ) ) parser.error( `Expected closing ')'`);
+		}
 
 		parser.inEvent = true;
 
--- a/src/view/items/element/ElementEvents.js
+++ b/src/view/items/element/ElementEvents.js
@@ -55,11 +55,12 @@ class DOMEvent {
 }
 
 class CustomEvent {
-	constructor ( eventPlugin, owner, name ) {
+	constructor ( eventPlugin, owner, name, args ) {
 		this.eventPlugin = eventPlugin;
 		this.owner = owner;
 		this.name = name;
 		this.handler = null;
+		this.args = args;
 	}
 
 	bind () {}
@@ -68,14 +69,14 @@ class CustomEvent {
 		runloop.scheduleTask( () => {
 			const node = this.owner.node;
 
-			this.handler = this.eventPlugin.call( this.owner.ractive, node, ( event = {} ) => {
+			this.handler = this.eventPlugin.apply( this.owner.ractive, [ node, ( event = {} ) => {
 				if ( event.original ) event.event = event.original;
 				else event.original = event.event;
 
 				event.name = this.name;
 				event.node = event.node || node;
 				return directive.fire( event );
-			});
+			} ].concat( this.args || [] ) );
 		});
 	}
 
--- a/src/view/items/shared/EventDirective.js
+++ b/src/view/items/shared/EventDirective.js
@@ -1,5 +1,5 @@
 import { ANCHOR, COMPONENT } from 'config/types';
-import { warnOnceIfDebug } from 'utils/log';
+import { warnIfDebug, warnOnceIfDebug } from 'utils/log';
 import { addToArray, removeFromArray } from 'utils/array';
 import noop from 'utils/noop';
 import fireEvent from 'src/events/fireEvent';
@@ -11,6 +11,8 @@ import RactiveEvent from '../component/RactiveEvent';
 import { resolveArgs, setupArgsFn } from '../shared/directiveArgs';
 import Context from 'src/shared/Context';
 import { isArray, isString } from 'utils/is';
+import getFunction from 'shared/getFunction';
+import resolveReference from '../../resolvers/resolveReference';
 
 const specialPattern = /^(event|arguments|@node|@event|@context)(\..+)?$/;
 const dollarArgsPattern = /^\$(\d+)(\..+)?$/;
@@ -22,21 +24,33 @@ export default class EventDirective {
 		this.template = options.template;
 		this.up = options.up;
 		this.ractive = options.up.ractive;
-		//const delegate = this.delegate = this.ractive.delegate && options.up.delegate;
 		this.events = [];
+	}
 
+	bind () {
 		if ( this.element.type === COMPONENT || this.element.type === ANCHOR ) {
 			this.template.n.forEach( n => {
 				this.events.push( new RactiveEvent( this.element, n ) );
 			});
 		} else {
-			// make sure the delegate element has a storag object
-			//if ( delegate && !delegate.delegates ) delegate.delegates = {};
+			let args;
+			if ( args = this.template.a ) {
+				const rs = args.r.map( r => {
+					const model = resolveReference( this.up, r );
+					return model ? model.get() : undefined;
+				});
+				try {
+					args = getFunction( args.s, rs.length ).apply( null, rs );
+				} catch ( err ) {
+					args = null;
+					warnIfDebug(`Failed to compute args for event on-${this.template.n.join( '- ')}: ${err.message || err}`);
+				}
+			}
 
 			this.template.n.forEach( n => {
 				const fn = findInViewHierarchy( 'events', this.ractive, n );
 				if ( fn ) {
-					this.events.push( new CustomEvent( fn, this.element, n ) );
+					this.events.push( new CustomEvent( fn, this.element, n, args ) );
 				} else {
 					this.events.push( new DOMEvent( n, this.element ) );
 				}
@@ -45,9 +59,7 @@ export default class EventDirective {
 
 		// method calls
 		this.models = null;
-	}
 
-	bind () {
 		addToArray( ( this.element.events || ( this.element.events = [] ) ), this );
 
 		setupArgsFn( this, this.template );
--- a/tests/browser/events/basic.js
+++ b/tests/browser/events/basic.js
@@ -1,3 +1,4 @@
+import { hasUsableConsole, onWarn } from '../../helpers/test-config';
 import { initModule } from '../../helpers/test-config';
 import { test } from 'qunit';
 import { fire } from 'simulant';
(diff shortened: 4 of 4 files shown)


PaulMaly:
Could you please provide some real case here?

evs-chris:
Sure. The case I'm actually working with at the moment is a bit too complex for an example, but here's the key event:

```js
function KeysEvent(node, fire, ...keys) {
  const ctx = this.getContext(node);
  function listener(ev) {
    const which = event.which || event.keyCode;
    if (~keys.indexOf(which)) {
      event.preventDefault();
      fire({ node, event });
    }
  }
  ctx.listen('keydown', listener);
  return {
    teardown() { ctx.unlisten('keydown', listener); }
  };
}
```

```handlebars
<input on-keys(10,13)="@.accept()"
  on-keys(27)="@.cancel()" 
  on-keys(40)="@.nextItem()"
  on-keys(38)="@.previousItem()"
  on-keys(92)="@.previousPage()"
  on-keys(93)="@.nextPage()"
  on-input="@.grid.filter(@node.value)"
/>
```

Here we have a hypothetical input that acts as filter and control point for an auto-paginated grid of items where we're looking to select one of the items. The enter/return key accepts the current item, esc cancels the selection process, up and down arrows move the selection in the grid, and page up and down move pages in the grid. On input, the grid is filtered using the current value of the input.

Another case would be an editable grid wherein the cells allow navigation using arrow keys and value acceptance on enter/cancellation on esc.

The same event plugin could be extended to handle modifiers (shift, ctrl, alt) pretty easily as well, whereas right now it would be a massive explosion of additional event plugins registered in the instance.

For reference for this usecase, I work mostly with line of business apps and am currently working on a project that's moving an old school .net windows forms app into an spa. This particular program has its roots all the way back to an original DOS program, so a decent chunk of the userbase _really_ likes their keyboard navigation.

evs-chris:
Here's a swipe event that allows bounds on the start point, works with touches and (mostly) mouse events, and will bind the current distance or percent of swipe to allow things like menu peeking:

```js

// based on ractive-event-tap

export default function makeSwipe(name, direction, distance = 150, flick = 200) {
  return function setup({ Ractive, instance }) {
    instance.events[name] = function swipe(node, fire, opts = {}) {
      let handler;
      if (handler = node.__r_swipes__) {
        handler.subscribe(direction, opts.distance || distance, opts.flick || flick, opts, fire);
      } else {
        handler = new Handler(Ractive.getContext(node));
        node.__r_swipes__ = handler;
        handler.subscribe(direction, opts.distance || distance, opts.flick || flick, opts, fire);
      }

      return { teardown() { handler.unsubscribe(fire); } };
    }
  }
}

export const left = makeSwipe('swipeleft', 'left');
export const right = makeSwipe('swiperight', 'right');
export const up = makeSwipe('swipeup', 'up');
export const down = makeSwipe('swipedown', 'down');

class Handler {
  constructor(context) {
    this.context = context;
    this.node = context.node;
    this.fires = [];

    this.bind();
  }

  subscribe(direction, distance, flick, opts, fire) {
    this.fires.push({ direction, distance, flick, opts, fire });
    this.hasBinding = !!this.fires.find(f => f.opts.bindPx || f.opts.bind);
    this.hasBounds = !!this.fires.find(f => f.opts.maxX != null || f.opts.maxY != null || f.opts.minX != null || f.opts.minY != null);
  }

  unsubscribe(fire) {
    this.fires = this.fires.filter(f => f.fire !== fire);
    if (!this.fires.length) this.teardown();
    this.hasBinding = !!this.fires.find(f => f.opts.bindPx || f.opts.bind);
    this.hasBounds = !!this.fires.find(f => f.opts.maxX != null || f.opts.maxY != null || f.opts.minX != null || f.opts.minY != null);
  }

  bind() {
    // listen for mouse/pointer events...
    if (window.navigator.pointerEnabled) {
      this.context.listen('pointerdown', handleMousedown);
    } else if (window.navigator.msPointerEnabled) {
      this.context.listen('MSPointerDown', handleMousedown);
    } else {
      this.context.listen('mousedown', handleMousedown);
      this.context.listen('dragstart', handleDragstart);

      // ...and touch events
      this.context.listen('touchstart', handleTouchstart);
    }
  }

  fire(event, startx, starty, endx, endy, duration) {
    const node = this.node;
    let fired;

    this.fires.forEach(f => {
      if (!f.active) { f.active = true; return; }

      const distx = duration < f.flick ? endx - startx + ((f.flick / duration) * (endx - startx)) : endx - startx;
      const disty = duration < f.flick ? endy - starty + ((f.flick / duration) * (endy - starty)) : endy - starty;
      const dist = f.opts.bindPx;
      const pct = f.opts.bind;

      if (distx > 0 && f.direction === 'right' && distx >= f.distance) {
        f.fire({ node, event });
        fired = true;
      } else if (distx < 0 && f.direction === 'left' && -distx >= f.distance) {
        f.fire({ node, event });
        fired = true;
      }

      if (disty > 0 && f.direction === 'down' && disty >= f.distance) {
        f.fire({ node, event });
        fired = true;
      } else if (disty < 0 && f.direction === 'up' && -disty >= f.distance) {
        f.fire({ node, event });
        fired = true;
      }

      if (dist) this.context.set(dist, 0);
      if (pct) {console.log('reset pct', pct, 0, this.context.resolve(pct)); this.context.set(pct, 0);}
    });
    console.log('fire')

    return fired;
  }

  checkBounds(startx, starty) {
    const rect = this.node.getBoundingClientRect();
    const x = startx - rect.x, y = starty - rect.y;

    this.fires.forEach(f => {
      const { maxX, maxY, minX, minY } = f.opts;
      if (maxX > 0 && x > maxX) { f.active = false; return; }
      if (maxX < 0 && x > rect.width + maxX) { f.active = false; return; }
      if (maxY > 0 && y > maxY) { f.active = false; return; }
      if (maxY < 0 && y > rect.height + maxY) { f.active = false; return; }
      if (minX > 0 && x < minX) { f.active = false; return; }
      if (minX < 0 && x < rect.width - minX) { f.active = false; return; }
      if (minY > 0 && y < minY) { f.active = false; return; }
      if (minY < 0 && y < rect.width - minY) { f.active = false; return; }
      f.active = true;
    });

    return !!this.fires.find(f => f.active);
  }

  updateBindings(startx, starty, endx, endy) {
    this.fires.forEach(f => {
      if (!f.active) return;

      if (!f.opts.bindPx && !f.opts.bind) return;

      const dist = f.opts.bindPx;
      const pct = f.opts.bind;
      const distx = endx - startx;
      const disty = endy - starty;

      if (dist) {
        if (f.direction === 'left') this.context.set(dist, distx < 0 ? -distx : 0);
        else if (f.direction === 'right') this.context.set(dist, distx > 0 ? distx : 0);
        else if (f.direction === 'up') this.context.set(dist, disty < 0 ? -disty : 0);
        else if (f.direction === 'down') this.context.set(dist, disty > 0 ? disty : 0);
      }

      if (pct) {
        if (f.direction === 'left') this.context.set(pct, distx < 0 ? (-distx / f.distance) * 100 : 0);
        else if (f.direction === 'right') this.context.set(pct, distx > 0 ? (distx / f.distance) * 100 : 0);
        else if (f.direction === 'up') this.context.set(pct, disty < 0 ? (-disty / f.distance) * 100 : 0);
        else if (f.direction === 'down') this.context.set(pct, disty > 0 ? (disty / f.distance) * 100 : 0);
      }
    });
  }

  mousedown(event) {
    if (this.preventMousedownEvents) {
      return;
    }

    if (event.which !== undefined && event.which !== 1) {
      return;
    }

    const start = new Date();
    const x = event.clientX;
    const y = event.clientY;

    if (this.hasBounds && !this.checkBounds(x, y)) return;
 
    // This will be null for mouse events.
    const pointerId = event.pointerId;

    const handleMouseup = event => {
      this.fire(event, x, y, event.clientX, event.clientY, new Date() - start) && event.preventDefault();
      cancel();
    };

    const handleMousemove = event => {
      if (event.pointerId != pointerId) {
        return;
      }

      if (this.hasBinding) {
        this.updateBindings(x, y, event.clientX, event.clientY);
      }  
    };

    const cancel = () => {
      this.node.removeEventListener('MSPointerUp', handleMouseup, false);
      document.removeEventListener('MSPointerMove', handleMousemove, false);
      document.removeEventListener('MSPointerCancel', cancel, false);
      this.node.removeEventListener('pointerup', handleMouseup, false);
      document.removeEventListener('pointermove', handleMousemove, false);
      document.removeEventListener('pointercancel', cancel, false);
      this.node.removeEventListener('click', handleMouseup, false);
      document.removeEventListener('mousemove', handleMousemove, false);
    };

    if (window.navigator.pointerEnabled) {
      this.node.addEventListener('pointerup', handleMouseup, false);
      document.addEventListener('pointermove', handleMousemove, false);
      document.addEventListener('pointercancel', cancel, false);
    } else if (window.navigator.msPointerEnabled) {
      this.node.addEventListener('MSPointerUp', handleMouseup, false);
      document.addEventListener('MSPointerMove', handleMousemove, false);
      document.addEventListener('MSPointerCancel', cancel, false);
    } else {
      this.node.addEventListener('click', handleMouseup, false);
      document.addEventListener('mousemove', handleMousemove, false);
    }
  }

  touchdown(event) {
    const touch = event.touches[0];

    const start = new Date();
    const x = touch.clientX;
    const y = touch.clientY;

    if (this.hasBounds && !this.checkBounds(x, y)) return;

    const finger = touch.identifier;

    const handleTouchup = event => {
      const touch = event.changedTouches[0];

      if (touch.identifier !== finger) {
        cancel();
        return;
      }

      // for the benefit of mobile Firefox and old Android browsers, we need this absurd hack.
      this.preventMousedownEvents = true;
      clearTimeout(this.preventMousedownTimeout);

      this.preventMousedownTimeout = setTimeout(() => {
        this.preventMousedownEvents = false;
      }, 400);

      this.fire(event, x, y, touch.clientX, touch.clientY, new Date() - start) && event.preventDefault();
      cancel();
    };

    const handleTouchmove = event => {
      if (event.touches.length !== 1 || event.touches[0].identifier !== finger) {
        cancel();
      }

      const touch = event.touches[0];

      if (this.hasBinding) {
        this.updateBindings(x, y, touch.clientX, touch.clientY);
      }  
    };

    const cancel = () => {
      this.node.removeEventListener('touchend', handleTouchup, false);
      window.removeEventListener('touchmove', handleTouchmove, false);
      window.removeEventListener('touchcancel', cancel, false);
    };

    this.node.addEventListener('touchend', handleTouchup, false);
    window.addEventListener('touchmove', handleTouchmove, false);
    window.addEventListener('touchcancel', cancel, false);
  }

  teardown() {
    const ctx = this.context;

    ctx.unlisten('pointerdown', handleMousedown);
    ctx.unlisten('MSPointerDown', handleMousedown);
    ctx.unlisten('mousedown', handleMousedown);
    ctx.unlisten('touchstart', handleTouchstart);
    ctx.unlisten('dragstart', handleDragstart);

    delete this.node.__r_swipes__;
  }
};

function handleMousedown(event) {
  return this.__r_swipes__.mousedown(event);
}

function handleTouchstart(event) {
  return this.__r_swipes__.touchdown(event);
}

function handleDragstart(event) {
  event.preventDefault();
}
```

evs-chris:
Gonna go ahead and merge as experimental for 0.10, as it's worked out quite well during dogfooding.

PR #3203, opened 2018-02-27
Auto-format source on local build...

evs-chris:
...because I'm starting to have a hard time context-switching on when to use the super-spacey style.

This PR is mostly to see how test coverage is affected by istanbul comments being re-arranged.


ghost:
Wondering if this would fit in somewhere https://github.com/prettier/eslint-plugin-prettier . TL;DR: 
- Suppresses ESLint formatting rules while preserving code-quality rules.
- Suppresses colliding rules between ESLint and Prettier.
- Prettier violations appear as ESLint warnings.
- `eslint --fix` will run Prettier instead.

Also, if I understand what Prettier does, some of the rules in `.eslintrc` might not be needed anymore. I imagine the `.eslintrc` would only contain:

```
{
  "extends": [
    "eslint:recommended",
    "plugin:prettier/recommended"
  ],
  "env": {
    "es6": true
  },
  "parserOptions": {
    "ecmaVersion": 6,
    "sourceType": "module"
  },
}
```

evs-chris:
Oooh, that's nicer. Can't quite remove _all_ of the rules from what I've seen, but all of the format-related ones can go. It can also fail the build if the formatting is weird, and it's easy enough to run `npm run lint:fix:all` to fix formatting.

PR #3214, opened 2018-03-14
Make sure progressive enhancement works with aliases

monoblaine:
## Description:

The function `render` of class `Alias` does not take `occupants` as an argument. In order to make progressive enhancement work, it needs to do so. (I hope I'm not missing some bigger picture.)

## Is breaking:

Should not break anything.


evs-chris:
Nope, that's it. Thanks for the fix!

Issue #3097, opened 2017-09-24
Real app scenario I cannot get working with Ractive, but can with Vue (


Inversion-des:
Hello Ractive community!

Last few issues I reported recently are all from my attempts to get working one scenario where dynamically sorted list should be interactive (inline voting, adding new items with transition).
And still I cannot manage to make it fully working (

The problem is that if I use each with some filter ( `{{#sorted(.pros)}}` ) — wrong item is animated and also Vote button doesn't work for path like "`@sorted(task.solutions.0.pros).2`". So I'm playing with the workaround using observer and merge changes into separate .pros_sorted array to support both dynamic sorting and shuffling with correct transitions. But this is cumbersome and there is a bug with canceled previous transition (#3095) even in edge…

**I do not know Ractive very well, so please help me: provide some working code for any version of Ractive to make this scenario working.**

Yesterday I continued my research with attempt to make same sample working on Vue, to compare results. And I was surprised that from the first attempt I have got working solution.

Here are my results for both with some comments:

Ractive 0.8-dev (works same on edge)
https://jsfiddle.net/Inversion/crvqof7t/

- \+ I like how template looks
- – filter cannot be used because wrong node animated on push and Voting is hard   |  `sorted(sol.pros)`  works in Vue
- – observer with additional .pros_sorted is cumbersome solution
- – bug with transitions
- – verbose methods

Vue 2.4.4
https://jsfiddle.net/Inversion/eymc6jz4/

- \+ works as expected without bugs
- \+ all the methods are simpler because of direct access to data, not to relative keypath (compare: vote, sel_sol, sol_cls, add_pro, oncomplete/onmounted). In Ractive I tried to use magic mode for similar effect, but it comes with more bugs (
- – I don't like their additional `<transition>`/`<transition-group>` tags  |  `slide-in` is better
- – I don't like `v-for` inside tags  |  `#each` / `{{#.pros}}` is better

I really like Ractive but in comparison it works not so well for me right now…

ghost:
Hello! Can you describe the requirement (what needs to happen and steps to reproduce?). I'd like to avoid this becoming an [XY problem](https://meta.stackexchange.com/a/66378) where we lock ourselves to the current attempt, when there could be a better solution or alternate approach. 😁 

And if the other issues are related, can you merge them into this issue. That way, we don't have multiple issues of the same bug.

evs-chris:
[...]

Issue #2947, opened 2017-04-26
Built-in filtering and sorting for iterative sections? - RFC 0.10


evs-chris:
Iteration is one of the most common constructs in view libraries, and iteration often needs filtering and/or sorting applied to it. You can currently achieve this to some extent with an expression in ractive, but there are some strong disadvantages, as the expression breaks the connection between the source array and the items in the template e.g. `{{#each articles}}` -> `{{#each filtered(articles)}}` puts the iterations in a computed context. 0.9 has support for bindings in computed contexts, but they're not cheap for non-trivial arrays or expressions.

To address those issues, I'd propose that we adjust the iteration syntax a little and introduce special aliases to handle filter, sort, limit, and offset. With the special aliases applied, the iterative section would adjust the array before it rendered/updated any of the iterations. It would also subscribe to each of the special aliases, so that any updates to them or their deps would cause the section to update as well. Here's roughly what the render cycle would look like:

### Render process

1. Get start array.
2. Apply filter resulting in a mapping of index to index or `-1`.
3. Apply sort resulting in a further mapping of index to target position.
4. Scope the mapping using offset and limit.
5. Unrender any iterations that should no longer be visible and render those that should based on the mapping.
    * Optionally, render transitions could wait on unrender transitions to complete, which would make items appear to leave the list and re-enter at a new position. That could also be done for shuffled sets.

### Syntax change

The index and/or key for an iterative section can currently be aliased by using a `:` after the expression and providing names separated by commas. That doesn't really work with additional things, and it's extra code for a function that's already handled by the normal aliasing that came along later. To accommodate additional special aliases, the index refs would be dropped and replaced with aliases:

```html
{{#each items: i}} ... {{/each}}
{{#each items as item: i}} ... {{/each}}
{{#items: i}} ... {{/}}
{{#each obj: k, i}} ... {{/each}}
{{#each obj as val: k, i}} ... {{/each}}
{{#obj: k}} ... {{/}}
would become
{{#each items with @index as i}} ... {{/each}}
{{#each items as item, @index as i}} ... {{/each}}
{{#items with @index as i}} ... {{/}}
{{#each obj with @key as k, @index as i}} ... {{/each}}
{{#each obj as val, @key as k, @index as i}} ... {{/each}}
{{#obj with @key as k}} ... {{/}}
```

Everything gets a little longer, but it is also considerably more clear. The additional special aliases could appear anywhere in the alias list e.g.

```html
[...]

PR #3221, opened 2018-03-30
Add support for auto shuffling at the iterative block level

evs-chris:
## Description:
This adds experimental support for automatic shuffling at the iterative block level, which means that you can tell an `{{#each}}` block that (or how) to shuffle and it will do so even on regular set updates. This opens up the ability to have shuffling in an iterative section that is dealing with a computed value.

```hbs
{{#each items, true as shuffle}}...{{/each}}
{{#each items as item, true as shuffle}}...{{/each}}
{{#each items, 'some.key.path' as shuffle}}...{{/each}}
{{#each items as item, 'some.key.path' as shuffle}}...{{/each}}
<!-- and of course, you can also leave off the 'as' if that's your thing -->
{{#each items item, 'some.key.path' shuffle}}...{{/each}}
```

The alias syntax is used for consistency and parse-abilty, even though it's not necessarily the _most_ ideal way to convey this particular directive. I'm also planning to add the ability to set the iteration context based on a lookup in a source array via

```hbs
{{#each compute(some(items)), items source}}
  the actual context here will be items[indexOf the current iteration in the source]
{{/each}}
```

That will allow, among other things, filtering lists with bindings without having to do the whole `syncComputedChildren` thing.

Regarding the keypath version of the shuffle setting, I chose to use a string rather than a reference to allow future shuffling based on a function.

If I feel particularly frisky, I may also try to get a shuffle transition in here too, but I may punt on that for now, as this is the last set of things that I want to get in before I cut 0.10.

#3097 #2947

## Is breaking:
Nope

### DIFF
--- a/src/model/Computation.js
+++ b/src/model/Computation.js
@@ -17,6 +17,7 @@ export default class Computation extends Model {
     this.signature = signature;
 
     this.isReadonly = !this.signature.setter;
+    this.isComputed = true;
 
     this.dependencies = [];
 
--- a/src/model/ComputationChild.js
+++ b/src/model/ComputationChild.js
@@ -9,6 +9,7 @@ export default class ComputationChild extends Model {
 
     this.isReadonly = !this.root.ractive.syncComputedChildren;
     this.dirty = true;
+    this.isComputed = true;
   }
 
   get setRoot() {
--- a/src/shared/Context.js
+++ b/src/shared/Context.js
@@ -34,6 +34,8 @@ class ContextData extends Model {
   getKeypath() {
     return '@context.data';
   }
+
+  rebound() {}
 }
 
 export default class Context {
--- a/src/shared/methodCallers.js
+++ b/src/shared/methodCallers.js
@@ -22,9 +22,6 @@ export function marked(x) {
 export function markedAll(x) {
   x.markedAll();
 }
-export function rebound(x) {
-  x.rebound();
-}
 export function render(x) {
   x.render();
 }
--- a/src/shared/methodCallers.js
+++ b/src/shared/methodCallers.js
@@ -22,6 +22,9 @@ export function marked(x) {
 export function markedAll(x) {
   x.markedAll();
 }
+export function rebound(x) {
+  x.rebound();
+}
 export function render(x) {
   x.render();
 }
--- a/src/view/Fragment.js
+++ b/src/view/Fragment.js
@@ -5,6 +5,7 @@ import { getContext, findParentWithContext } from 'shared/getRactiveContext';
 import {
   bind,
   destroyed,
+  rebound,
   shuffled,
   toEscapedString,
   toString,
@@ -222,6 +223,10 @@ export default class Fragment {
 
   rebind(next) {
     this.context = next;
+  }
+
+  rebound() {
+    this.items.forEach(rebound);
   }
 
   render(target, occupants) {
--- a/src/view/Fragment.js
+++ b/src/view/Fragment.js
@@ -5,7 +5,6 @@ import { getContext, findParentWithContext } from 'shared/getRactiveContext';
 import {
   bind,
   destroyed,
-  rebound,
   shuffled,
   toEscapedString,
   toString,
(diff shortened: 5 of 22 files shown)


dagnelies:
Hi Chris,

as a comment, I'd like to point out the syntax is somewhat misleading. First, `shuffling` usually denotes randomly reordering an array, so it's kind of disturbing to see that word. ...the other is: what the heck does this do?! ....I read your description twice, and I still have no clue what this is about 😛

PaulMaly:
The same thoughts. I think we need to have some real example of this feature.

evs-chris:
I'll agree that the syntax is not the best possible way to communicate this, but it's the best compromise I could come up with that fits the rest of what's available in ractive. The `shuffle` alias derives its name directly from the `set` option that does the same thing at the model level. When an item that is rendered in a list is moved within the list, instead of simply updating the dom associated with the item, this will move the dom to its new position. This is keyed updating in other frameworks, and had only been possible in ractive prior to this with non-computed lists using array methods (splice and friends) or a shuffling set. This has come up in a number of issues.

The shuffle part is also very closely related to (well, overlapping with) the second part of this pr, which is mapped iteration contexts, wherein you can instruct an each block to use a list element's source model for the iteration context rather than a computed child context. The easiest way to explain that is `{{#each items.slice()}}{{@keypath}} === @items.slice().${@index}{{/each}}` and `{{#each items.slice(), items source}}{{@keypath}} === items.${@index}{{/each}}`. The benefits are less obvious with such a simple expression to compute a list, but it's very helpful for filtering and sorting lists that needed to have items or parts of items bound later without causing everything associated with the list computation the be recomputed every time some part is touched (syncComputedChildren).

All that said, these are experimental features that I wanted to get a bit of exposure for for feedback, and they don't affect anything if not specifically invoked, making them safe to include in a stable build.

giovannipiller:
> This is keyed updating in other frameworks

hmmm, just to have another example. This:
```
{{#each people as person, 'name' as shuffle}}
	<div>{{person.name}}</div>
{{/each}}
```

Would work as [Svelte's keyed each blocks](https://svelte.technology/guide#keyed-each-blocks)?
```
{{#each people as person @name}}
	<div>{{person.name}}</div>
{{/each}}
```

evs-chris:
> Would work as Svelte's keyed each blocks?

[It does](https://ractive.js.org/playground/#N4IgFiBcoE5SAbAhgFwKYGcUgL4BoRtIQAeDAYxgEsAHFAAhQE8a0BeAHULQA8UB6MCgC2CLvSoATNumE1k6AHwcAdhxQB7VisZg09NAjTC0KlBno0kKw-Qwbdqekw0BXeuWt20+lHoz6klQAbhbWkh4IVOQA1vQkAEauKJo6GioAtORRsZwgAAIAdADmGgAUAJRcihhgrgBm9UYk-Ekp6cpqKMDAAMRoSORglmhaRvRIFqww9ip49ADkKkgmCxMWtQ1NaDg4qurqJEHBij3Ts4XLJrstx53qPfwDQ7uqLRTUdPcqRyES0lwUEgYMU0ChqrcQooQAQMPByOksPQYPQ2PQbAB3egAJUGKBCaDKwH2KWBoJQkEWvSBILBCzwJNk8lQaEpC2pxmZ6HpJMkqCQlOJXXUrDGrPoAG0SQduuiVuKFvUYGhJGt8NLwbKrgrgoZhEg1QzhZrgHKTGzakhisUmIaNQ8zQq+TQwDY1RqALok9VdUqVehCmV+KgYQoBFBle0oBai+RoHnG9RSxMmx1spUqu0ph3ai0IjQJW30H0ymWm3OLZ2u+PFo2l1MVha6hD6rP1nPyi1gK0290pr3GqpdPYqHAVADc+xUMEK9SoKkkAEEEAgygtjgsKrONDAAKKDMBlMpoeZUCqoxQGMNgxcpahtQnr-kZHdUYrzpAIekSCoT1S4IA) indeed.

giovannipiller:
Thank you Chris! This is an awesome feature! 😊
Will start playing with it very soon!

PaulMaly:
Thanks, I, at last, have understood a purpose. Excellent work, @evs-chris !

PR #3249, opened 2018-06-15
Fix project homepage in banner comment and bower.json

m90:
## Description:

The currently used URL `http://ractivejs.org` points to a domain parking page. This uses the `.js.org` version.

n/a


evs-chris:
Thanks for opening this! It looks like an upstream linter change may be killing the build. I'll get that fixed in a bit and see if this passes the build.

evs-chris:
I've cherry-picked this commit onto dev and 0.10, so starting with 0.10.4, this should now be resolved. Thanks again!

PR #3282, opened 2018-11-05
Prevent Safari from moving cursor when setting focused input’s value to the same value

### LINKED ISSUE #3281: <input> on Safari, caret position moves when updateStringValue(...) is not caused by twoway
## Description:
Title might need some improvement I know.
Anyway, I noticed this issue that is Safari specific, but might be interesting to work-around internally.

**Steps:**
1. have an input with twoway="false": `<input type="text" twoway="false" value="{{who}}">`
2. attach an event that manages the "twoway", by listening for input and setting `this.set({ who });` when appropriate
3. load Safari, and start typing at the beginning of the input

**Result:**
As soon as you type the first key, the caret will be moved at the end of the input.
See attached GIF:
caret-position-change-safari

## Versions affected:
1.0.1, likely prior as well

## Platforms affected:
Reproduced on Safari 12, 11

## Reproduction:
[JSFiddle here.](https://jsfiddle.net/qjs09xg2/)

```javascript
const r = window.r = new Ractive({
  el: '#main',
  template: `
  <h1>Hello</h1>
  <div>
  	<p>Start typing <strong>BEFORE</strong> "World".</p>
  	<input type="text" on-input="edit-who" twoway="false" value="{{who}}">
  </div>
  <p style="color: red">{{error}}</p>
  `,

  data: {
    who: 'World'
  },

  on: {
    'edit-who': function(context) {
      const { event } = context;
      const who = event.target.value;

      const originalCaretPosition = event.target.selectionStart;

      this.set({ who });

      const finalCaretPosition = event.target.selectionStart;

      console.log({
        originalCaretPosition,
        finalCaretPosition
      })

      if (originalCaretPosition !== finalCaretPosition) {
        this.set({
          error: 'Caret has been moved to the end of the input.'
        });
      } else {
        this.set({
          error: null
        });
      }
    },
  },
});
```

## Fix (aka workaround):
Looks like Safari moves caret position whenever the value of the focused input is changed, as long as this change is not directly performed by a user action (ex. normal user typing). Which is why `twoway="true"` would work just fine.

Example: `input.value = input.value;` moves the caret at the end in Safari. Not in Chrome/Firefox.

By setting the value **after** the input phase, Ractive will rightfully attempt to update input's value during one of its `update` calls, using the `updateStringValue(..)` function.

An internal workaround might be to check if `input.value` need to be updated at all.
For example, by changing updateStringValue to something like this:

```javascript
// excerpt from src/view/items/element/attribute/getUpdateDelegate.js
function updateStringValue(reset) {
  if (!this.locked) {
    if (reset) {
      this.node._ractive.value = '';
      this.node.removeAttribute('value');
    } else {
      const value = this.getValue();

      this.node._ractive.value = value;

      const safeValue = safeToStringValue(value);

      // FIX: prevents Safari from moving caret position when replacing a value with the same value
      if (this.node.value !== safeValue) {
        this.node.value = safeToStringValue(value);
      }

      this.node.setAttribute('value', safeToStringValue(value));
    }
  }
}
```

## TODO
~~I still have to check WebKit's Bugzilla for this behaviour. In the meantime I'm opening this issue to start the discussion.~~ WebKit issue: https://bugs.webkit.org/show_bug.cgi?id=191255

## TL;DR
`input.value = input.value;` moves the caret at the end of the input in Safari. Not in Chrome/Firefox.
Might be worth to add a workaround for edge cases.

giovannipiller:
fixes #3281

### DIFF
--- a/src/view/items/element/attribute/getUpdateDelegate.js
+++ b/src/view/items/element/attribute/getUpdateDelegate.js
@@ -189,8 +189,14 @@ function updateStringValue(reset) {
 
       this.node._ractive.value = value;
 
-      this.node.value = safeToStringValue(value);
-      this.node.setAttribute('value', safeToStringValue(value));
+      const safeValue = safeToStringValue(value);
+
+      // fixes #3281 – Safari moves caret position when setting an input value to the same value
+      if (this.node.value !== safeValue) {
+        this.node.value = safeValue;
+      }
+
+      this.node.setAttribute('value', safeValue);
     }
   }
 }


src/view/items/element/attribute/getUpdateDelegate.js
```diff
@@ -189,8 +189,14 @@ function updateStringValue(reset) {
...
-      this.node.value = safeToStringValue(value);
-      this.node.setAttribute('value', safeToStringValue(value));
+      const safeValue = safeToStringValue(value);
+
+      // fixes #3281 – Safari moves caret position when setting an input value to the same value
+      if (this.node.value !== safeValue) {
+        this.node.value = safeValue;
+      }
+
+      this.node.setAttribute('value', safeValue);
```
ghost, line 199: Any reason not to include this in the conditional? Just wondering. 😁

giovannipiller: My thinking is that something else might have changed the attribute 'value', for example a decorator.
Like: we could have a `<input .... value="something">` that has `this.node.value === "something-else"`. In this case, we'd rather always update the "value" attribute, just to make sure.

evs-chris:
Thanks! I've got this cherry-picked over to 0.10 and 1.0 as well, so it'll go out with the next round of releases.

PR #3286, opened 2018-11-16
Manage partials’ expressions in CSP mode

### LINKED ISSUE #3285: Partials with expressions are not CSP compliant when { csp: true }
## Description:
Partials (inline or pre-parsed) cannot be used on environments that enforce a strict Content Security Policy (CSP).

It turns out that **only** the expressions in templates are successfully pre-converted to functions.

I worked together with @marcalexiei on this issue.
A full list of test-cases, along with a patch, will be provided shortly in a related pull-request.

## Versions affected:
1.0.x

## Platforms affected:
all pages that enforce a strict CSP policy (ex. [Chrome extensions](https://developer.chrome.com/extensions/contentSecurityPolicy))

## Reproduction:
[JSFiddle](https://jsfiddle.net/5uw08oca/)

```javascript
// inline partial in a template
const a = Ractive.parse(`{{#partial a}}{{1 + 2}}{{/partial}}`, { csp: true });
// look for the expression converted as function in `e`
console.info(a.e) // undefined – doesn't work

// plain template
const b = Ractive.parse(`{{1 + 2}}`, { csp: true });
console.info(b.e) // {1+2: function() {...} } – works just fine
```

Issue also appears with partials pre-rendered separately, or imported from other components (ex. via `Ractive.extend(...)`).

marcalexiei:
## Description:
- Add partials’ expressions to parsed object.
- Add partials’ expressions functions to the `getFunctions` pool.

## Additional notes
- We tried to create a custom registry for partials based `src/Ractive/config/custom/template.js` but encountered issues and maybe it wasn’t worth it since the patch is very small.
- We only patched configure registry function. Should we add it elsewhere?

Fixes #3285

## Is breaking:
No

### DIFF
--- a/src/Ractive/config/registries.js
+++ b/src/Ractive/config/registries.js
@@ -1,4 +1,5 @@
 import { assign, create, keys } from 'utils/object';
+import { addFunctions } from 'shared/getFunction';
 
 const registryNames = [
   'adaptors',
@@ -40,6 +41,12 @@ class Registry {
     assign(registry, option);
 
     target[name] = registry;
+
+    if (name === 'partials' && target[name]) {
+      keys(target[name]).forEach(key => {
+        addFunctions(target[name][key]);
+      });
+    }
   }
 
   reset(ractive) {
--- a/src/parse/_parse.js
+++ b/src/parse/_parse.js
@@ -124,8 +124,10 @@ const StandardParser = Parser.extend({
   },
 
   postProcess(result, options) {
+    const [parserResult] = result;
+
     if (options.expression) {
-      const expr = flattenExpression(result[0]);
+      const expr = flattenExpression(parserResult);
       expr.e = fromExpression(expr.s, expr.r.length);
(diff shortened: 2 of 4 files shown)


giovannipiller:
Do we need to add more info/tests?
If anybody has doubts, feel free to chime in.

We've been developing on this patch for the last 10 days, and haven't noticed issues so far.
We use Webpack and pre-compile templates through a [custom loader](https://github.com/marcalexiei/ractive-html-loader) that accepts { csp: true } as an option.

evs-chris:
Sorry, I meant to merge this earlier. My only concern is with non-object partials, but looking through the points that pull out the expression functions, it looks like they shouldn't cause any issues. Thanks for the fix! I'll cherry-pick this over to the active branches in a bit.

evs-chris:
This is now available in 0.10.13, 1.0.3, and 1.1.1. Thanks!

PR #3294, opened 2019-02-03
Shallow copy value when handling data bound to named checkboxes (fix #3293)

### LINKED ISSUE #3293: Incorrect oldvalue when observing data bound to checkboxes via the name directive
## Description:

When observing array data bound to checkboxes via the name directive, callback is called with incorrect oldValue. OldValue is equal to newValue and shouldn't.

## Versions affected:
Any version since 0.8

## Reproduction:
https://jsfiddle.net/7sxu0r2p/2/

## Fix attempt :
In [/src/view/items/element/binding/CheckboxNameBinding.js](https://github.com/ractivejs/ractive/blob/dev/src/view/items/element/binding/CheckboxNameBinding.js#L71), in method handleChange(), line 71 : we should't get value by reference in order not to modify it too early.

If we replace 
`this.group.value = this.model.get();`
by
`this.group.value = this.model.get().slice();`
we get a shallow copy of the array and it works.

I can submit a PR if this solution seems suitable.

keosion:
## Description:
Before : ractive get named checkboxes value by reference.
After : ractive get a shallow copy of named checkboxes value.

Test is included.

#3293

## Is breaking:
No

### DIFF
--- a/src/view/items/element/binding/CheckboxNameBinding.js
+++ b/src/view/items/element/binding/CheckboxNameBinding.js
@@ -68,7 +68,7 @@ export default class CheckboxNameBinding extends Binding {
 
   handleChange() {
     this.isChecked = this.element.node.checked;
-    this.group.value = this.model.get();
+    this.group.value = this.model.get().slice();
     const value = this.element.getAttribute('value');
     if (this.isChecked && !this.arrayContains(this.group.value, value)) {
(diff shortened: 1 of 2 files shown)


evs-chris:
Thanks! This is now available in 0.10.4, 1.0.4, 1.1.2, 1.2.1, and 1.3.0.

Issue #3311, opened 2019-08-08
Attribute content not escaped correctly in .toHTML()


MartinKolarik:
## Description:

```js
let html = new Ractive({
    template: `<div data-attr='{ "x": 1 }'></div>`,
}).toHTML();

console.log(html);

// actual output: <div data-attr="{ "x": 1 }"></div>
// expected output: <div data-attr="{ &quot;x&quot;: 1 }"></div>
```

## Versions affected:
0.10.1 and up

## Platforms affected:
N/A

## Reproduction:
actual: https://jsfiddle.net/xke7dz2p/1/
expected: https://jsfiddle.net/8q1bhLfe/

MartinKolarik:
https://github.com/ractivejs/ractive/commit/1b58fe544b7fbc023cd3ffea98699363be448b60 seems to be the only commit touching that code between 0.10.0 and 0.10.1 but not yet sure what was the goal there.

MartinKolarik:
So [here](https://github.com/ractivejs/ractive/commit/1b58fe544b7fbc023cd3ffea98699363be448b60#diff-cf3e7b910e9eefd97c69c9b89ebf1cb8R293-R298) almost all attributes get a `g = 1` property, which causes them to go through [this branch](https://github.com/ractivejs/ractive/commit/1b58fe544b7fbc023cd3ffea98699363be448b60#diff-1af7602427c785a77ef5894d8bf09b7cR49-R52) on render, which doesn't use the `Text` renderer that would handle escaping.

PR #3312, opened 2019-08-08
Escape quotes in static attributes - #3311

evs-chris:
This one slipped through the cracks, but there's a test and fix now.

Static attributes were added as a perf enhancement, because _every_ attribute was getting its own fragment, which causes a bunch of extra allocation for something that's just going to end up rendering the same string it already knows.

### DIFF
--- a/src/view/items/Element.js
+++ b/src/view/items/Element.js
@@ -387,7 +387,10 @@ export default class Element extends ContainerItem {
 
     if (this.statics)
       keys(this.statics).forEach(
-        k => k !== 'class' && k !== 'style' && (attrs = ` ${k}="${this.statics[k]}"` + attrs)
+        k =>
+          k !== 'class' &&
+          k !== 'style' &&
+          (attrs = ` ${k}="${safeAttributeString(this.statics[k])}"` + attrs)
       );
 
     // Special case - selected options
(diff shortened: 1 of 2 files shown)


src/utils/html.js
```diff
@@ -387,6 +387,7 @@ export function decodeCharacterReferences(html) {
 const lessThan = /</g;
 const greaterThan = />/g;
 const amp = /&/g;
+const quote = /"/g;
```
MartinKolarik: Hmm, it somehow worked correctly without this before, why is it needed now?

MartinKolarik: I think it used this before: https://github.com/ractivejs/ractive/blob/dev/src/utils/dom.js#L154-L159 (which escapes the correct chars for attributes)

evs-chris: Ah! I was wondering that too, but I didn't look around hard enough. Thanks!

MartinKolarik (approved):
Looks good now. Thanks for a quick fix!

evs-chris:
Do you need this backported to anything beyond the default supported 1.3 and 1.2 series?

MartinKolarik:
1.3 is enough.

Issue #3319, opened 2019-11-01
Observing escaped keypaths with a wildcard doesn't work properly


ilessiivi:
## Description:
Observing a data model like `foo: { "bar.baz": { quuz: 1 } }` with an observer like `foo.*.quuz` doesn't get observed if the last part is an object, or gets called with an old value if the last part is an array (`quuz: [1]`).

The underlying cause could be an unescaped key in [ModelBase.js line 85](https://github.com/ractivejs/ractive/blob/8455933ad8825486be5e347fa6a801b399da4769/src/model/ModelBase.js#L85). I've submitted the pull request #3320 for a possible fix.

## Versions affected:
1.3.8, edge

## Platforms affected:
Tested on Firefox and Safari (both on macOS)

## Reproduction:
https://jsfiddle.net/pwq710hy/1/ (my possible fix is included as an alternative script)

evs-chris:
Your assessment appears to be spot on. I've merged your PR, and will release updates to the 1.2 and 1.3 branches sometime in the next day or so. Thanks!

PR #3320, opened 2019-11-01
Fix observing escaped keypaths (#3319)

ilessiivi:
## Description:
Escapes the child's key (some.key -> some\\.key) in ModelBase's joinKey() and adds a test case for it.

#3319

## Is breaking:
Should not be


evs-chris:
Nice. Thanks for the fix!

PR #3328, opened 2020-04-04
docs: Fix simple typo, werdo -> weirdo

timgates42:
There is a small typo in src/view/items/element/attribute/getUpdateDelegate.js.

Should read `weirdo` rather than `werdo`.


evs-chris:
Thanks! One day I'll learn to typ.

Issue #3172, opened 2018-01-11
Rewrite in TypeScript


ghost:
Just putting this out as a long-term goal. Been using TypeScript, and despite the overhead of learning and writing in a more verbose language that tries to look like Java, the type safety and better auto-complete compensates for a lot of that overhead.

It's also not just about a rewrite. It's also:

- Structured, type-safe code. Writing with discipline is a good habit, but not worrying about it is better.
- Generatable documentation. A side-effect of code that is written in a predictable manner.
- Probably better code optimization. With types and staticness comes better code analysis, linting, etc.
- Better intellisense and integration with projects that use TypeScript.

To the consumer, the library will still be the same old `Ractive`. It's only the library's source that will change. Ractive is also already written in a very class-y manner, so I guess it's already half-way there. Still looking into what else TypeScript can offer, especially in terms of library development.

Lemme know your thoughts 😁

ceremcem:
Just thinking, if any other compile-to-js language fits the bill, would you consider that language too?

ghost:
@ceremcem If it brings the same safeties as TypeScript, why not?

But you'd also need to consider some other aspects that makes TS quite popular:

- It's just JavaScript, and the learning curve is not steep.
- Several other libraries have similar rewrite requests on Github.
- It's backed by Microsoft, so it's not going away anytime soon or die of non-maintenance.
- Its wide adoption in more popular libraries (Angular, RxJS, etc.).
- Syntax support is practically baked everywhere (VS Code, VS, Intellij, Netbeans, etc.).
- ...and more.

It's also much less about "converting it to yet another compile-to-js language" and more of "what can we do in order to do more things out of the box with less effort and hairloss".

Java can compile to JS, but what workflows are there for it? CoffeeScript is also a fine language, but not everyone likes using it. Elm is also good, but I doubt anyone has heard of it. Things like that.

evs-chris:
There's enough cruftiness floating around in the codebase from, what, 6 years of code now? That's a few millennia in JS framework years. I've actually been toying with a rewrite in JS that restructures things a bit to try to lighten up creating instances and rendering templates, and TypeScript sounds like it may speed things up a bit by catching more of my dumber mistakes before hitting the test suite.

avoto:
@fskreuz talking about Java one would be able to even leverage Ractive in native Java by using http://www.jsweet.org/ :)

dagnelies:
@avoto This one makes me curious. Have you already made some project with it by chance? What's your experience?

PR #3334, opened 2020-05-13
Typescript poc

marcalexiei:
## Description:
Hi Chris,
We (@marcalexiei & @giovannipiller) have been thinking to contribute a bit to the project and had a couple of ideas.

### Convert Ractive to Typescript
As suggested in #3172. This could help a lot while refactoring the code or hunting for bugs.

From our initial tests it looks like having a project which mixes JS and TS is possible, which _should_ make things easier for us. We only noticed problems when importing JS from a TS file, but not the other way around.

### Automate generation of API documentation
Documentation seems to be kind of an issue.
Storing part of it alongside the code should make it easier to maintain, reduce duplication and avoid inconsistencies (ex. changing a type in the code should also update the doc)

----

We’ve set up this PR to start some progress and gather some discussion.

To generate the documentation:
1. `npm i`
2. `npm run doc`

As a proof of concept, we added some content for`Ractive.prototype.toHTML()` method and `Ractive.prototype.observe`, which will be found in the following paths:

- {RepoFolder}/docs/modules/_ractive_prototype_observe_.html
- {RepoFolder}/docs/modules/_ractive_prototype_tohtml_.html

## Is breaking:
Nothing, but it's a WIP


evs-chris:
Just wanted to say that I think that a gradual move to ts is a great idea. There are some really chewy parts of ractive that will probably fight typing pretty hard, but it can probably be persuaded. I think it may help to modernize the build system a bit too, since gobble is no longer developed and likely stuck at node 8.

It's great to see a pr open up here! Let me know if you have any questions about the wild and wooly code base. I'd be happy to chat, and I'll try to lend a hand if you'd like when my schedule opens up a bit.

giovannipiller (approved):

src/global/TransitionManager.ts
```diff
@@ -0,0 +1,179 @@
...
+      while (i--) {
+        list = TransitionManager.collectAllOutros(manager.children[i], list);
+      }
+
+      // grab any from this manager if there are any
+      if (manager.outros.length) list = list.concat(manager.outros);
+
+      return list;
+    }
+  }
```
giovannipiller: This is an interesting scenario.
We have converted a non-exported module function into a private static method.

From one point of view, it makes sense to convert `collectAllOutros` into a private function, however we noticed that TypeScript makes all private functions publicly available after they’re compiled.

The big difference here would be that you could technically call `TransitionManager.collectAllOutros(....)` from another JS file.

@evs-chris do you think that it would be a problem?   I’d say we pick one approach and ideally stick to it for every similar scenario in Ractive codebase.

src/global/css.ts
```diff
@@ -11,50 +11,35 @@ let isDirty = false;
...
-          el.innerHTML !== css && (el.innerHTML = css);
-        }
-      }
-    });
-  } else {
-    if (useCssText) {
-      styleElement.styleSheet.cssText = getCSS(null);
-    } else {
-      styleElement.innerHTML = getCSS(null);
-    }
```
marcalexiei, line 51: Hi @evs-chris,
I have removed the use of `useCssText` variable because it works with `styleSheet` property which is not available on `HTMLStyleElement`. Does this makes sense to you?

evs-chris: It does, but it brings up a question of legacy browser support. This was only here to support IE with its funky non-standard properties. Personally, I'm for dropping support for IE and legacy edge, but ractive has supported them for forever.

src/global/TransitionManager.ts
evs-chris: I lean more towards module functions than methods because they minify and tree shake so much better. In fact, part of my long term plan is to reduce some of the classiness in the code base, as it makes things a bit heavier than they should be.

src/global/css.ts
kouts: I think IE11 should be supported

src/global/TransitionManager.ts
marcalexiei: Ok, I'll revert the static methods added in Runloop and TransitionManagar

src/global/css.ts
marcalexiei: I see that Ractive drop support for IE8 from version 0.8.0. So I tested the current build on IE 9, 10 and 11 (using IE 11 simulator) and with `innerHTML` it works fine.

(Click on image to view it full size)

### IE 9
ractive-css-ie9

### IE 10
ractive-css-ie10

### IE 11
ractive-css-ie11

Anyway if you feel more comfortable with the previous implementation just let me know and I will restore it.

evs-chris: Most of the weird stuff has comments around it, usually pointing at IE or Safari, so as long as it doesn't break a supported browser, I'm fine with it. I'm also fine with dropping IE support, as we could term the ts-ified version ractive 2.0 and I can continue backporting bugfixes for the last 1.x branch on-demand. This particular weird thing may have been for an older Safari, which I've never had access to to test, but either way, I'm fine with it, as html5 is good enough for legacy now.

marcalexiei: > Most of the weird stuff has comments around it, usually pointing at IE or Safari, so as long as it doesn't break a supported browser, I'm fine with it.

I totally agree with that.

> I'm also fine with dropping IE support, as we could term the ts-ified version ractive 2.0 and I can continue backporting bugfixes for the last 1.x branch on-demand.

Maybe some people who use Ractive have to support IE11 and may not update to version 2 if it doesn't support IE11, so I agree with @kouts.

> This particular weird thing may have been for an older Safari, which I've never had access to to test, but either way, I'm fine with it, as html5 is good enough for legacy now.

🥳

giovannipiller: I would suggest keeping support for IE11, to make It easier for people to keep upgrading Ractive from earlier versions. I'm not against introducing breaking changes, but removing browser support is tricky, since I'm sure a lot of people using Ractive are still forced to support IE11.

I'm one of them 🙌 😭

.eslintrc.json
```diff
@@ -0,0 +1,109 @@
...
+			}, {
+				"pattern": "utils/**",
+				"group": "internal"
+			}],
+			"newlines-between": "always",
+			"alphabetize": {
+				"order": "asc",
+				"caseInsensitive": true
+			}
+		}],
```
marcalexiei: HI @evs-chris, I've added `eslint-plugin-import` to consistently handle imports. Are you ok with the following configuration?

src/parse/Parser.ts
```diff
@@ -1,52 +1,82 @@
...
+export type Converter = (parser: Parser) => any;
+
+export interface CustomParser {
+  init: (str: string, options: BaseParseOpts) => void;
+  postProcess: (str: string, options?: BaseParseOpts) => any;
+}
+
+// todo add correct return types on props and methods
+// todo need to add parser type <T> and use it in result and postProcess function
+class Parser {
```
marcalexiei: Hi @evs-chris, 
I've rewritten `Parser` / `JsonParser` and `StandardParsers` using class syntax in order to have better code navigation during development of the various converters.
I've also added an interface that describe every item returned by converters function:
- `src/parse/converters/expressions/expressionDefinitions.ts`
- `src/parse/converters/templateItemDefinitions.ts`

Let me know if there is something that needs to be changed before I can move to other Parser sections.

.eslintrc.json
evs-chris: Yep, that looks fine to me.

src/parse/Parser.ts
evs-chris: No, looks good!

marcalexiei:
Hi @evs-chris, I have more than 50% of the conversation (I think :D).
If you have some spare time could you please take a look and let me know what do you think about it.
I left a large amount of `any` across the code but I think we will can get rid of them in the future.

Thanks in advance!

evs-chris:
Yes! I should be able to clone this weekend and give you a bit of feedback.

evs-chris:
I forgot how much code is tucked into the various corners of this code base. I think you're well on track to getting everything converted. It's a tremendous undertaking.

The only thing that jumps out is how very out of date the build system is. I need to spend a little time moving off of gobble, which has been unsupported for a while. I think we can safely drop buble too, since typescript handles all of that and more.

marcalexiei:
> I forgot how much code is tucked into the various corners of this code base. I think you're well on track to getting everything converted. It's a tremendous undertaking.

Now I'm arguing with `MustacheContainer `which is not properly TS friendly :D

> The only thing that jumps out is how very out of date the build system is. I need to spend a little time moving off of gobble, which has been unsupported for a while. I think we can safely drop buble too, since typescript handles all of that and more.

Yes, I think we can do all the bundling via Rollup, which is already used, but IMHO I would like to do this after TS refactor is complete.

marcalexiei:
Hi @evs-chris,  refactor is nearly complete but there are some points that I would like to bring to your attention:

### General information
- I have created a RactiveDefinition file which define Ractive as a class. 
   Should we consider making Ractive as a class?
- There are many `any` keywords which cause many eslint warning.  I would like to leave them and progressively remove them in the future. In the meantime, I added `--quiet` option on `lint:all` task to exclude them from the build process log
- All implementation change due to this refactor have a comment `TSRChange` followed by a briefly description of the change
- I updated node to version 12 (ESlint no longer supports node 8)

### Improvements To-Do
- It would be nice to generate ts definition file automatically. 
- Update build system
- Automatic documentation generation
- Some Ractive prototype function may use generic typings to improve their param signature

Do you want to procede with a merge to a 2.x branch or would you like to make some adjustments?

giovannipiller:
Hi everyone, just as a FYI, next week I'm planning to run a few tests using a build of this branch on a pretty complex web-app.
As of now I've been able to build and run everything without any obvious issue.

I'll report back next week!

evs-chris:
Nice! I think this would be a good opportunity to start a 2.x branch.

I personally think a factory may be a better pattern than a class, but I'm not positive. Going into 2.0, we have the opportunity to adjust some things.

@marcalexiei would you like to be added as a maintainer?

On a somewhat related topic, I cleaned up the rollup-plugin-ractive-bin project a bit last week, and it now integrates much more cleanly with the typescript rollup plugin. What do you usually use to build your projects?

marcalexiei:
> I personally think a factory may be a better pattern than a class, but I'm not positive. Going into 2.0, we have the opportunity to adjust some things.

Yes, I imagine that there is some way to combine factory pattern and types. 
In my opinion the main goal is to find a way to have automatic definition generation. 
Any solution that makes this possibile is a good to go.

> @marcalexiei would you like to be added as a maintainer?

Yes. Thank you 😊

> What do you usually use to build your projects?

I use webpack with a [custom HTML loader](https://github.com/marcalexiei/ractive-html-loader) which parse HTML into a template object (basically a fork from official webpack-html-loader).
Basically for each component there are 3 files:
- component.rhtml -> which is parsed by the custom loader
- component.css
- component.js -> import the two other files and export the Component "definition"

marcalexiei:
Hi @evs-chris, do you want to make other adjustments before merging into a new branch?

Just re-checking 😀

giovannipiller:
I'm having a few issues while using the runtime.js bundle of Ractive. I'll check them out with @marcalexiei and report back.

evs-chris:
No, I think we should probably go ahead and merge this into the new 2.0 branch and carry on from there!

marcalexiei:
As a next step I will try to remove as many "any" as possible from the source code.

PR #3350, opened 2020-12-02
bump node version to 12 and do a bit of audit fixing

evs-chris:
## Description:

This moves the build to node 12 and does some auditing to reduce npm complaining. I'm mostly curious to see if it will pass travis, as there used to be issues with some deps and node > 8.

None.

## Is breaking:

Shouldn't be.


evs-chris:
This seems to work with travis, and local dev seems to remain intact. Any objections to merging? It leaves us leaning heavily on the package lock to get things working, but that's just where things are until we move away from gobble for builds.

marcalexiei (approved):

marcalexiei:
After merge I'll see if this works on 2.0 branch

PR #3353, opened 2020-12-08
Rollup as only build tool

marcalexiei:
## Description:
Hi Chris,
Today I had some free time and decided to play with rollup to update the build system.
At the moment:
- I am able to serve dev build and server for sandbox (without hot reload)
- `bundle:dev` and `bundle:release` now are fully based on rollup
  - using rollup seems to reduce also build time by 30/40%!
Screenshot 2020-12-08 at 18 26 34

- `bundle: test` is not converted

Let me know if you find this approach feasible.

None

## Is breaking:
Hopefully not!


evs-chris:
This is absolutely on the list of things to fix! Replacing gobble would clean up 99% of our old deps issues, possibly along with moving from buble to something like sucrase. I like rollup as the main replacement, and if there are a few weird corners it can't cover, bash scripts can fill in those gaps.

evs-chris (changes requested):
Looking good! If you can get the tests moved over to rollup too, this will be a much better replacement for #3350. Can you make the swap to node 12 or 14 here too?

marcalexiei:
> this will be a much better replacement for #3350

Do you want to have this PR merged on dev?

> Can you make the swap to node 12 or 14 here too?

At the moment this branch is based on 2.0 branch which already have Node 12 ([commit reference](https://github.com/ractivejs/ractive/commit/ae7fff68517a64064c07140c606b27cc6a5c3df0)) in order to use ESlint.
I don't see any blocker for update to Node 14.

evs-chris:
Ah, I missed the target branch. In that case, no, the other build pr will suffice for dev for now. Good to know that 2.0 will have saner build machinery, though.

marcalexiei:
### General overview
Gobble is no longer the main build tool of the project. Now all files are created directly by rollup.
- update Node to v14
- for every build configuration  there is `rollup.{config}.js` file located in the root of the project
- build utils / configuration are stored inside `rollup.utils.js`
- build process does the same task of gobble but it seems more fast.
- transpile tool has not changed. 
- to minify code uglify has been replaced by terser

### `perf` folder
I noticed that there’s a gobble task to serve content inside these folder. I migrated this task to rollup as well, but it seems that the page is not working. After a brief investigation, the issue seems to occur also on dev branch. The tests use an old Ractive Syntax, which may be causing the issue.

### Security issues
We have fewer security issues:
before (dev): 354 vulnerabilities (257 low, 5 moderate, 92 high)
now (ts + rollup): 251 vulnerabilities (176 low, 3 moderate, 72 high)

---

I think that this PR can be merged. 
Let me know if there is something that needs to be changed.

marcalexiei:
Hi @evs-chris, 

Just re-checking 😀, do you want to make additional adjustments/changes before merging?

evs-chris:
Nope, LGTM!

Issue #3369, opened 2021-03-19
Follow up #3335 - Error during Webpack compilation


marcalexiei:
## Description:
We use a [custom Webpack loader](https://github.com/marcalexiei/ractive-html-loader) to pre-parse template on compile time.
After upgrading to 1.4.0-build-50 we get the following error:
```
UnhandledPromiseRejectionWarning: TypeError [ERR_INVALID_ARG_TYPE]: 
The "data" argument must be one of type string, Buffer, TypedArray, or DataView. Received type undefined
```

After some debugging we noticed that the error was caused by `require('ractive')` inside loader file.

We narrowed down the issue to `dom.js` and figured out that the cause was the new check for custom elements:
```
const customStr = 'registerElement' in doc;
```

## Versions affected:
1.4.0-build-50

## Platforms affected:
Node

## Reproduction:
N/A

PR #3370, opened 2021-03-19
make sure we are in a browser before registerElement check  - #3369

marcalexiei:
## Description:
Adding isClient check before `registerElement ` check to avoid issue while require Ractive in Node env.

#3369

## Is breaking:
Hopefully not


evs-chris:
Makes sense, thanks!

PR #3431, opened 2024-03-12
Add ondestruct() to typings

seijikun:
## Description:
This adds the `ondestruct()` lifecycle method to the typings. That's currently missing.


evs-chris:
Thanks! Do you need this published now, or can it wait a bit to see if any more changes happen along?

seijikun:
> Thanks! Do you need this published now, or can it wait a bit to see if any more changes happen along?

Thanks for the quick merge!
Unfortunately, we already depend on it, so if it's not too much work for you, I'd really appreciate a release.

